PR #10, opened 2018-07-23
add Amazon US Optional list, IRC channel

charlesrocket:
* formatting
* typos


charlesrocket:
I created dedicated channel on freenode, @rootzoll grab `oO` if you would like keep it running 😄 maybe we could fix the leak on webseed

rootzoll:
thanks for cleanup :) I am not always-on IRC ... but lets give it a try :)

PR #16, opened 2018-07-27
expand FS command

### LINKED ISSUE #15: my nodes sd card ran out of space
My SD card was getting close to being full (7gb),
and autocomplete did not work properly.

`df -h` showed that /dev/root was not full size
as i have an 16gb card 

a 
1. `sudo raspi-config`
2. `7 Advanced Options`
3.  `A1 Expand Filesystem`
followed by a reboot did the trick that system can use all the sd cards storage.

Is the script missing the expand filesystem command?
would make sense to add `raspi-config --expand-rootfs`

vnnkl:
expands filesystem,
does require a reboot though,

so maybe move it to 90finishsetup.sh instead before we reboot?

closes #15

### DIFF
--- a/home.admin/10setupBlitz.sh
+++ b/home.admin/10setupBlitz.sh
@@ -91,7 +91,7 @@ if [ ! -f "home/admin/.setup" ]; then
   sudo apt-get update
   sudo apt-get upgrade -f -y --force-yes
   echo "OK - System is now up to date"
-
+  sudo raspi-config --expand-rootfs
 fi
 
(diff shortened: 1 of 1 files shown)


rootzoll:
I moved it into 90finishSetup.sh for now, to prevent a reboot during setup. 

That the rootFS fills up during Setup seems strange - would be better to find the reason first, before just giving it more space.

vnnkl:
fyi,
it was not during setup, i've had the node running for a couple of days already.

PR #20, opened 2018-07-28
extend LCD dialog to show ipv6 addresses correctly

ppachur:
If your network is ipv6 only, $l2 will be wrapped and the address is truncated. As a result, $l3 is also not shown. This change adjusts the width of the box to allow everything to be displayed correctly.


rootzoll:
much thanx

PR #31, opened 2018-08-05
parse ipv4

charlesrocket:
fix display breaking for ipv6 network adapter prt


rootzoll:
thanks merged. FYI @piezzo @charlesrocket you both were working at this point

PR #39, opened 2018-08-09
WIP: add TorService script

### LINKED ISSUE #4: Run RaspiBlitz behind TOR
Running the RaspiBlitz behind Tor as a hidden service would be great ... not just because beeing more anon. Its perfect to make the node publicly available even when behind a NAT - without having to set port forwarding, etc.

See discussions for inspiration here: https://twitter.com/Stadicus3000/status/1017704449303867393

First step would be to write a shell script that installs and setups a TOR hidden service and takes care about the settings. If we then have also a script that would reverse this process, we could add a option into the main menu to switch TOR on and off.

vnnkl:
scripted worked on my machine,
no way to reverse it yet, so adding only for now

- [ ] needs menu item

Let's discuss how much we want in there
closes #4

### DIFF
--- a/home.admin/96addTorService.sh
+++ b/home.admin/96addTorService.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+
+# https://medium.com/@lopp/how-to-run-bitcoin-as-a-tor-hidden-service-on-ubuntu-cff52d543756
+# will output: "Codename: stretch" -.- always...
+# codename="$(lsb_release -c)" |  cut -c11-
+torrc="/etc/tor/torrc"
+
+function uncommentTorrcRelayBitcoinOnly {
+	echo "uncommenting #RunAsDaemon 1"
+	sudo sed -i "s/^#RunAsDaemon 1/RunAsDaemon 1/g" $torrc
+	echo "adding PortForward 1 & ControlPort 9051 after RunAsDaemon 1"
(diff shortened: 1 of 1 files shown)


rootzoll:
cool. I will test beginning next week.

rootzoll:
I commited further TOR integration based on this PR:
https://github.com/rootzoll/raspiblitz/commit/1c8ad20bb794b968de60bfce916d9cd3f033b726

I got bitcoin behind TOR running.  But LND seems to have a problem, that the compiled/binary version for ARM of 0.4.2-beta does not contain the TOR freatures - LND exits on start when given tor parameters - I tried --tor.v2 and --tor.v3 ... someone in the LND Slack recomended to compile latest LND on ARM fresh from source ... but thats where I stoped because of time.

Will get back later to this.

vnnkl:
yes, stumbled upon that as well. only running bitcoind behind tor yet. haven't looked much further into yet either

rootzoll:
LND Slack provided this help to build LND from source on Raspi ... to be tested:

1) Install Go 1.10
`wget https://storage.googleapis.com/golang/go1.10.linux-armv6l.tar.gz`
`sudo tar -C /usr/local -xzf go1.10.linux-armv6l.tar.gz`
2) Add these lines to .bashrc
`export GOPATH=~/gocode`
`export PATH=$PATH:$GOPATH/bin`
3) Install lnd, follow instructions here: https://github.com/lightningnetwork/lnd/blob/master/docs/INSTALL.md#installing-lnd

PR #40, opened 2018-08-09
WIP: update script mechanics

charlesrocket:
right now the only stable way to get an update without hardcoding the version is to use github api

### DIFF
--- a/home.admin/XXupdateScripts.sh
+++ b/home.admin/XXupdateScripts.sh
@@ -6,17 +6,15 @@ cd /home/admin/raspiblitz
 get_latest_release_tag() {
   curl --silent "https://api.github.com/repos/charlesrocket/raspiblitz/releases/latest" |
     grep '"tag_name":' |
-    sed -E 's/.*"([^"]+)".*/\1/'                                  
+    sed -E 's/.*"([^"]+)".*/\1/'
 }
-
 vTAG=`get_latest_release_tag`
-
 wget https://github.com/charlesrocket/raspiblitz/archive/$vTAG.zip
-
+unzip $vTAG.zip
 cd ..
 rm *.sh
 rm -r assets
-sudo -u admin cp /home/admin/raspiblitz/home.admin/*.sh /home/admin
+sudo -u admin cp /home/admin/raspiblitz/raspiblitz-$vTAG/home.admin/*.sh /home/admin
 sudo -u admin chmod +x *.sh
 sudo -u admin cp -r /home/admin/raspiblitz/home.admin/assets /home/admin/
 echo "******************************************"
--- a/home.admin/XXupdateScripts.sh
+++ b/home.admin/XXupdateScripts.sh
@@ -9,7 +9,7 @@ get_latest_release_tag() {
     sed -E 's/.*"([^"]+)".*/\1/'
 }
 vTAG=`get_latest_release_tag`
-wget https://github.com/charlesrocket/raspiblitz/archive/$vTAG.zip
+wget https://github.com/raspiblitz/raspiblitz/archive/$vTAG.zip
 unzip $vTAG.zip
 cd ..
 rm *.sh
--- a/home.admin/XXupdateScripts.sh
+++ b/home.admin/XXupdateScripts.sh
@@ -4,7 +4,7 @@ echo "*** UPDATING SHELL SCRIPTS FROM GITHUB ***"
 echo "******************************************"
 cd /home/admin/raspiblitz
 get_latest_release_tag() {
-  curl --silent "https://api.github.com/repos/charlesrocket/raspiblitz/releases/latest" |
+  curl --silent "https://api.github.com/repos/rootzoll/raspiblitz/releases/latest" |
     grep '"tag_name":' |
     sed -E 's/.*"([^"]+)".*/\1/'
 }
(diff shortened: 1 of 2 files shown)


charlesrocket:
releases should be tagged in digit-only format `0.5`, `0.6` (or stand-alone tag)

rootzoll:
thx. I will take a look at it next week.

rootzoll:
sorry this week was kind of full to get to it. Hope to be able to review it soon.

charlesrocket:
@rootzoll 💯 no worries

home.admin/XXupdateScripts.sh
```diff
@@ -1,16 +1,23 @@
 #!/bin/bash
 echo ""
 echo "*** UPDATING SHELL SCRIPTS FROM GITHUB ***"
-echo "justincase, not the final upadte mechanism"
 echo "******************************************"
 cd /home/admin/raspiblitz
-git pull
+get_latest_release_tag() {
+  curl --silent "https://api.github.com/repos/rootzoll/raspiblitz/releases/latest" |
```
vv01f, line 7: results in error:
```
{
  "message": "Not Found",
  "documentation_url": "https://developer.github.com/v3/repos/releases/#get-the-latest-release"
}
```
because there is no release tag ([see](https://github.com/rootzoll/raspiblitz/releases)) yet.

suggestions:
* test if that is the case before acting (think: this could be a fork repo as well)
  * other possibility to fetch the tag: `latestTag=$(git describe --tags $(git rev-list --tags --max-count=1))
` while it needs the git locally
  * while not having releases downloading the (latest state) branch is an option: `https://github.com/rootzoll/raspiblitz/archive/master.zip`
* put URL or parts in variables
* keep in mind: sketched fn only works for GitHub that way, other Git hosters might work differently

home.admin/XXupdateScripts.sh
```diff
@@ -1,16 +1,23 @@
...
+    grep '"tag_name":' |
+    sed -E 's/.*"([^"]+)".*/\1/'
+}
+vTAG=`get_latest_release_tag`
```
vv01f: better readable subshell: `$( function_call )`

home.admin/XXupdateScripts.sh
```diff
@@ -1,16 +1,23 @@
...
+wget https://github.com/raspiblitz/raspiblitz/archive/$vTAG.zip
+unzip $vTAG.zip
 cd ..
 rm *.sh
 rm -r assets
-sudo -u admin cp /home/admin/raspiblitz/home.admin/*.sh /home/admin
+sudo -u admin cp /home/admin/raspiblitz/raspiblitz-$vTAG/home.admin/*.sh /home/admin
```
vv01f: better variable usage: `${varname}`

home.admin/XXupdateScripts.sh
```diff
@@ -1,16 +1,23 @@
...
 sudo -u admin chmod +x *.sh
 sudo -u admin cp -r /home/admin/raspiblitz/home.admin/assets /home/admin/
+rm /home/admin/raspiblitz/$vTAG.zip
```
vv01f: better variable usage: `${varname}`

home.admin/XXupdateScripts.sh
```diff
@@ -1,16 +1,23 @@
...
 echo "******************************************"
 cd /home/admin/raspiblitz
-git pull
+get_latest_release_tag() {
+  curl --silent "https://api.github.com/repos/rootzoll/raspiblitz/releases/latest" |
+    grep '"tag_name":' |
+    sed -E 's/.*"([^"]+)".*/\1/'
+}
+vTAG=`get_latest_release_tag`
+wget https://github.com/raspiblitz/raspiblitz/archive/$vTAG.zip
```
vv01f: when using `curl` elsewhere already there is no need for `wget`

vv01f (commented):
suggestions…

home.admin/XXupdateScripts.sh
charlesrocket: for method to work maintainer should create a release like described in the PR

charlesrocket: created sample in 
https://github.com/charlesrocket/raspiblitz/releases/latest you can test it replacing `rootzoll` -> `charlesrocket` in the update script

rootzoll:
I will start to make sure more stabel version building gets possible starting from version 1.0 ... targeted for early October now. I will be on holiday for the rest of the month .. then i will catch up.

charlesrocket:
@rootzoll sounds good! let me know ill walk you thro the method

rootzoll:
@charlesrocket @vv01f sorry for taking so long to come back to this. I choose another way for update the update mechanism - let me explain. 

The XXupdateScripts was just a hot fix during development - even when working with tag versions there was always this one problem that was bugging me: The scripts may get updated, but the data and the OS on the SD card will stay behind and get out of sync. First I was thinking about adding a "migration feature" (adding, removing services, new bitcoind/lnd, changing data, etc) to the updatescript, but it felt not like a solid way for the future. 

So I turned the the idea upside down. Why not just add a new sd card image into the raspiblitz with all the newest software and continue with the old data. Sure also data migration between versions may be needed in the future ... but that way the same mechanism can be used for recovery when the SD card fails.

What I most like about it, people can get wild changing the RaspiBlitzOS (adding services, etc) but if it fails, you just make a fresh SD card and all the standard services should continue to run as normal again - this also makes support easier, because with a fresh sd card I know the state of the system more clearly .. at least I hope so.

Thanks for your contribution - the idea with the tags was worth a shot .. but I chose another way to go on this. Closing this PR.

PR #47, opened 2018-08-15
configure ufw

### LINKED ISSUE #45: configuring firewall
I was looking to see if raspiblitz has a firewall activated but realized you just install it but don't configure it.

https://github.com/rootzoll/raspiblitz/blob/ee67947a631926779783d4cc22127108b4d8e284/build.sdcard/raspbianStretchDesktop.sh#L92

for basic usage (bitcoin + lightning) the following patch should fix this:

```
ufw allow ssh
ufw allow 8333
ufw allow 9735
ufw default deny incoming
ufw default allow outgoing
ufw enable
```

vnnkl:
added configuration as suggested by #45 
+ testnet

closes #45

### DIFF
--- a/build.sdcard/raspbianStretchDesktop.sh
+++ b/build.sdcard/raspbianStretchDesktop.sh
@@ -97,7 +97,9 @@ echo "allow: bitcoin testnet"
 ufw allow 18333
 echo "allow: bitcoin mainnet"
 ufw allow 8333
-echo "allow: lightning"
+echo 'allow: lightning testnet'
+ufw allow 19735
+echo "allow: lightning mainnet"
 ufw allow 9735
 echo "deny incoming connection on other ports"
 ufw default deny incoming


rootzoll:
thx. 
There is also litecoin mainnet: 9333
And maybe Port 80 for the Web-Admin to come in the future?

vnnkl:
was not sure if we wanted to open ports which might not be used later rather than opening them on demand, your call, then shango 10009 as well

rootzoll:
Also not sure if the torrent-download might have problems with a too tight firewall (some seeding while downloaded would be nice). FTP should be fine with downloading from behind that.

charlesrocket:
@rootzoll @vnnkl `transmission` will require open firewall or port forwarding, so it might be easier to put `ufw` option after the torrent stage is complete?

PR #49, opened 2018-08-17
add bash completion to bonuses

vnnkl:
adds bash completion, gets automatically installed at end with lnchannels & wallet before restart.


rootzoll:
makes total sense :) thx

PR #89, opened 2018-09-10
Fix Download.progress file name

lsilvs:
Using single quote creates a file named `.${name}.progress` rather than `.Download.progress`


rootzoll:
good catch :) thx

PR #109, opened 2018-09-21
Update SD card from Class 4 to Class 10

coblee:
Update prices


rootzoll:
Yeah - much cheaper now :D thx

PR #153, opened 2018-10-23
shoppinglist for czech republic

karliatto:


geco91:
@karliatto the power outlet seems to be too weak with 2,5A. To run the HDD without power outages at least a 3A power outlet is recommended - even a bit more is better. Can you update the the list?

rootzoll:
This power outlet looks good to use - needs just maybe an additional micro usb cable:
https://www.alza.cz/EN/alzapower-q100-quick-charge-3-0-black-d5324257.htm?o=2

PR #154, opened 2018-10-23
Create Shopping list_AU

DyptoDurrency:
Hi guys,

I love what you guys are doing. I have created my own lightning node with another guide and wanted to help anyone in Australia have an easier time following your guide by adding a no-brainer shopping list.

Keep it up


geco91:
@DyptoDurrency the power outlet seems to be too weak with 2,5A. To run the HDD without power outages at least a 3A power outlet is recommended - even a bit more is better. Can you update the the list?

PR #171, opened 2018-11-11
Typo/linebreaks fixes and remove more unneeded packages

botweb:
Small fixes in setup outputs and remove more unneeded packages.


rootzoll:
such correct. much thx :)

PR #179, opened 2018-11-25
Update README.md

rytisbitcoin:
Fixed various typos, and, more importantly, fixed the instructions to ask the user to connect using "admin", not "root", as "root" does not work in this situation / with the password "raspiblitz". Also added the step of the user having to press the "Cancel" button to cancel the installer which launches automatically, as it was not mentioned in the guide.


rootzoll:
much thx :)

PR #184, opened 2018-11-29
get lnd-version on lcd-display from lncli

ppachur:
LND-version on display was hardcoded. This uses the version supplied by lncli (without calling lncli more often).


rootzoll:
much thx.

PR #186, opened 2018-11-29
Add ufw allow for REST API

wbobeirne:
Just helped a user debug hooking up their raspiblitz to my chrome extension Joule (https://github.com/wbobeirne/joule-extension) which hit a snag because that extension has to use the REST API rather than the gRPC API due to browser limitations. I've added a ufw allow for 8080 to open it up. I personally haven't been able to test this, but I assume this is enough to make it accessible. Let me know if I'm wrong!


rootzoll:
Thas a good idea. Thanks.

PR #200, opened 2018-12-15
Raspiblitz 3D printer case 

whiteyhat:
Hi!

Here there is a prototype to have a 3D printer case open source. The idea is to have a container for the HDD, where on top of it is the raspiblitz. 

Let me know your thoughts!


fluidvoice:
I'm kinda wondering the purpose or usefulness of this as hardly anyone will want or be able to print their own.  The vast majority of Raspiblitz users will opt to buy their printed case (with HDD container) from twitter.com/CryptoCloaks or other case maker.  Is the purpose just to design a perhaps better case implementation than that from CryptoCloaks?

whiteyhat:
The cost of production of this case is around 1.40-2.00 EUR. The 3D files are open sourced as well as this project. There is a vast majority with access to a 3D printer. 

You can always go to cryptocloaks and buy a case for 30-40USD 😉

rootzoll:
I just liked the idea of having the option and the open files to print a case if you want to. Over the next days we will print some of these at the 35c3 at the bitcoin assembly.

The CryptoCloaks is cool, but not OpenSource yet (can be added later if they like). More options, the better. Buy Amazon standard, order a custom 3D printed cloak or print a basic shell yourself .. its all about the freedom of choice. Happy christmas to all :)

PR #212, opened 2018-12-24
Add shopping list for Switzerland and driver for custom display

guggero:
Some parts from the shopping list for Germany cannot be delivered to Switzerland by Amazon.

That's why I went ahead and found similar parts in Swiss online shops.

Unfortunately, the display is not the same and needs a custom driver.
So I have added a dialog in the `./raspbianStretchDesktop.sh` script that asks which display you are using.


rootzoll:
Yeah thats cool .. selecting a display can be extended in the future. Like ..

- No Display at all .. use HDMI
- Amazon
- Swiss-Version
- ePaper Display

Thanks.

PR #232, opened 2018-12-31
Quote PasswordC in Unlock Command

GeorgeSpark-Stahl:
auto-unlock wasn't working for me and it turns out it was because it had spaces in it.


rootzoll:
good catch :) much thanks

PR #299, opened 2019-01-31
Introduce DietPi compatibility

openoms:
Added DietPi specific parts to the raspbianStretchDesktop.sh to run on an Odroid HC1
Added to README.md and install_On_DietPi.md to describe install process with screenshots and pictures.
Readded an extended function to be able to copy blockchain from a 2nd HDD. It was required because that is  my default process. Extended 10setupBlitz.sh with the option.
Introduced sudo in 70initLND.sh to be able to start scanning with LND.
Removed --externalip=${publicIP} from lnd.service as it resulted to error.
Couple of typos were fixed while reading through.
Tested extensively on the ODroid HC1, did not run on the rPI.
Only known problem: the status screen does not show the PublicIP despite reachable node with 13 peers. Address appears once behind Tor.


openoms:
Next step is to automate the DietPi setup by utilizing the the https://github.com/Fourdee/DietPi/blob/master/dietpi.txt.
The initial OS update process was not smooth with the manual install, so will need to build a couple more times to make sure it does not break.

openoms:
The change by @21isenough fixed the bug with the PublicIP address.
He also successfully built on the ODroid HC1
I will test-build again on the ODroid just to confirm and will also try Dietpi on my RaspberryPi.

rootzoll:
This looks awesome :) Will take some days for me until I find the time to work thru it and test it out. Much thx :D

rootzoll:
@openoms much thanks for this :) 

BTW why did you changed the boostrap service from "one shot" to "simple":
https://stackoverflow.com/questions/39032100/what-is-the-difference-between-systemd-service-type-oneshot-and-simple

I think its much cleaner to have boostsrap as a blocking script - perparing all that is needed and then let the other services start. Was this a problem on your side?

rootzoll:
I the dev branch I moved the build script from /build.sdcard/raspbianStrechDesktop.sh to /build_sdcard.sh

I already updated the main README & FAQ but left all the links under /dietpi in the old structure becaue ithey ppoint zto your fork. If you merge dev/master again - make sure to update those and make a PR again. Thanks.

openoms:
Thanks for merging! I can see you have fixed my the slight confusion with the LCD setup, it works well now.  Just installing fresh on my Raspi (with raspbian) including the latest changes in the dev branch, all good so far.
I think our problem with the bootstrap.service was due to a couple of dirty setup restarts when the _bootstrap.sh was needed to be manually started to avoid errors. I have reverted to Type=oneshot now and there is no issue yet. Will test a bit further and let you know.
I like the new name and path of the sdcard_build.sh, will update and PR to change it in the dietpi folder as well.

PR #318, opened 2019-02-15
minor typo + shortened URL

vnnkl:
URL seems easier to type out


rootzoll:
Thanks. The URL seems also invalid now because I moved that point into the FAQ. I think I will change it into a general hint instead of a hard URL.

Issue #193, opened 2018-12-06
distinguish os


vv01f:
regarding: https://github.com/rootzoll/raspiblitz/blob/adb1ff09e2caf3c000f191210a8a5f1e37f4e8b2/home.admin/97addMobileWalletZap.sh#L51

to distinguish the os, you can switch on env `$OSTYPE` and maybe use `uname` as a fallback (never used a macos).

source: https://stackoverflow.com/questions/394230/how-to-detect-the-os-from-a-bash-script

geco91:
maybe this gets solved by using the LCD for QR code display - see PR
https://github.com/rootzoll/raspiblitz/pull/334
recommend for version 1.1 milestone

nyxnor:
Rootzoll, Im going to issues since the beggining, try to clean them all. If I have close permissions, it wont ping every time.
This is outdated about the qr code, since no lndconnect is being refactored and the web UI. Can close this.

rootzoll:
@nyxnor Thanks for the ping ... I close this now.

How can I give just "Close Permission in Issues" on github? I only see that I can make somebody (like openoms) a "collaborator" but that would give even rights to manage the code (commit, merge PRs, etc) - what would be a bit much:

https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/permission-levels-for-a-user-account-repository#collaborator-access-for-a-repository-owned-by-a-user-account 

I think "an organization" has Triage as access level ... but that would mean i would need to transfer the user account repo to an organizations account.

https://docs.github.com/en/organizations/managing-access-to-your-organizations-repositories/repository-permission-levels-for-an-organization

We cant do this right now because of time ... but its someting for the v1.7.1 we could do. I opened an issue about that: https://github.com/rootzoll/raspiblitz/issues/2184

PR #347, opened 2019-02-23
detect os and print corresponding QR command

thehapax:
if merged this pull would close issue #193 for Linux and OSX

### DIFF
--- a/home.admin/97addMobileWalletZap.sh
+++ b/home.admin/97addMobileWalletZap.sh
@@ -77,7 +77,21 @@ else
   lndconnect --host=${dynDomain}
 fi
 
-echo "(To shrink QR code: OSX->CMD- / LINUX-> CTRL-) Press ENTER when finished."
+
+platform='unknown'
+unamestr=`uname`
+if [[ "$unamestr" == 'Linux' ]]; then
+   platform='linux'
+elif [[ "$unamestr" == 'Darwin' ]]; then
+   platform='Darwin' # mac OSX
+fi
+
+if [[ $platform == 'Linux' ]]; then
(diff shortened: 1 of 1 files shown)


raumi75:
Great idea, but it does not work for me, yet. When I ssh into my RaspiBlitz from macOS, I have to use the macOS keys, but uname returns "Linux". How can we detect the client OS?

thehapax:
@raumi75 This pull request was for one script only. It hasn't been accepted yet. If you want to try it you can test it out on my repository. see issue #193 for code examples on how to detect the client OS

raumi75:
I've tried it and it does not work. 

`uname` returns `Linux` on my Raspiblitz even if I connect from my Mac. That's the expected behavior. If I copy the script to my Mac, it would return Darwin, but that's not a solution to our problem.

Unfortunately I could not find a way to detect the OS of the client within the script. Maybe the `sshd` can log the ssh-client string or something. But after fiddling with it for an hour, I've given up for now.

thehapax:
> Unfortunately I could not find a way to detect the OS of the client within the script. 

I don't believe what you are seeking is possible as it is beyond  the environment any script running in. Good luck if you can find something. 

However, you are incorrect in saying `it does not work.`.  It is doing what it can do within its scope. Its just not doing what you want specifically.

raumi75:
Hi @thehapax, your commit is trying to display the correct key combination for decreasing or increasing the font size of the QR-code via ssh on a Mac/Linux/Windows client, right?

rootzoll:
I merged into dev and will test again as part of the #193 - if its not working the issue will stay open.

PR #379, opened 2019-03-02
Add canadian shopping list

antonleviathan:
We just put together two lightning nodes at [Ledn](ledn.io) using your guide and thought it would be nice to share the shopping list we used on the Canadian version of Amazon (amazon.ca). Let me know if any adjustments are required!

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ These are the community currated shopping lists based on country:
 
 * [Germany](shoppinglist_de.md) *(reference shopping list)*
 * [USA](shoppinglist_usa.md)
+* [Canada](shoppinglist_ca.md)
 * [UK](shoppinglist_uk.md)
 * [Switzerland](shoppinglist_ch.md)
 * [France](shoppinglist_fr.md)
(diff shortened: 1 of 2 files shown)


rootzoll:
Hi @antonleviathan much thanks for the canadian shopping list. I merged it to the master branch for people to see and use. 

But one small thing is bugging me. I had a similar power adapter (small size + switch on/off on cable) and after a while I was seeing "undervoltage warnings" ... even it was advertising 3A, it weems that such small sized adapters have problems in delivering a stable output - also someone remarked that a on/of switch might add an additional problem point. I changed to a really big sized adapter with no on/off plug on cable ... running fine so far. 

So if you can find/test an alternative to that adapter on the canadian amazon feel free make an update to your PR or make a new one.

antonleviathan:
@rootzoll No problemo, and thanks for the followup. I will probably be getting a few more pi over the upcoming months so I'll try out a different adapter. In the meantime I'll monitor these and see how they do.

PR #408, opened 2019-03-07
Added shopping list for Amazon Spain

conejoninja:


rootzoll:
Muchas Gracias :) I changed Power Adapter (to the one I had best stable experience with so far) and the case to provide support for the display. @conejoninja if you can quick check if those are good choices on spain amazon, that would be great.

conejoninja:
Yes, they're ok. I didn't use a case, I choose the cheapest one, in the second image looks like the lidl could be taken off and expose the GPIO pins. The case looks curved, so maybe that's a problem, but didn't check myself.

Power adapter seems ok. Thanks for merging.

PR #410, opened 2019-03-08
minor display changes

fluidvoice:
On my Pi the % was not displaying correctly during syncing.  Also I think during syncing (always really) the CPU load and temp should be displayed.  For now just the temp.


rootzoll:
good catch :) got it merged into dev branch

PR #467, opened 2019-03-28
Update LCD Display shopping link

thelwyn:
The previous one doesn't work with the drivers embedded in Raspiblitz. This one does, purchased and works flawlessly immediately when plugged.


rootzoll:
perfect. much thanks.

PR #481, opened 2019-04-01
Updated broken Amazon.de SD-Card Link

bavarianledger:


rootzoll:
Much thanks - good catch :)

Issue #88, opened 2018-09-10
InfoService for future Version


frennkie:
**Update**
It would be great to have a more flexible way to generate and show the content for the screen (either LCD or ePaper). 
Decoupling this into a "backend" and "frontend" service seems to make sense. This issue is to discuss and develop these ideas further.

**Initial opening comment**
I have IPv6 on my local network and the infoBlitz Screen does not look pretty:

```
              RaspiBlitz v0.86  raspiblitz
              bitcoin Fullnode + Lightning Network + TOR
        ,/    -------------------------------------------
      ,'/     load average: 1.66, 2.12, 1.36, CPU 61°C
    ,' /      Free Mem 561M / 927M  Free HDD 642G (70%)
  ,'  /_____, ssh admin@192.168.178.116 ▼17.1MiB ▲47.2MiB
 .'____    ,'
      /  ,'   bitcoin v0.16.2 mainnet Sync OK (100.00%)
     / ,'     Public 2001:XXXX:XXXX:XXXX:XXXX:XXXX:XXX:XX59
2001:XXXX:XXXX:XXXX:XXXX:XXXX:XXXX:XX37:8333
    /,'
   /'         LND v0.5-beta wallet 124345 sat
              2/3 Channels 222531 sat

03730a3bxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxd9a@123.123.123.123:9735
```

When I have some more time I'll try to create a PR for this.

frennkie:
Also - I have a 3,5" LCD running at `480X320` and it cuts off the last digit of the `Public 2001...:XX59` line.

frennkie:
Hm.. the following line finds my non Onion/Tor address and then thinks I use Tor (which I don't):
```
onionAddress=$(${network}-cli -datadir=${bitcoin_dir} getnetworkinfo | grep '"address"' | cut -d '"' -f4)
``` 

Maybe it would be "safer"/better to parse the JSON returned by `getnetworkinfo` and check whether `["networks"]["name"] == "onion"` and `["networks"]["reachable"] == true` for the Onion/Tor check..?!

rootzoll:
I could be that in the future the display info will be managed with another programming language - parsen JSON would be the better available option then.

frennkie:
Personally I have to say I do like the current *old school* style of the info screen. But I also think it's currently limited regarding functionality.

I think there would be two alternative approaches:

[...]

PR #549, opened 2019-04-27
remove unfinished code snippets

frennkie:
Some month ago I added a python based script - mainly to demo some code for #88. 
So far the code did not prove to be a significant improvement over the great job you have been doing in `bash`. 

Additionally - and unfortunately - I have not been able to keep up with the changes that were made to the inner workings of RaspiBlitz (e.g. raspiblitz.conf file..).

I suggest to remove this *broken* code from the `master` and release (`v1.2`) branches in order to not confuse users with outdated, seemingly duplicate scripts.


rootzoll:
In the long view I think its stil important to get to a infoservice like deamon similar you concepted in #88. But in the short view it showed that there is much prio to focus on home network management, stability and data security of the node first and put time resources there - thats what is v1.2 and also v1.3 will be about.

But while going thru the scripts I already started to create more seperation between UI and config logic what is needed to make the switch to a web UI in the future. I also started to fresh up my python skills (some new scripts are in python) and experimenting with a way to use the existing bash scripts from python scripts. That way existing and tested bash scripts can still be part of the future versions and it will reducing a lot of rewrites and retests. So it will be a an iterative process over a longer time than expected until a switch can be made.

Possibly chaos camp can be a good point in time for that combined with the possobility to sync up in person and plan on how to finally switch to a WebUI/deamon structure. So much thanks for the input so far, keep it up ... we take the long way home ... but we will get there :)
https://www.youtube.com/watch?v=aoEUeTxmYHk

PR #623, opened 2019-05-20
Fixes wrong ANSI colors if "Free HDD" < 10%

bavarianledger:
The characters and next line (lightning ANSI graphic) after "Free HDD" were incorrectly highlighted once "Free HDD" is < 10%.


rootzoll:
much thx :D

PR #639, opened 2019-06-01
Update italian power link

Siimone:
Update italian power link with a better product


rootzoll:
perfect :) much thanks!

PR #693, opened 2019-07-30
Update 10setupBlitz.sh

ejcobi:
Fixed typo that offered one option count more than available. 
"..choose from these 5 options ..." > "..choose from these 4 options.."
"..choose from these 3 options ..." > "..choose from these 2 options.."

Just had to do the setup more than once.


rootzoll:
thx :)

PR #733, opened 2019-08-24
Remove screensaver

frennkie:
By default the screensaver turns the screen black after ~10 minutes.
Often this is not desireable.

This PR should fix it.. tests pending.


rootzoll:
perfect :)

PR #736, opened 2019-08-27
rename logo dirs

frennkie:
I was having problems (on Windows for dev)) when checking out the logo dir (directory 3 has a colon in the name). Would you mind changing this? I searched but did not find any references to the files so merging this PR should not break anything, right...?!


rootzoll:
np. those logos are just there for external use - not part of the code. merged.

PR #772, opened 2019-09-11
add Go to PATH

openoms:
Resolves the issue when `lndconnect -j`  is not available if Go is not freshly installed.

LNDconnect displays this under the QR, causing confusion if the command is not available: 
Screenshot from 2019-09-11 10-15-18


rootzoll:
makes sense - thx :)

PR #775, opened 2019-09-12
Fix restore from channel.backup file to support different chains.

coblee:


rootzoll:
Thanks for checking on that :)

PR #777, opened 2019-09-13
Run LND behind TOR v3 and streamisolation

Tracachang:
bitcoin.conf added 

proxy=127.0.0.1:9050
bind=127.0.0.1
dnsseed=0
dns=0

lnd.services
replaced tor.v2 for .v3
added --tor.streamisolation

### DIFF
--- a/home.admin/config.scripts/internet.tor.sh
+++ b/home.admin/config.scripts/internet.tor.sh
@@ -105,6 +105,10 @@ activateBitcoinOverTOR()
       echo "Addding TOR config ..."
       sudo chmod 777 /home/bitcoin/.${network}/${network}.conf
       echo "onlynet=onion" >> /home/bitcoin/.${network}/${network}.conf
+      echo "proxy=127.0.0.1:9050" >> /home/bitcoin/.${network}/${network}.conf
+      echo "bind=127.0.0.1" >> /home/bitcoin/.${network}/${network}.conf
+      echo "dnsseed=0" >> /home/bitcoin/.${network}/${network}.conf 
+      echo "dns=0" >> /home/bitcoin/.${network}/${network}.conf 
       if [ "${network}" = "bitcoin" ]; then
         # adding some bitcoin onion nodes to connect to to make connection easier
         echo "addnode=fno4aakpl6sg6y47.onion" >> /home/bitcoin/.${network}/${network}.conf
@@ -150,7 +154,7 @@ activateLndOverTOR()
     sudo systemctl disable lnd 2>/dev/null
 
     echo "editing /etc/systemd/system/lnd.service"
-    sudo sed -i "s/^ExecStart=\/usr\/local\/bin\/lnd.*/ExecStart=\/usr\/local\/bin\/lnd --tor\.active --tor\.v2 --listen=127\.0\.0\.1\:9735/g" /etc/systemd/system/lnd.service
+    sudo sed -i "s/^ExecStart=\/usr\/local\/bin\/lnd.*/ExecStart=\/usr\/local\/bin\/lnd --tor\.active --tor\.streamisolation --tor\.v3 --listen=127\.0\.0\.1\:9735/g" /etc/systemd/system/lnd.service
   
     echo "Enable LND again"
     sudo systemctl enable lnd
--- a/home.admin/config.scripts/internet.tor.sh
+++ b/home.admin/config.scripts/internet.tor.sh
@@ -137,6 +137,10 @@ deactivateBitcoinOverTOR()
   echo "*** Changing ${network} Config ***"
   sudo sed -i "s/^onlynet=.*//g" /home/bitcoin/.${network}/${network}.conf
   sudo sed -i "s/^addnode=.*//g" /home/bitcoin/.${network}/${network}.conf
+  sudo sed -i "s/^proxy=.*//g" /home/bitcoin/.${network}/${network}.conf
+  sudo sed -i "s/^bind=.*//g" /home/bitcoin/.${network}/${network}.conf
+  sudo sed -i "s/^dnsseed=.*//g" /home/bitcoin/.${network}/${network}.conf
+  sudo sed -i "s/^dns=.*//g" /home/bitcoin/.${network}/${network}.conf
   sudo sed -i '/^ *$/d' /home/bitcoin/.${network}/${network}.conf
   sudo cp /home/bitcoin/.${network}/${network}.conf /home/admin/.${network}/${network}.conf
   sudo chown admin:admin /home/admin/.${network}/${network}.conf


rootzoll:
Thx. Will merge and test once I start working on v1.4 branch.

openoms:
we also need to add:
`sudo chown -R bitcoin:bitcoin /run/tor`
to the end of the tor setup to avoid this logging problem in Tor:
```
Oct 23 05:49:24 HC2 tor[30287]: Oct 23 05:49:24.000 [warn] /run/tor is not owned by this user (bitcoin, 1002) but by debian-tor (108). Perhaps you are running Tor as the wrong user?
Oct 23 05:49:24 HC2 tor[30287]: Oct 23 05:49:24.000 [warn] Before Tor can create a control socket in "/run/tor/control", the directory "/run/tor" needs to exist, and to be accessible only by the user and group account that is running Tor.  (On some Unix systems, anybody who can list a socket can connect to it, so Tor is being careful.)
```

Or we could consider running Tor with the isolated (and default) user `debian-tor`.

openoms:
I had some problems running tor v3:
two peers running Tor V2 RaspiBlitzes could not connect.
On my side:
```
$ lncli connect URI@Tor.onion:9735
[lncli] rpc error: code = Unknown desc = socks connect tcp 127.0.0.1:9050->jTor.onion:9735: unknown error TTL expired
```

Switching back to Tor v2 sorted the connections.

To test the lnd settings in this PR:

`sudo nano /etc/systemd/system/lnd.service`  
Comment out:
`#ExecStart=/usr/local/bin/lnd --tor.active --tor.v2 --listen=127.0.0.1:9735`

and use instead:
`ExecStart=/usr/local/bin/lnd --tor.active --tor.streamisolation --tor.v3 --listen=127.0.0.1:9735` 
`sudo systemctl daemon-reload` 
`sudo systemctl restart lnd`  

`lncli getinfo`  
will show the long Tor address

openoms (changes requested):
If we are adding entries to the bitcoin.conf they will need to be removed too if Tor is deactivated.

openoms:
@Tracachang 
can you give some reasoning for these:
```
bitcoin.conf added

proxy=127.0.0.1:9050
bind=127.0.0.1
dnsseed=0
dns=0
```

Would not be it the case with 
```
dnsseed=0
dns=0
```
that there would be no new connections made to Tor nodes apart from the ones in the bitcoin.conf list already. How is that desirable?

Do these
```
proxy=127.0.0.1:9050
bind=127.0.0.1
```
add  anything, when `onlynet=onion` is set?

Tracachang:
@openoms 
> that there would be no new connections made to Tor nodes apart from the ones in the bitcoin.conf list already. How is that desirable?

Yes there will be new connections apart from the fallback nodes added. You just not get them from the DNS server for new peers instead you fetch peers from fallback nodes. If you do not trust those fallbacks nodes you just can change them, but I would better trust to get peers from a dev core developper than from public DNS server and also as far as I read this option was set to avoid sybil attacks.

Also can read more here

https://stadicus.github.io/RaspiBolt/raspibolt_69_tor.html

> DNS lookups for other peers could potentially be used to deanonymize you, at least it happened in the past. Some people might want to deactivate DNS request that are usually used to find other nodes on the network.

With this configuration, your node is not capable to find peers on its own. That’s why it is necessary to bootstrap it with a hardcoded list of a few nodes to contact on startup by specifying addnode.


> add anything, when `onlynet=onion` is set?

If you do not set onlynet=onion, since you have set proxy and bind all traffic wil go throught tor but you will connect to clearnet and onion peers. Setting onlynet=onion will connect you only with onion peers.

Tracachang:
@openoms 
From stadicus guide:

> Deactivate DNS lookup:

DNS lookups for other peers **could potentially be used to deanonymize you, at least it happened in the past**. Some people might want to deactivate DNS request that are usually used to find other nodes on the network.

With this configuration, your node is not capable to find peers on its own. That’s why it is necessary to bootstrap it with a hardcoded list of a few nodes to contact on startup by specifying addnode. 
https://stadicus.github.io/RaspiBolt/raspibolt_69_tor.html

openoms:
> > that there would be no new connections made to Tor nodes apart from the ones in the bitcoin.conf list already. How is that desirable?
> 
> Yes there will be new connections apart from the fallback nodes added. You just not fetch the list from the DNS server for new peers instead you fetch peers from fallback nodes. If you do not trust those fallbacks server you just can change them, but I would better trust to get peers from a dev core developper than from public DNS server and also as far as I read this option was set to avoid sybil attacks.

So new peers will connect, but they are taken from the peer list of the fallback onion nodes? 
That might be ok, but I am not sure if we should standardise the connections this way.

The `dnsseed=0` would stop the RaspiBlitz advertising it`s peers. That would be also against the principle of decentralisation.  Do you have a link of a discussion about of the drawback of these options vs protection from a sybil attack?


> > add anything, when `onlynet=onion` is set?
> 
> If you do not set onlynet=onion, since you have set proxy and bind all traffic wil go throught tor but you will connect to clearnet and onion peers. Setting onlynet=onion will connect you only with onion peers.

That makes sense, but since we use the `onlynet=onion` when Tor is on, no communication is done outside of Tor anyway. That's what I meant about this setting not adding much. 
We could set the Tor proxy and bind, but then would make sense to switch the `onlynet=onion` off to benefit from connections to both (clearnet and Tor) type peers.

Tracachang:
> So new peers will connect, but they are taken from the peer list of the fallback onion nodes?
That might be ok, but I am not sure if we should standardise the connections this way.

Yes and I understand your concerns, however this is and avanced option, maybe could it be choosen when editing  "tor enabled script" like advanced options?

> The `dnsseed=0` would stop the RaspiBlitz advertising it`s peers. That would be also against the principle of decentralisation. Do you have a link of a discussion about of the drawback of these options vs protection from a sybil attack?
> 

I do not understand why you would keep from (dns and dnseed) one ON are another OFF it would not bring any benefit, keeping both ON or both OFF makes sense. I've been searching but I did not find that discussion, ever read long ago... just found the quote I've copied before : 

"DNS lookups for other peers could potentially be used to deanonymize you, at least it happened in the past."

> That makes sense, but since we use the `onlynet=onion` when Tor is on, no communication is done outside of Tor anyway. That's what I meant about this setting not adding much.
We could set the Tor proxy and bind, but then would make sense to switch the `onlynet=onion` off to benefit from connections to both (clearnet and Tor) type peers.

onlynet=tor  avoids having to pass throught a tor exit node & better speed. Can it be also safer?

But I do agree that some users may want to connect both clearnet and Tor (even if they will not have inbound connections from clearnet peers), maybe as mentioned with dns (adding an option when enabling tor script).

openoms:
Ok, agree with maximum privacy by default when Tor is activated. More choices would only confuse the users being exposed to Tor the first time and advanced users can edit their bitcoin.conf any time.

Now only need to make the script remove the new Tor related entries when Tor is switched off to have this merged.

Tracachang:
> Now only need to make the script remove the new Tor related entries when Tor is switched off to have this merged.

done

openoms (approved):
LGTM
I have tested these settings with my node before without any problems.

openoms:
pasted this to my Tor node to test:
```
network=bitcoin      
sudo chmod 777 /home/bitcoin/.${network}/${network}.conf

echo "proxy=127.0.0.1:9050" >> /home/bitcoin/.${network}/${network}.conf
echo "bind=127.0.0.1" >> /home/bitcoin/.${network}/${network}.conf
echo "dnsseed=0" >> /home/bitcoin/.${network}/${network}.conf 
echo "dns=0" >> /home/bitcoin/.${network}/${network}.conf 

sudo chmod 444 /home/bitcoin/.${network}/${network}.conf
# copy new bitcoin.conf to admin user for cli access
sudo cp /home/bitcoin/.${network}/${network}.conf /home/admin/.${network}/${network}.conf
sudo chown admin:admin /home/admin/.${network}/${network}.conf

sudo sed -i "s/^ExecStart=\/usr\/local\/bin\/lnd.*/ExecStart=\/usr\/local\/bin\/lnd --tor\.active --tor\.streamisolation --tor\.v3 --listen=127\.0\.0\.1\:9735/g" /etc/systemd/system/lnd.service
sudo systemctl daemon-reload
sudo systemctl restart lnd
lncli unlock
```

Running fine.

PR #795, opened 2019-09-22
add tmux

### LINKED ISSUE #793: Feature Request - add tmux
I have seen that the sd card image includes the `screen` tool. I am more used to `tmux` and it appears to be much more popular than `screen` these days. 

Of course I can simply install it myself after every update but it would be nice if this could be included directly. I'm already working on a PR for this. @rootzoll against which branch should I file the PR?

Here is a preview:
raspi_tmux

frennkie:
Fix #793

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -154,7 +154,7 @@ if [ "${baseImage}" = "ubuntu" ] || [ "${baseImage}" = "armbian" ]; then
   sudo adduser pi sudo
 fi
 
-# special prepare when Nvidia Jetson Nano 
+# special prepare when Nvidia Jetson Nano
 if [ ${isNvidia} -eq 1 ] ; then
   # disable GUI on boot
   sudo systemctl set-default multi-user.target
@@ -407,7 +407,7 @@ if [ ${isAARCH64} -eq 1 ] ; then
 fi
 if [ ${isX86_64} -eq 1 ] ; then
   bitcoinOSversion="x86_64-linux-gnu"
-fi 
+fi
 if [ ${isX86_32} -eq 1 ] ; then
   bitcoinOSversion="i686-pc-linux-gnu"
 fi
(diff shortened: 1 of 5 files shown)


build_sdcard.sh
```diff
@@ -594,6 +594,10 @@ sudo apt-get -y install cpulimit
 # for background downloading
 sudo apt-get -y install screen
 
+# for multiple (detachable/background) sessions when using SSH
+sudo apt-get -y install tmux
+sudo -u admin git clone https://github.com/gpakosz/.tmux.git /home/admin/.tmux
```
rootzoll: Is there a way to avoid the external dependency on another github? When its just about one or tweo config files we use from that repo ... can we copy it over to the raspiblitz repo?

frennkie: I agree that this could also be a simple download instead of a git clone. Although a download will still be a dependency, right? Of course we could simple use the current version of that file and never change/update.. but then we would miss out on additional features and bug fixes that are done on that project... 

I'm fine either way.

frennkie:
@rootzoll Any objections..?! You mentioned you would like to have a quick intro to tmux.. right?!

openoms:
Doing a bit of shilling of Tmux: https://github.com/openoms/bitcoin-tutorials/blob/master/joinmarket/README.md#keep-the-offers-running-in-the-background-with-tmux
It is an awesome tool, I am in full support of integrating it.

fluidvoice:
> Doing a bit of shilling of Tmux: https://github.com/openoms/bitcoin-tutorials/blob/master/joinmarket/README.md#keep-the-offers-running-in-the-background-with-tmux
> It is an awesome tool, I am in full support of integrating it.

I use [Terminator](https://gnometerminator.blogspot.com/p/introduction.html) but have not used it via SSH and don't know if it can save/reopen sessons. That's a nice feature.

PR #801, opened 2019-09-28
Stresstest stoped to early on PI4

Emzy:
The sysbench parameter it to low for the fast Raspberry PI 4 "--cpu-max-prime=10000"
It stopes before the 60 seconds are run out.
Increased to it to 100000


frennkie:
What's the effect/impact on RPi3 devices..?

PR #811, opened 2019-10-06
add option for duration of delay

frennkie:
I would like to be able to bypass the **PAUSE** to `00infoLCD.sh`. @rootzoll If I understand correctly the main reason for the **PAUSE** is that a freshly booted RaspiBlitz should wait a bit before trying to display the dashboard as a lot of stuff needs to be started in the beginning.

From the GUI I would call/restart `00infoLCD.sh` at later times and then the **PAUSE** does not make sense, right?

Change is backwards compatible (so not giving -p N will default to the current value of `12`).


frennkie:
@openoms let me know if you find issue with this one.. ;-)

PR #828, opened 2019-10-19
Add custom pause

frennkie:
The pause on the infoLCD screen is not always needed. This adds a parameter to be able to skip the delay.


frennkie:
@openoms could you also please test this from your end...?

frennkie:
damn it.. duplicate of #811

PR #864, opened 2019-11-16
recovery sheet added

LightRider5:


openoms:
Great idea!

rootzoll:
This is great :) Much thx!

rootzoll:
Hey @animalmother1988 the PDF is great, but is there also a source file that can be included so that it can be edited for future changes?

PR #880, opened 2019-11-25
Update CH shoppinglist

0xjacobb:
Added sources for Switzerland. 1TB SSD is the same price as 500GB. Therefore 500GB was deleted. In addition, a display version is chosen which is available in CH. It should work but is not tested yet.


rootzoll:
Ah OK... there it is already :) thx

PR #906, opened 2019-12-14
Add Lightning Loop Service

openoms:
Based on top of https://github.com/rootzoll/raspiblitz/pull/904

Add Lightning Loop Service install script with systemd service for loopd


rootzoll:
thx :)

PR #907, opened 2019-12-15
Update CH shopping list

0xjacobb:
Added new resources for working and tested SSD (from German list). Actually it was only a HDD in the CH shopping list. Additionally, I've added a comment to the pi-shop.ch display.


rootzoll:
Thx :)

PR #915, opened 2019-12-17
services: add lndmanage

openoms:
added links to the readme-s to display on install for lnmanage and loop


openoms:
Sorry, fat finger, wanted to leave @rootzoll to merge.

Not reverting as there are working additons only.

Issue #929, opened 2019-12-24
RTL error on restore (v1.4 manual build) 


openoms:
RTL keeps failing on restore when my blitz is started with the manual sd card build from v1.4 

https://pastebin.com/99t8vzYi

https://termbin.com/r5zs

Seems to be a downoad failure:
```
*** Run: npm install ***
 
> node-sass@4.12.0 install /home/admin/RTL/node_modules/node-sass
> node scripts/install.js
 
Downloading binary from https://github.com/sass/node-sass/releases/download/v4.12.0/linux-arm-64_binding.node
Cannot download "https://github.com/sass/node-sass/releases/download/v4.12.0/linux-arm-64_binding.node":
 
HTTP error 404 Not Found
```

needs further testing.

Was able to progress with the bootstrap once ran:
`./config.scripts/bonus.rtl.sh off`

openoms:
This is fixed in #931, successfully tested on the RC1 image too.

Issue #454, opened 2019-03-21
Add the Loop service to RaspiBlitz


rootzoll:
Lightning Labs released the Aplha of Loop:
https://github.com/lightninglabs/loop

For example running Loop on the RaspBlitz it could be possible to monitor the channel balances in the background and automatically send funds to on-chain wallets (hardware wallet) to regain inbound capacity and store earnings in a cold-wallet.

At the moment its very experimental and you would need to replace LND on the raspiblitz with a fresh self-compiled version from the LND Github master. And the loopd services needs to be installed. 

Interesting to see would be:
- how much resources does the loopd service consumes on the RaspiBlitz
- what config parameter would a automatic loop-out script need
...

So if someone is getting loopd runnign with LND on a RaspiBlitz, please report/share your experiences.

-

openoms:
AFAIK you will need to build LND from source to be able to test this feature until the new release is out.

rootzoll:
I heard that LND 0.6 release is just around the corner. Shortly after that the raspiblitz v1.2 will be out - containing the LND 0.6 and I think then everybody can start playing/researching on how to best integrate it to the RaspiBlitz.

openoms:
Started experimenting with the Loop out service.
The installation script is functional:
https://github.com/openoms/raspiblitz-extras#loop-service

Started this new repo to help testing new features in a compatible way but as a separate extension package. Will aim to include the other extra features too.

openoms:
My first successful Loop out (could specify a cold-storage address, but did not so funds returned to my LND wallet on-chain):
Note the max predicted fee of 25147 and final fee of 4XXX sats. Still looking forward to be able to control the miner fees: https://github.com/lightninglabs/loop/issues/36

`$ loop out --channel XXXXXXXXXXXXX --amt 500000`                                                                                                                                 
```
Max swap fees for 500000 Loop Out: 25147                                                                                                                                                                    
CONTINUE SWAP? (y/n), expand fee detail (x): x                                                                                                                                                              
                                                                                                                                                                                                            
Max on-chain fee:                 13851                                                                                                                                                                     
Max off-chain swap routing fee:   10010                                                                                                                                                                     
Max off-chain prepay routing fee: 36                                                                                                                                                                        
Max swap fee:                     1250                                                                                                                                                                      
Max no show penalty:              1337                                                                                                                                                                      
CONTINUE SWAP? (y/n): y                                                                                                                                                                                     
[...]

PR #931, opened 2019-12-24
Fix bootstrap.provision when all services are on

openoms:
* fixing nodeJS download path 
* install nodeJS services with admin https://github.com/rootzoll/raspiblitz/issues/929

* start services outside of install script to avoid provision getting stuck https://github.com/rootzoll/raspiblitz/issues/454

* Loop install needs restart only if GOPATH is not available yet 

- [x] test if provision runs without errors with all services on

### DIFF
--- a/home.admin/00settingsMenuServices.sh
+++ b/home.admin/00settingsMenuServices.sh
@@ -128,18 +128,20 @@ if [ "${loop}" != "${choice}" ]; then
   errorOnInstall=$?
   if [ "${choice}" =  "on" ]; then
     if [ ${errorOnInstall} -eq 0 ]; then
-      # add Go vars to current session
+      sudo systemctl start loopd
       if [ ${#GOPATH} -eq 0 ]; then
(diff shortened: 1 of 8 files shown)


openoms:
The services get stopped during bootstrap provision because they are waiting fro either bitcoind or lnd to run.

RTL is waiting for LND
BTC-RPC-Explorer is waiting for bitcoind 
Loop is waiting for LND
Electrs is waiting for the bitcoind.service 

Solution is to remove the `systemctl start service` entries from the install scripts.

openoms:
Provision now restores everything without stopping, but takes a long time ~1 h to compile everything (expected with RTL, BTC-RPC-Explorer and ElectRS).

Running another round of test starting from the SDcard build.

openoms:
OK, last test is successful.
RTL, BTC-RPC-Explorer, ElectRS and Loop restores correctly with a new SD.

Should be good for 1.4 RC1.

PR #948, opened 2020-01-02
update txindex status line

frennkie:
Change 

```
BTC-RPC-Explorer: Indexing 476261/610976 (please wait)
```

to 

```
BTC-RPC-Explorer: Indexing is at 77.95% (please wait)
```


frennkie:
@openoms what do you think?

PR #961, opened 2020-01-06
Fix lndmanage

frennkie:
The `cd lndmanage` caused an issue during my restore.. I assume the directory had not been reset from a previous script.

### DIFF
--- a/home.admin/config.scripts/bonus.lndmanage.sh
+++ b/home.admin/config.scripts/bonus.lndmanage.sh
@@ -16,12 +16,18 @@ fi
 
 # install
 if [ "$1" = "1" ] || [ "$1" = "on" ]; then
+
+  if [ -d "/home/admin/lndmanage" ]; then
+    echo "LNDMANAGE already installed"
+    exit 1
+  fi
+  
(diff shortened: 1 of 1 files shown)


openoms:
this happened to me in couple of scripts as I was tetsing and running the scripts with the `admin` user from the `00settingMenu.sh` but the restore script is run with sudo.

Other way is to make sure it the bonus.lndmanage.sh is run everywhere with `sudo -u admin`.

Thanks for fixing!

rootzoll:
With the fixes its now not working anymore when I try to install from service menu. I will change back until its working and try the ´sudo -u admin´ in the provision.

rootzoll:
OK. Install from service menu is now working again ... but running lndmanage I get this error - could this be python3 related?
Bildschirmfoto 2020-01-20 um 23 19 26

rootzoll:
Trying if lndmanage is working during update/recovery now.

rootzoll:
OK now LNDMANAGE is also getting installed during recovery. But still when running LNDMANAGE I get that error like above.

PR #962, opened 2020-01-07
add check to background for Tor

### LINKED ISSUE #894: _background.sh should be changed for tor users
Hello

Checking the _background.sh in home.admin folder I see it requests your public IP every 15 minutes, using an external service with the command `curl -s http://v4.ipv6-test.com/api/myip.php`

Since the idea to use bitcoind and lnd with tor is to hide your IP, there is no interest in requesting your real IP every 15 minutes and could even harm privacy in the worst scenario.

The idea of this script it is to restart lnd if public IP has changed, but even user is not using tor is it really worth to check it every 15 minutes? 

For those who have static IP there is no need and those who has dynamic IP they will probably use a DNS service in order to have a static IP (if they want other lnd peers to be able to open channels to them).

Let me know your thoughts.

frennkie:
When `runBehindTor` is set the "Public IP Check" is bad for privacy and most likely useless (at least as far as I see it - correct me if I'm wrong.. :-) ).

Fix #894

### DIFF
--- a/home.admin/_background.sh
+++ b/home.admin/_background.sh
@@ -91,6 +91,12 @@ do
   if [ ${#lndAddress} -gt 3 ]; then
     recheckPublicIP=0
   fi
+
+  # prevent also when runBehindTor is on
+  if [ "${runBehindTor}" = "1" ] || [ "${runBehindTor}" = "on" ]; then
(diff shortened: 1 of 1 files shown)


rootzoll:
Good Catch :D

PR #970, opened 2020-01-11
add check and install for python2

frennkie:
RTL currently still depends on Python2. Make sure that it installed.


openoms:
is this necessary? I did not need to install python2 on 1.4RC2 to have RTL 0.6.0 running.

frennkie:
I guess it would be more 'friendly' to only remove Python2 if it was only just installed by this script.

PR #978, opened 2020-01-16
Disable Bluetooth

CandleHater:
Disabling Bluetooth offers more security and also brings a small performance improvement. As Bluetooth is anyway not actively used on a RaspiBlitz, it's services and packages can be removed. Also the Bluetooth module is disabled in ``/boot/config.txt``.

I only tested this using Rasbian and therefore restricted this step to that distribution. Someone else could change it accordingly to serve other distributions.


rootzoll:
thx :D

Issue #214, opened 2018-12-25
Add BTCPayServer as service option


rootzoll:
See this great base work from @normandmickey
https://gist.github.com/normandmickey/6d6e74ebb6b3ddbd317bc1450f48f08e

Test and convert into a 'config.script' for RaspiBlitz.

chacham18:
Nice, will try it next week

rektlesspinko:
I followed the instructions and got it working. But for some reason can't integrate joule with it, don't see the LND services listed in crypto services in BTCPay / server settings / services so can't get the QR details to link it to joule. Be interested if you get the same issue @gabrieldov

DaReaLDeviL:
Is this still a thing? A install script with in the "services" menu whould be awesome!

normandmickey:
Yes, I haven't had any time to try to script it.  But if you can follow the
instructions they should work.

rootzoll:
@normandmickey let me know if you have a basic script ready .. I can then integrate it into the rest of RaspiBlitz.

normandmickey:
Updated Gist to install BTCPayServer on Raspiblitz V1.3 for RPI4B
https://gist.github.com/normandmickey/3f10fc077d15345fb469034e3697d0d0

Appears to work but would consider it a WIP.  Would still like to "scriptify" so it could be added as a menu option.  RPI4B improved performance over RPI3 makes this a viable option now.

openoms:
Made the script, but be aware that it is experimental, recommended only for test nodes first. Got it working once already, but SSL certificate asks are limited (5 in an hour - https://letsencrypt.org/docs/rate-limits/) so need to give it a break.
Nginx is very fussy, the setup tends to break previous Nginx configs (Like Electrs SSL), so added a warning if previous nginx.conf is found.
Script: https://raw.githubusercontent.com/openoms/bitcoin-tutorials/master/BTCPayServer/btcpay_to_blitz.sh

to download and run: 
`wget https://raw.githubusercontent.com/openoms/bitcoin-tutorials/master/BTCPayServer/btcpay_to_blitz.sh && bash btcpay_to_blitz.sh`

There is a readme too for the manual process based on @normandmickey`s gist:
https://github.com/openoms/bitcoin-tutorials/blob/master/BTCPayServer/BTCPayServer_on_the_RaspiBlitz.md

By the way for debugging  Nginx found:
`sudo nginx -t`
very useful. Runs a test and gets detailed info on which line is problematic.

[...]

PR #992, opened 2020-01-21
BTCPay fix and update

openoms:
Major update of the whole script bringing BTCpay and NBXplorer up to the latest and updating .NET to 3.1

Fix many path related issues causing install troubles: https://github.com/rootzoll/raspiblitz/issues/214


frennkie:
@openoms I have a question regarding the handling of the admin.macaroon file. I line [280](https://github.com/rootzoll/raspiblitz/blob/master/home.admin/config.scripts/bonus.btcpayserver.sh#L280) the file gets copied over to the `btcpay` user's home. Is there any code/mechanism to handle a case where a user changes the macaroons (by deleting `macaroons.db`)?

openoms:
@frennkie you are right, I need to add the line to lnd.export.sh to copy the macaroon over for btcpay if reset and restart BTCpay.
(This implies that the internal LN node is used, which was not the case for me so far)

PR #997, opened 2020-01-24
Fixed misspelled sudo command

gottfriedsz:
replaced sudp with sudo


rootzoll:
good catch - thanks :)

PR #1005, opened 2020-01-28
btcpay:improve guidance on install

openoms:
Found that the http connection was not necessary when using a self signed certificate, therefore the UFW is not opened for 23001 in that case.
BTCPay serves on the port 23000 which is encrypted and forwarded by Nginx to 443 (HTTPS).

Forwarding the outside port 80 to 23001 (where also nginx is listening for HTTP connections) is unfortunately not helping certbot.
It is necessary for certbot to listen on port 80 to be able to generate a signed certificate for a legit HTTPS connection (web server):
https://certbot.eff.org/faq/#can-i-issue-a-certificate-if-my-webserver-doesn-t-listen-on-port-80

Will need to be able to use port 80 for BTCpayServer to be fully functional. 
An alternative solution could be a remote SSH tunnel to a VPS, but that is far too advanced.


rootzoll:
When we just start with TOR for the v1.4 - do we still need to have port 80 & a self signed cert in the mix?

openoms:
The port 80 is not needed at all, so no certbot needs to be installed either.

The self signed cert generated by openSSL and  is not port dependent, only used by nginx to make a HTTPS connection. It is used for ElectrRS too.

It might be still ok, to keep a LAN connection with the self signed cert as an option. At least the management of BTCPay can be done there and there is no need to.send it round via Tor on.the LAN.

openoms:
Adapted this so there is no conflict.
It is only to provide some improved guidance on port forwarding.

PR #1025, opened 2020-02-12
universal shopping list for all countries

LightRider5:
Universal Shopping List which redirects to the specific amazon page


rootzoll:
I first merged but then decided (not that it is one file) I copy the list and place it directly into the main README. SO I will delete now this file and the old shopping lists. Thx.

PR #1028, opened 2020-02-12
split nginx and the cert script

geco91:


rootzoll:
Will take care yout nginx and certs in the v1.5 with a fresh start ... closing issue.

PR #1035, opened 2020-02-16
Update Funding and Opening channel sections

bitcoinheiro:
Add disclaimers about the public nature of the lightning channels and some tips on how to achieve increased privacy.

Maybe you want to leave it simpler, but I believe this is important since some noobs have the assumption they are shielded by the LN when in reality public channel points show everyone their TXs and can link all their UTXOs under 1 entity. If not careful, they can expose a lot of information to chain analysis crossing the info.

I tried to leave it short and simple, just so noobs think about it and consider the option of coinjoining and opening private channels instead of going with the public default


rootzoll:
Thanks :)

PR #1045, opened 2020-02-20
Fix bitcoin core version number

gcomte:
0.19.1 will be out soon, but not quite yet ;-)


kilrau:
Maybe also fix the litecoin core version number to `0.17.1` (https://github.com/rootzoll/raspiblitz/blob/master/FAQ.md#whats-new-in-version-14-of-raspiblitz)

rootzoll:
Got to excited there ;) Thanks

PR #1050, opened 2020-02-23
Fix typos in initHDD, migration, and LND rescue scripts

dergigi:
Found two trivial typos while upgrading to 1.4. Turns out they were in multiple places.


rootzoll:
Thx :)

PR #1069, opened 2020-03-09
Adding Specter-Desktop to Raspiblitz

k9ert:
As described in #1059 :
* Installation of specter-desktop via services like the others in a virtualenv + pip install 
* cryptoadvance.specter (in home/admin/.specter/.env)
* On top it creates a self-signed-cert and exposes it via https (on port 25441) 
also it creates a Tor-hidden-service. 
* Authentication is enabled by default (Pin of "Password B")
* the installation-procedure is also creating udev-rules in /etc/udev/rules.d for all major hardware-wallets as described in https://github.com/bitcoin-core/HWI/blob/master/hwilib/udev/README.md
* deinstallation-procedure is asking the user whether to keep the data

Issue #1059, opened 2020-02-29
RESEARCH: Adding Specter-Desktop to RaspiBlitz


rootzoll:
_The goal of this project is to make a convenient and user-friendly GUI around Bitcoin Core with a focus on multisignature setup with airgapped hardware wallets._

See Demo:
https://twitter.com/CryptoAdvance/status/1233833767283941376?s=20

See code:
https://github.com/cryptoadvance/specter-desktop

stepansnigirev:
@k9ert is working on it :)

k9ert:
I've created a branch "specter" in my fork here:
https://github.com/k9ert/raspiblitz/tree/specter
which is currently this commit:
https://github.com/k9ert/raspiblitz/commit/6ebf1f5fd57e0a139d726a5911404826e24c5d16

The current state enables:
* Installation of specter-desktop via services like the others in a virtualenv + pip install cryptoadvance.specter (in home/admin/.specter/.env)
* On top it creates a self-signed-cert and exposes it via https (on port 25441)
* also it creates a Tor-hidden-service. 
* Authentication is enabled by default (Pin of "Password B")
![image](https://user-images.githubusercontent.com/117085/76194903-75ce1080-61e7-11ea-8d43-40c8f2adb6dc.png)

There are three things which are missing, still:
* The current deinstallation-procedure is deleting the whole /home/admin/.specter directory. That's not very good for updates. I want that to be optional. The user needs to be asked whether that directory should be deleted.
* The tor-service doesn't allow the use of the camera for security reasons. More research needed but a minor anyway for now.
* HWI doesn't work on the raspiblitz yet. This is really something we should get working as not many people have a specter yet and a multisig-setup would need it anyway.

Any hints on how to get the last point working very much appreciated.

rootzoll:
Cool .. I am just resetting/refreshing my development node and will test :)

rootzoll:
Merged and in testing for v1.5 release :)

rootzoll:
@k9ert got the following expection after install:

```
Mar 18 20:33:17 raspberrypi python3[31788]: Traceback (most recent call last):
Mar 18 20:33:17 raspberrypi python3[31788]:   File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
Mar 18 20:33:17 raspberrypi python3[31788]:     "__main__", mod_spec)
[...]

PR #1075, opened 2020-03-16
 #1059 adding specter-desktop

k9ert:
Second try after #1069 was somehow screwed up.

As described in #1059 :

- Installation of specter-desktop via services like the others in a virtualenv + pip install
cryptoadvance.specter (in home/admin/.specter/.env)
- On top it creates a self-signed-cert and exposes it via https (on port 25441) also it creates a Tor-hidden-service.
- Authentication is enabled by default (Pin of "Password B")
- the installation-procedure is also creating udev-rules in /etc/udev/rules.d for all major hardware-wallets as described in https://github.com/bitcoin-core/HWI/blob/master/hwilib/udev/README.md
- deinstallation-procedure is asking the user whether to keep the data

### DIFF
--- a/home.admin/00mainMenu.sh
+++ b/home.admin/00mainMenu.sh
@@ -57,6 +57,9 @@ fi
 if [ "${loop}" == "on" ]; then
   OPTIONS+=(LOOP "Loop In/Out Service")  
 fi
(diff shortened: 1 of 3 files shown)


rootzoll:
Thanks - merged into the new v1.5 branch :)

Issue #842, opened 2019-10-30
Joinmarket on the RaspiBlitz


openoms:
Started a project to run JoinMarket on the RaspiBlitz: https://github.com/openoms/bitcoin-tutorials/blob/master/joinmarket/README.md
It is largely python command-line based so it is only a guide for now.

Suggestions and additions are welcome.

woeisme:
Sounds awesome!

rootzoll:
@CandleHater is also big on that topic - can you please report your experience here, so that we can turn this into a feature?

Tracachang:
It is a great idea to implement joinmarket, I've been following your guide and it is well explained and really easy to follow. Being implemented in raspiblitz will bring lots of new users which may feel uncomfortable using terminal so they just use samourai or wasabi.  Are you planning to implement also yield generator?

openoms:
I don`t know what else we can do on top of an automated install script. Joinmarket is awesome, but is mostly controlled via the command-line.

The current QT-GUI is made to be a standalone desktop application, cannot be separated from the backend and does not support the Yield Generator itself.

I think the realistic route here is to make the Joinmarket installation an option in the SERVICE menu  and link to the documentation about usage.

Tracachang:
I do think making a script like the one in the menu (what appears when log in using ssh would be the best) now for example has: FUNDING, SEND, RECEIVE... SERVICES (there add COINJOIN) and inside coinjoin create 3 or 4 tabs (if tumbler is included): 

> FUND or RECEIVE: when hit enter executes `(jmvenv) $ python wallet-tool.py wallet.jmdat history` with a message saying to fund (external address only, one or many addresses but always ext). 

> SEND (single coinjoin payment with default options): when hit enter executes `(jmvenv)$ python sendpayment.py wallet.jmdat` which prompt user to eneter amount and address since the full command it is  like: 
`(jmvenv)$ python sendpayment.py wallet.jmdat 5000000 mprGzBA9rQk82Ly41TsmpQGa8UPpZb2w8c`

> Yield Generator: When hit enters it will execute in tmux (jmvenv) $ python yield-generator-basic.py wallet.jmdat (ON and OFF button). with a message saying that before turn ON be sure your tx have at least 5 conf.

I would also add some text saying that a GUI is possible using Joinmarket-QT (pointing to your github of bitcoin tutorials and adding just some info about it, like after install in desktop  edit the joinmarket.cfg with raspiblitz IP + password B etc) 

[...]

PR #1093, opened 2020-03-24
joinmarket: autostart jmvenv

openoms:


openoms:
Fixes starting joinmarket (https://github.com/rootzoll/raspiblitz/issues/842#issuecomment-603512045):
* activates the jmvenv and switches to the scripts directory
* leaves the port 6697 in joinmarket.cfg

PR #1102, opened 2020-04-01
build_sdcard: remove MagPi pdf from Raspbian

openoms:
Remove the MagPi magazine from the RaspiBlitz image :grin:


rootzoll:
I already build the base sd card for the v1.5 release ... if this is the only change in the build script until the final v1.5 release, I would like to push it this PR to the v1.6 release. Background: If possible I like to keep the base sd card build from a Release Cadidate so that its well tested for the final release und just update the on-top scripts for the final release.

openoms:
The MagPi could still be removed from the next release with:
`sudo rm -rf /home/pi/MagPi `
when updating the scripts.
It bloats with 30+ MB.

rootzoll:
Now that we for next LND release will build a new sd card anyway .. I will put in in v1.5

PR #1108, opened 2020-04-04
bugfix #1059 necessary for cryptography python package

k9ert:
It's unknown why cryptography-pakage get installed, though. But i verified:
* a pip install --upgrade cryptoadvance.specter is indeed installaing that package
* a pipdeptree | less doesn't list that dependency.


k9ert:
It's pyOpenSSL which needs cryptography. Not sure why this slipped through in the past. Maybe because of the order of stuff got installed. Probably some other bonus-package is installing it as well and when i did the upgrade, for some reason specter has been installed before that other bonus-package.

rootzoll:
Much thanks :)

PR #1125, opened 2020-04-17
skip LND config check

frennkie:
The Python included  "configuration checker" I used in BlitzTUI does not accept multiple lines with the same settings. As we are not using LND config information for anything at the moment it's safe to simply remove this.


frennkie:
@openoms removed it from both BlitzTUI and the CLI Python config check.. should be ready to merge.

PR #1132, opened 2020-04-22
lnbits: update branding and installation script

eillarra:
Hi @rootzoll 

The PR includes some changes to the LNbits installations script and other minor stuff:

- All references to LNBits have been changed to LNbits (branding).
- The installation script has been updated:
  - only the `.env` file is updated now (no need to mess around with the Python code)
  - Instead of using the LND API, it uses the LND gRPC now. `LND_GRPC_ENDPOINT`, `LND_GRPC_PORT` and `LND_CERT` need to be checked
  - no development packages are installed with PIpenv (no `--dev` flag)
  - `gunicorn` runs with `gevent`now
  - the code is pulled from the 0.1.0 tag

There is one thing that we might need to leave as it was, for backwards compatibility. I have changed the HDD data folder to `/mnt/hdd/app-data/lnbits` (lowercase), because it seems more appropriate for a folder, but maybe it messes up the few installation already out there, so you might want to keep `LNBits`.

Another thing you might want to change is how the code is pulled. We could also have a "raspiblitz" tag, that moves, or maybe users can choose to install another version during installation... or master. Or maybe have an option to update to a new version? I don't know what the best option is...

Cheers.

PR #1135, opened 2020-04-24
lnbits: missing parts

eillarra:
These are the missing changes.

### DIFF
--- a/home.admin/config.scripts/bonus.lnbits.sh
+++ b/home.admin/config.scripts/bonus.lnbits.sh
@@ -83,7 +83,7 @@ if [ "$1" = "write-macaroons" ]; then
   macaroonAdminHex=$(sudo xxd -ps -u -c 1000 /mnt/hdd/lnd/data/chain/${network}/${chain}net/admin.macaroon)
   macaroonInvoiceHex=$(sudo xxd -ps -u -c 1000 /mnt/hdd/lnd/data/chain/${network}/${chain}net/invoice.macaroon)
   macaroonReadHex=$(sudo xxd -ps -u -c 1000 /mnt/hdd/lnd/data/chain/${network}/${chain}net/readonly.macaroon)
-  sudo sed -i "s/^LND_CERT=.*/LND_CERT=_____TO-DO_____/g" /home/admin/lnbits/.env
+  sudo sed -i "s/^LND_CERT=.*/LND_CERT=/mnt/hdd/lnd/data/chain/${network}/${chain}net/tls.cert/g" /home/admin/lnbits/.env
   sudo sed -i "s/^LND_ADMIN_MACAROON=.*/LND_ADMIN_MACAROON=${macaroonAdminHex}/g" /home/admin/lnbits/.env
   sudo sed -i "s/^LND_INVOICE_MACAROON=.*/LND_INVOICE_MACAROON=${macaroonInvoiceHex}/g" /home/admin/lnbits/.env
   sudo sed -i "s/^LND_READ_MACAROON=.*/LND_READ_MACAROON=${macaroonReadHex}/g" /home/admin/lnbits/.env
@@ -131,9 +131,14 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
     echo "LNBITS_DATA_FOLDER=/mnt/hdd/app-data/LNBits" >> /home/admin/lnbits/.env
 
     # to the install
+    echo "# installing application dependencies"
     cd /home/admin/lnbits
     sudo -u admin pipenv install
     sudo -u admin /usr/bin/pipenv run pip install python-dotenv
+
+    # to the install
+    echo "# updating databases"
+    sudo -u admin /usr/bin/pipenv run flask migrate
 
     # open firewall
     echo


frennkie:
I saw the discussion in the other PR regarding re-branding and changing the directory name. If you feel really passionate about this my suggestion would to get this done early on. So `LNbits` is only at 0.1... 

How about moving the directory from `LNBits` to `LNbits` and creating a symlink..?

frennkie:
`LND_CERT=/mnt/hdd/lnd/data/chain/${network}/${chain}net/tls.cert` that is not the right path for the LND cert...

Should be `/mnt/hdd/lnd/tls.cert`.

frennkie:
Am I missing something here..?!

```
 macaroonInvoiceHex=$(sudo xxd -ps -u -c 1000 /mnt/hdd/lnd/data/chain/${network}/${chain}net/invoice.macaroon)
...
  sudo sed -i "s/^LND_INVOICE_MACAROON=.*/LND_INVOICE_MACAROON=${macaroonInvoiceHex}/g" /home/admin/lnbits/.env
```

Running `xxd` extracts the macaroon as a hex string. But the env-setting `LND_INVOICE_MACAROON` is later used in `lnd_grpc.Client` as a file path.

frennkie:
Another question: Are you intending to run LNbits as user `admin` on the RaspiBlitz? Because `admin` does not have access to the macaroons without using `sudo`.

I would suggest to either use `bitcoin` or to use a dedicated user for `LNbits`.

eillarra:
Yes, the problem is that @arcbtc was going to test this but his Blitz is broken now... and mine is not ready yet.

So I am just guessing the path here... As I said when I opened the original PR the cert path needs to be checked. #1132 was closed too fast :)

As for the macaroons, that's the original code by @rootzoll , not sure if he wants to use a dedicated user or keep `admin`.

frennkie:
@rootzoll / @openoms  How should we do this? Should `LNbits` get a dedicated user? 

I guess the key part here is: How do "bonus" scripts get access to macaroons. A copy of the `admin.macaroon` is available (to `admin`) here: `/home/admin/.lnd/data/chain/bitcoin/testnet/admin.macaroon`

But this doesn't solve:
* accessing only with `readonly` or `invoice` macaroon
* what happens if the macaroon db is re-created.. (there is no sync between `/home/admin/.lnd/data/chain/bitcoin/testnet/admin.macaroon` and `/mnt/hdd/lnd/data/chain/bitcoin/testnet/admin.macaroon`)



@eillarra For some reason there does not seem to be a wheel available for the `grpcio` package for Python 3.7 on Arm. Therefore my Raspberry Pi 3 had to compile it which caused the activation of `LNbits` to take more than 25 minutes.. :-/

Juized:
I had a similar experience of it taking more than 20 min to compile on my rPi 4. and when it finished i had some errors:
Apr 24 11:38:15  : Started lnbits.
Apr 24 11:38:18  : Apr 24 11:38:18.000 [notice] Guard angeltest9 ($9288B75B5FF8861EFF32A6BE8825CC38A4F9F8C2) is failing more circuits than usual. Most likely this means the Tor network is overloaded. Success counts are 128/183. Use counts are 91/91. 128 circuits completed, 0 were unusable, 0 collapsed, and 0 timed out. For reference, your timeout cutoff is 60 seconds.
Apr 24 11:38:21  : Loading .env environment variables…
Apr 24 11:38:22  : [2020-04-24 11:38:22 +0100] [19105] [INFO] Starting gunicorn 20.0.4
Apr 24 11:38:22  : [2020-04-24 11:38:22 +0100] [19105] [INFO] Listening at: http://0.0.0.0:5000 (19105)
Apr 24 11:38:22  : [2020-04-24 11:38:22 +0100] [19105] [INFO] Using worker: sync
Apr 24 11:38:22  : [2020-04-24 11:38:22 +0100] [19687] [INFO] Booting worker with pid: 19687
Apr 24 11:38:22  : [2020-04-24 11:38:22 +0100] [19687] [ERROR] Exception in worker process
Apr 24 11:38:22  : Traceback (most recent call last):
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/site-packages/gunicorn/arbiter.py", line 583, in spawn_worker
Apr 24 11:38:22  :     worker.init_process()
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/site-packages/gunicorn/workers/base.py", line 119, in init_process
Apr 24 11:38:22  :     self.load_wsgi()
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/site-packages/gunicorn/workers/base.py", line 144, in load_wsgi
Apr 24 11:38:22  :     self.wsgi = self.app.wsgi()
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
Apr 24 11:38:22  :     self.callable = self.load()
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 49, in load
Apr 24 11:38:22  :     return self.load_wsgiapp()
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 39, in load_wsgiapp
Apr 24 11:38:22  :     return util.import_app(self.app_uri)
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/site-packages/gunicorn/util.py", line 358, in import_app
Apr 24 11:38:22  :     mod = importlib.import_module(module)
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/importlib/__init__.py", line 127, in import_module
Apr 24 11:38:22  :     return _bootstrap._gcd_import(name[level:], package, level)
Apr 24 11:38:22  :   File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
Apr 24 11:38:22  :   File "<frozen importlib._bootstrap>", line 983, in _find_and_load
Apr 24 11:38:22  :   File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
Apr 24 11:38:22  :   File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
Apr 24 11:38:22  :   File "<frozen importlib._bootstrap_external>", line 728, in exec_module
Apr 24 11:38:22  :   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
Apr 24 11:38:22  :   File "/home/admin/lnbits/lnbits/__init__.py", line 3, in <module>
Apr 24 11:38:22  :     from flask import Flask
Apr 24 11:38:22  :   File "/home/admin/.local/share/virtualenvs/lnbits-ztOIulw8/lib/python3.7/site-packages/flask/__init__.py", line 14, in <module>
Apr 24 11:38:22  :     from jinja2 import escape
Apr 24 11:38:22  : ModuleNotFoundError: No module named 'jinja2'
Apr 24 11:38:22  : [2020-04-24 11:38:22 +0100] [19687] [INFO] Worker exiting (pid: 19687)
Apr 24 11:38:22  : [2020-04-24 11:38:22 +0100] [19105] [INFO] Shutting down: Master
Apr 24 11:38:22  : [2020-04-24 11:38:22 +0100] [19105] [INFO] Reason: Worker failed to boot.
Apr 24 11:38:22  : lnbits.service: Main process exited, code=exited, status=3/NOTIMPLEMENTED
Apr 24 11:38:22  : lnbits.service: Failed with result 'exit-code'.

arcbtc:
The easiest solution would be switching back to REST API, I'll make a funding source now. It shouldn't take long

arcbtc:
Thanks for all the testing folks!

frennkie:
@arcbtc Is Python3.7 a hard requirement? If no then I would say simply switching to Python3.6  is the easiest solution as there is a wheel available for that: https://pypi.org/project/grpcio/#files  `grpcio-1.28.1-cp36-cp36m-linux_armv7l.whl`

>  I'll make a funding source now. 

What do you mean by that?

eillarra:
@frennkie no, it should work with 3.6. If that helps for the Blitz I see no problem in using 3.6

frennkie:
Ah damn.. not that trivial after all

```
$ pipenv --python 3.6
Warning: Python 3.6 was not found on your system…
```

So we would have to install Python 3.6 extra for this.. which of course would be possible. But may be switching to the REST as @arcbtc suggest might really be the more maintainable solution.

frennkie:
Pi Wheels has the pre build gRPC wheel for Python 3.7: https://www.piwheels.hostedpi.com/project/grpcio/

But I have to admit that I have not heard of this site until. Using this is also a switch from PyPI to there.. no a change I would make without further consideration..

arcbtc:
LND REST funding source has been restored so let's just use that rather than RPC, keep it simple

openoms:
I would very much support a dedicated user for LNbits.
The approach I took with JoinMarket and BTCpay too.
* the problem is that admin is a sudo user without needing password so running anything as admin is similar to root
* with a dedicated user a clean uninstall in very easy with
`sudo userdel -rf lnbits`
(which would keep the data directory on the disk)

eillarra:
@arcbtc are we are reverting to `LndRestWallet`?

rootzoll:
> I guess the key part here is: How do "bonus" scripts get access to macaroons. A copy of the admin.macaroon is available (to admin) here: /home/admin/.lnd/data/chain/bitcoin/testnet/admin.macaroon

Yes with more and more users for extra apps getting introduced, the strategy to copy them all into seperate drectories comes to a limit.

To prevent copying we could place the original macaroons to a seperate directory in `/mnt/hdd/app-data/lnd` and use unix file permissions to allow different user access (lndadmin-group, lndread-group).

In the main lnd config we could force the different location for the macaroons with the following settings:

```
; Path to write the admin macaroon for lnd's RPC and REST services if it
; doesn't exist. This can be set if one wishes to store the admin macaroon in a
; distinct location. By default, it is stored within lnd's network directory.
; Applications that are able to read this file, gain admin macaroon access.
; adminmacaroonpath=~/.lnd/data/chain/bitcoin/simnet/admin.macaroon

; Path to write the read-only macaroon for lnd's RPC and REST services if it
; doesn't exist. This can be set if one wishes to store the read-only macaroon
; in a distinct location. The read only macaroon allows users which can read
; the file to access RPCs which don't modify the state of the daemon. By
; default, it is stored within lnd's network directory.
; readonlymacaroonpath=~/.lnd/data/chain/bitcoin/simnet/readonly.macaroon

; Path to write the invoice macaroon for lnd's RPC and REST services if it
; doesn't exist. This can be set if one wishes to store the invoice macaroon in
; a distinct location. By default, it is stored within lnd's network directory.
; The invoice macaroon allows users which can read the file to gain read and
; write access to all invoice related RPCs.
; invoicemacaroonpath=~/.lnd/data/chain/bitcoin/simnet/invoice.macaroon
```

So every bonus app user gets part of the fitting unix user group and has in his home directory also a ./.lnd/lnd.conf that points to those macroon locations - so that even lncli should work for those users.

@frennkie this way we could prevent any copying when macaroons get updated, correct?

I opened a new issue https://github.com/rootzoll/raspiblitz/issues/1143 - to make such a infrastructure change for the v1.6 release ... fo the the v1.5 I think we can work still what we have.

rootzoll:
@eillarra OK I merged and tested and it seems to run so far (loading, create wallet, send, receive). It was a bit slow on my testing machine - but maybe that needs a restart of my Blitz. Is it so far ready for a v1.5 release or is this reverting to LndRestWallet (or something else) something to wait for?

eillarra:
> @eillarra OK I merged and tested and it seems to run so far (loading, create wallet, send, receive). It was a bit slow on my testing machine - but maybe that needs a restart of my Blitz. Is it so far ready for a v1.5 release or is this reverting to LndRestWallet (or something else) something to wait for?

Did it work with the gRPC library? I assumed you will be using the REST library in the end, because of the installation problems. I have a PR to make this requirements optional, so if you use REST the initial installation should be much lighter.

@arcbtc wanted to use gRPC because it is easier to check a individual invoice status than with REST. I think in REST it goes through all existing invoices to find one, so it can take long on some Blitzs with lots of transactions.

-----

Another thing we should discuss is what is the best option to install this (tags or not). Should we create a `raspiblitz` tag that can move between LNbits stable releases? So now it will point to 0.1.0, but then we can silently move the tag to 0.1.1 or other stable release.

I want to avoid `master`, but the current solution is not clean either: I think I have moved the 0.1.0 tag 3-4 times already as we were fixing Blitz related issues and other issues :)

arcbtc:
LNDs REST doesnt allow you to check a single outgoing payment, which is bizarre, but I have limited it to only fetching 20 outgoing payments and parsing that, its still a big blob of data though, and might need to be revised if someone has a really busy LND REST/LNbits install

arcbtc:
If GRPC takes ages to install REST should be used, and Eneko has submitted a PR our end to not include the GRPC library in the initial install
https://github.com/lnbits/lnbits/pull/26

openoms:
Adding the separate user: https://github.com/rootzoll/raspiblitz/pull/1144

Tested installing, it takes couple of minutes, but nothing extreme. 
Still quicker much than to build RTL (shouldn'tt even mention electrs)

arcbtc:
GRPC is a much better choice as it is faster and better supported. Not being able to check a single outgoing payment over LND REST is an issue. Although if using GRPC causes a load of other install issues, REST can be used

openoms:
The installation issues were solved by sorting the dependencies before 0.1.0.
I vote for staying with GRPC.

frennkie:
> GRPC is a much better choice as it is faster 

I agree

> and better supported. 

I *partially* agree.. But gRPC for sure adds a heavy dependency (on a Raspiberry Pi I think this is manageable.. on a ESP32.. not really..).

> Not being able to check a single outgoing payment over LND REST is an issue.

https://api.lightning.community/#v1-invoice - no problem to retrieve a single payment via REST.  - **Correction**: This is only for inbound invoices.

arcbtc:
> https://api.lightning.community/#v1-invoice - no problem to retrieve a single payment via REST.

this endpoint is to check an invoice added, ie incoming.

https://api.lightning.community/#v1-payments - is for outgoing and returns all payments, there is no v1/payment equivalent to v1/invoice

As far as I am aware LND can not be installed on an ESP32, if the install time is fine on the Pi, then GRPC is the way to go.

frennkie:
I was a bit quick with commenting and missed the "outgoing" part - sorry.

arcbtc:
No worries, I just don't understand why /v1/payment doesn't exist anyway. It seems like a very basic requirement for LND to have. Easy LND pr for someone who knows what they're doing!

arcbtc:
Is this saving the macaroon in a different file or is it in fact a string, as I thought lnd_rpc requires a file path?
https://github.com/rootzoll/raspiblitz/commit/178ca9b520bdd8036488212d426cda03478dfe9f#r38794927

rootzoll:
@eillarra the install is working on my side. But I can see that the app is slow and wallet balances take time to update. I think a moving tag for raspiblitz makes sense, so that you can improve still after the v1.5 release.

arcbtc:
This is an installation error, some lag in the communication between the node and lnbits.

rootzoll:
If its something to be fixed during installation its critical for the v1.5 release. After the weekend I would release a v1.5 RC2 and by the end of the week (for the hacksprint) I like to have the final release ready. So if there is something to fix .. it should be commited latest mid of next week.

Issue #1104, opened 2020-04-02
BTCpay Server showing LN invoices as partially paid since LND 0.9.2


openoms:
Linking to this issue on the BTCpay side: https://github.com/btcpayserver/btcpayserver/issues/1343#issuecomment-606533009

Seems to be a problem with connecting to LND v0.9.2 and above, waiting for an update to fix.

openoms:
Updating to LND v0.10.0 https://github.com/rootzoll/raspiblitz/issues/1116 will break the LN invoice checking in BTCpay. It will be not possible to downgrade the nodes to the last compatible (lnd 0.9.0) version from v0.10.0 as major versions have irreversible database migrations.

Our priority is to keep up with LND.

Will update BTCpay once they sort out the compatility with LND 0.9.2 and above.
Note that this bug does not affect the onchain invoices,PayJoin capability and any other function of BTCPay.

Possible solution temporarily is to connect BTCpay to an external lightning node running LND 0.9.0.

Roasbeef:
If someone has the raw listinvoices output for an invoice showing as "partial paid" we can likely get to the bottom of this pretty quickly.

rootzoll:
Seems like RockstarDev wants also to work on this this weekend:
https://twitter.com/r0ckstardev/status/1253788719800336384?s=20

openoms:
This seems to be fixed now in BTCPay and compatible with LND v0.10.0 RC4 too. 
Will need to checkout the latest master or wait for the next release: https://github.com/btcpayserver/btcpayserver/commits/master

openoms:
Can confirm that the issue is fixed when building BTCpay with the commit: 
https://github.com/btcpayserver/btcpayserver/commit/d4dbe6fe17d844caf1636d29dc7d96414f0b18c6

We can incorporate this version to the next release. 

Fixed here: https://github.com/btcpayserver/btcpayserver/commit/c7305ba5e163698150a2acc8e1038a6bb5d197e6

pretyflaco:
Can confirm that v1.5rc2 solves this. 
👌👏

PR #1140, opened 2020-04-27
BTCPay: copy admin macaroon and update TLS thumbprint when changed, update to latest master.

openoms:
As discussed in https://github.com/rootzoll/raspiblitz/pull/992#issuecomment-619348301

### DIFF
--- a/home.admin/config.scripts/bonus.btcpayserver.sh
+++ b/home.admin/config.scripts/bonus.btcpayserver.sh
@@ -5,7 +5,7 @@
 # command info
 if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
   echo "config script to switch BTCPay Server on or off"
-  echo "bonus.btcpayserver.sh [on|off|menu] [ip|tor]"
+  echo "bonus.btcpayserver.sh [on|off|menu|write-tls-macaroon] [ip|tor]"
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


openoms:
Should this be held or put in somewhere else in the view of https://github.com/rootzoll/raspiblitz/issues/1143?

openoms:
Simplified this PR:
btcpay: add write-tls-macaroon option
btcpay,lnd.export: write-tls-macaroon when changed

home.admin/config.scripts/bonus.btcpayserver.sh
```diff
@@ -275,30 +311,8 @@ WantedBy=multi-user.target
...
-port=23000
-bind=127.0.0.1
-externalurl=https://$BTCPayDomain
-
-### NBXplorer settings ###
-BTC.explorer.url=http://127.0.0.1:24444/
-BTC.lightning=type=lnd-rest;server=https://127.0.0.1:8080/;macaroonfilepath=/home/btcpay/admin.macaroon;certthumbprint=$FINGERPRINT
-" | sudo -u btcpay tee -a /home/btcpay/.btcpayserver/Main/settings.config
-    fi
+    /home/admin/config.scripts/bonus.btcpayserver.sh write-tls-macaroon
```
frennkie, line 316: Calling itself could be avoided by implementing a bash function... but I guess it's not too bad here...

openoms: yes though so that  a function would look better, but this way it is easy to call this option externally from lnd.export.sh

openoms:
Updated and tested to the latest master. BTCPay is now comaptible with the newest LND releases.

Fixing: https://github.com/rootzoll/raspiblitz/issues/1104

rootzoll:
@openoms BTCPayServer 1.0.4.2 is out ... should we go with that?
https://twitter.com/BtcpayServer/status/1255188196130000896?s=20

openoms:
Sure, I could not wait :). It only seems to differ in documentation so should be good. 
BTCpay 1.0.4.2 all the way!

Issue #1143, opened 2020-04-27
Refactor macaroon storage and access


rootzoll:
see: https://github.com/rootzoll/raspiblitz/pull/1135#issuecomment-620123128

frennkie:
> Yes with more and more users for extra apps getting introduced, the strategy to copy them all into seperate drectories comes to a limit.

> To prevent copying we could place the original macaroons to a seperate directory in /mnt/hdd/app-data/lnd and use unix file permissions to allow different user access (lndadmin-group, lndread-group).

> @frennkie this way we could prevent any copying when macaroons get updated, correct?

Yes, I think this would be a suitable way to avoid copying around the macaroon files. Also this would make it easily checkable which user has access to which macaroon by running e.g. `id btcpay` and looking at the unix group memberships.

frennkie:
I would suggest to not include "group" in the group name.. ;-)

Proposal (including IDs to make sure that they will be same on every RaspiBlitz)

GID  | Needed? | Group Name | Macaroon File Name
---- | ------- | -----------| ------------------
9700 | YES | lndadmin | admin.macaroon
9701 | YES | lndinvoice | invoice.macaroon  
9702 | YES | lndreadonly | readonly.macaroon
9703 | ??? | lndinvoices | invoices.macaroon  
9704 | ??? | lndchainnotifier | chainnotifier.macaroon  
9705 | ??? | lndrouter | router.macaroon  
9706 | ??? | lndsigner | signer.macaroon  
9707 | ??? | lndwalletkit | walletkit.macaroon

~Order is alphabetical from the macaroon file as per `LND 0.9.2`.~

Questions

* [x] Should we only start with the three needed groups and add the others later if a need arises?
* [x] I would support placing these in: `/mnt/hdd/app-data/lnd` - agreed?
* [x] I would assume we also use the same sub directory structure: `data/chain/${network}/${chain}net/` - right?

frennkie:
This is done.. right? Or do you still see open tasks (e.g. changing `adminmacaroonpath` etc.)?

rootzoll:
Yeah basically this is done ... I still think having the macarrons directly written by LND to the app-data directory (instead of copy and manual keep in sync) would be best. 

Also maybe not to forget about the LNbits still - which wil be some time until that is also fixed.

So I push this to v1.7 for later impovements.

PR #1144, opened 2020-04-27
lnbits: add dedicated user 

openoms:
- added modifications to use a non-sudo separate user `lnbits`
- fixed typo in menu

### DIFF
--- a/home.admin/00settingsMenuServices.sh
+++ b/home.admin/00settingsMenuServices.sh
@@ -564,7 +564,7 @@ if [ "${LNBits}" != "${choice}" ]; then
     sudo -u admin /home/admin/config.scripts/bonus.lnbits.sh menu
   fi
 else 
-  echo "lndmanage setting unchanged."
+  echo "LNbits setting unchanged."
 fi
 
 # DropBox process choice
--- a/home.admin/config.scripts/bonus.lnbits.sh
+++ b/home.admin/config.scripts/bonus.lnbits.sh
@@ -87,6 +87,7 @@ if [ "$1" = "write-macaroons" ]; then
   sudo cp /home/bitcoin/.lnd/data/chain/${network}/${chain}net/invoice.macaroon /home/lnbits/.lnd/data/chain/${network}/${chain}net/
   sudo cp /home/bitcoin/.lnd/data/chain/${network}/${chain}net/readonly.macaroon /home/lnbits/.lnd/data/chain/${network}/${chain}net/    
   sudo chown lnbits:lnbits -R /home/lnbits/.lnd/data/chain/${network}/${chain}net/*.macaroon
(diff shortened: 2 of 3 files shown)


openoms:
Tested successfully on 1.5RC1.

frennkie:
@openoms as @arcbtc asked in https://github.com/rootzoll/raspiblitz/pull/1135#issuecomment-620529351. Do you have the macaroons as a hex string in the `.env` file now (instead of a file path)? If so - then how should this work when `LNbits` can only use a file path currently for gRPC...

openoms:
@frennkie 
no, actually the .env does  not even have the macaroons since there is only a `sed -i` to put them an the variables are not even there yet.
```
admin@raspberrypi:~ $ sudo cat /home/lnbits/lnbits/.env
FLASK_APP=lnbits
FLASK_ENV=production
LNBITS_BACKEND_WALLET_CLASS=LndWallet
LND_GRPC_ENDPOINT=127.0.0.1
LND_GRPC_PORT=10009
LNBITS_FORCE_HTTPS=0
```

are you saying we would need to copy the macaroon for the new user anyway?

openoms:
indeed it is looking for the files:
image
image
```
Apr 28 12:57:13 raspberrypi sh[25790]: Could not find macaroon in /home/lnbits/.lnd/data/chain/bitcoin/mainnet/admin.macaroon. This might happenin versions of lnd < v0.5-beta or those not using defaultinsta
```

arcbtc:
lnd_rpc needs file paths, which was the original concern @frennkie had. 
I have been trying to find a way to pretend the hex strings are file path objects in the lndrpc LNbits funding source, as I know there is limited access to the macaroons outside of lnd in the blitz. 
So far I have had no luck :( 
https://github.com/lnbits/lnbits/blob/master/lnbits/wallets/lndgrpc.py

frennkie:
I think it might make sense to open an issue with the ln rpc author so that they accept either hex bytes or a file path. The alternative would be to not use ln rpc..

arcbtc:
Back to REST it is! Unless we can just use the file paths. If we are able to set the macaroons as hex strings in env/as vars, what's the difference to allowing LNbits access to the macaroon files? 

Isnt LNbits using a user anyway? 
Its just the difference between reading the macaroons in the install and reading them on demand

openoms:
I have added to copy the macaroons to `/home/lnbits/.lnd/`.
It works now with the separate user.

Will need to include this in the refactor of macaroon access: https://github.com/rootzoll/raspiblitz/issues/1143

frennkie:
Shouldn't this script run `source /mnt/hdd/raspiblitz.conf` somewhere at the top?

I just pulled your branch and ran it on the console and $network and $chain are empty.

rootzoll:
@frennkie merged and added loading raspiblitz.conf

PR #1147, opened 2020-04-28
Add central LND credential store (e.g. macaroons)

frennkie:
As discussed in #1143.

Scripts that (potentially) need to be checked/updated:

* [x] _background.sh (channel backup is copied to ~/.lnd/..)

* [x] bonus.lndconnect.sh (runs as admin; uses ~/.lnd/)
* [x] blitz.touchscreen.sh
* [x] bonus.btcpayserver.sh
* [x] bonus.lnbits.sh
* [x] bonus.rtl.sh
* [x] lnd.check.sh
* [x] lnd.credentials.sh
* [x] lnd.export.sh
* [x] lnd.newtlscert.sh

User accounts that are affected
* [x] pi (BlitzTUI)
* [x] btcpay
* [x] lnbits
* [x] rtl

I created a new MainMenu entry `LNDCREDS` and moved `EXPORT` into it. I moved the submenu entry `RESET` from `EXPORT` to `LNDCREDS`.


frennkie:
I proposed `lnd.credentials.sh` here.. could also be called `blitz.credentials.sh`. But it has some overlap with the existing `lnd.check.sh`. Where should we do this in future?

rootzoll:
Sounds like the best place in th existing code is "lnd.export.sh" that already has the focus to export the credentials in different forms and also alows for a macaroon reset. Renaming it to "lnd.credentials.sh" can make sense in this extension of functionality.

frennkie:
Whoops.. yes.. I meant `lnd.export.sh`... So I'll look into updating and renaming that file.

frennkie:
Looking into it.

`lnd.check.sh` has `update-credentials` and `lnd.export.sh` has `reset`. I would move those two to a new `lnd.credentials.sh` script. 

But otherwise I think it makes sense to have these three

* lnd.check.sh - run **passive** checks whether everything is in order
* lnd.credentials.sh - "manage" (e.g. reset, sync from `LND` to shared dir) credentials (`tls.cert` and macaroons)
* lnd.export.sh - export LND credentials (hexstring|scp|http)

frennkie:
@rootzoll @openoms Going through with this... would user `admin` also access the macaroons in `/mnt/hdd/app-data/lnd`.. ? I would say yes - which means that we no longer copy the macaroons to `/home/admin/.lnd/` (but there can still be a symlink). 

Agreed?

frennkie:
@rootzoll I currently have `/home/admin/.lnd` as a symlink to `/mnt/hdd/app-data/lnd/`. I have found at least one location where it is assumed that `/home/admin/.lnd` is on the internal SD and not the external drive (HDD/SSD): `_background.sh:206`

Should I revert back from moving `~/.lnd` or should we dedicate a different SD-card local path as  a (kind of) backup location?

frennkie:
I think this is ready for a review and testing.. :-D

openoms:
RTL is running with the `rtl` user

frennkie:
Ah.. ok.. I currently don't have it activate and misread the scripts.. Does it use macaroons? Reading this I would say yes:

```
data.nodes[0].Authentication.macaroonPath = '/home/admin/.lnd/data/chain/${network}/${chain}net/';
```

openoms:
@frennkie yes, uses macaroons given it has payout functions with full access to lnd.
The admin.macaroon at /home/admin/.lnd is readable to all users currently so that is what it is using. 
It is just that the `rtl` user needs to be added to the permissioned group on install.

frennkie:
@rootzoll As I moved `reset` out of `lnd.export.sh` there should be a new place in the RaspiBlitz menu.... how about `PASSWORD     Change Passwords`.. ? Makes (much more) sense to me.

frennkie:
Adapted for #1156

PR #1150, opened 2020-04-30
add missing "fi"

frennkie:


frennkie:
@openoms FYI

openoms:
thanks!

frennkie:
https://www.shellcheck.net/ <- in some areas it's annoying.. but sometimes it's also really useful..

PR #1175, opened 2020-05-14
update electrs to newer stable version

DrFredPhD:
Installs release 0.8.3 instead of 0.8.0 fixing an issue where db is compacted on every service restart greatly impacting performance and making electrum server unusable for hours after every reboot. Also changes service unit file to load after lnd has started, optimizing boot time.

### DIFF
--- a/home.admin/config.scripts/bonus.rtl.sh
+++ b/home.admin/config.scripts/bonus.rtl.sh
@@ -153,7 +153,7 @@ EOF
 
     # open firewall
     echo "*** Updating Firewall ***"
-    sudo ufw allow 3000 comment 'RTL'
+    sudo ufw allow from 192.168.0.0/16 to any port 3000 comment 'RTL'
     sudo ufw --force enable
     echo ""
 
--- a/home.admin/config.scripts/bonus.rtl.sh
+++ b/home.admin/config.scripts/bonus.rtl.sh
@@ -153,7 +153,7 @@ EOF
 
     # open firewall
     echo "*** Updating Firewall ***"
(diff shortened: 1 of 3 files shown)


home.admin/config.scripts/bonus.rtl.sh
```diff
@@ -153,7 +153,7 @@ EOF
 
     # open firewall
     echo "*** Updating Firewall ***"
-    sudo ufw allow 3000 comment 'RTL'
+    sudo ufw allow from 192.168.0.0/16 to any port 3000 comment 'RTL'
```
openoms: this is problematic because not everyone has their subnet on 192.168.0 and they would have no access to RTL on the LAN.

openoms (changes requested):
Please remove the unrelated RTL firewall change.

openoms:
I agree with updating electrs to the latest version, but I do not experience the issue you describe. Not mentioned in the release notes either: https://github.com/romanz/electrs/blob/v0.8.3/RELEASE-NOTES.md

Sure it was not only the reinstall which solved it for you?

DrFredPhD:
Hi, I reverted the commit with the change I made to RTL firewall rule (still learning git should have used a branch). I'm not sure exactly what fixed the issue, but I did clean install from this script to try fix and still had issues with electrs service not shutting down gracefully, possibly a rocksdb issue as the marker that full compaction was not getting saved. Reinstalling 0.8.3 with the modified script fixed it. There's also been some changes since 0.8.3 in electrs/master to do with graceful shutdown, I'm testing a build of master now.

I've raised the issue at electrs with some explanation of the exact problem. https://github.com/romanz/electrs/issues/249

PR #1189, opened 2020-05-17
add blitzweb

### LINKED ISSUE #388: html Status Page
Discussing #88 I posted a command to use a browser to render a dynamic web page on the LCD of the RaspiBlitz without using too much resources.

This weekend I've toyed around with that idea and made some svg-graphics that should render on the pi. The html-code so far is only a mockup, but it's possible to breathe life into this. @frennkie has already prepared python code to make the info-page more flexible.

I'd love to hear feedback from everyone. It's okay to shoot me down, if you think this is not the RaspiBlitz-Way of doing it. I won't take it personal.

You can take a look at it right here: https://jsfiddle.net/raumi75/g8zae1tL/2/

Please feel free to fork and improve it!

frennkie:
Happy to separate this PR out into multiple if necessary, but this all goes together. Here's a brief explanation.

The intention here is to establish Nginx as a core component. It will host information pages for RaspiBlitz itself (e.g. intro explanation, info screen) and will als act as a TLS reverse proxy for multiple apps (e.g. LNbits, Btcpay, RTL).

Installation and configuration of Nginx is handled by a new script: `blitz.web.sh`. It will install packages and setup listeners for `0.0.0.0:80` and `127.0.0.1:443`. It will copy two sites to `/var/www`:

* public
* blitzweb

Nginx will use the TLS certificates generated by LND (symlinked to `/mnt/hdd/app-data/nginx/` by default.

I also added another script: `blitz.cache.sh`. It mounts a `tmpfs` filesystem at `/var/cache/raspiblitz` that is running on 4 MByte of RAM. The `00infoBlitz.sh` file now exports a JSON file with all the values it gathered to `/var/cache/raspiblitz/info.json`. From there other apps can read the data.

I decided to use a RAM disk as this file is written pretty often and it definitely makes no sense to waste read/write cycles on the SD card for this. As the data is only temporary/short lived it also does not make much sense to write to the external drive. Using 4 MByte from the 1, 2 or 4 GB of RAM sounds reasonable.

The `00infoBlitz.sh` script also generates a static html file on the RAM disk which is symlinked to `/var/www/blitzweb/info/index.html` by combining the `info.j2` file with the `info.json` data. I decided to use `Jinja2` as this is a good templating language that is also used by Flask (therefore quite a few people are already familiar with it). This also works well with a possible plan to extend `blitzweb` into a full Flask application later on.

The info/status page is based on the code from @raumi75 in #388. At the moment only one value is actually used from info.json - so this is a ToDo. But I wanted to know whether this is going in right direction before spending more time on it.

I definitely consider some of the information on the info/status screen as **confidential** (e.g. versions, balances, onion address, LND pubkey(?!)). I would say access from localhost is fine. But any access from outside (either LAN or WAN) should be required to authenticate in some way. My suggestion would be to use HTTP Basic Auth:

* username: `admin`
* password: ~`Password A`~  `Password B`

@openoms @rootzoll Agreed?


frennkie:
The Apache/Nginx standard for basic auth seems to support only MD5. Which is not considered a very secure hash. It is salted... But anyway.

I'm not sure whether the risk is acceptable for hashing `Password A` like this and would therefore prefer to use `Password B`. 

[x] using `admin`:`Password B`

frennkie:
[x] Should this listen on `localhost` by default? -> `0.0.0.0:80` and  `any:443`

geco91:
@frennkie I tried to test this on my RaspiBlitz ... but the web install script shows problems:
```
admin@raspberrypi:~ $ ./config.scripts/blitz.cache.sh on
Turn ON: Cache
admin@raspberrypi:~ $ ./config.scripts/blitz.web.sh on
Turning ON: Web
Synchronizing state of nginx.service with SysV service script with /lib/systemd/systemd-sysv-install.
Executing: /lib/systemd/systemd-sysv-install enable nginx
cp: cannot stat '/home/admin/assets/www_public/': No such file or directory
chown: cannot access '/var/www/public': No such file or directory
cp: cannot stat '/home/admin/assets/www_blitzweb/': No such file or directory
chown: cannot access '/var/www/blitzweb': No such file or directory
Looking in indexes: https://pypi.org/simple, https://www.piwheels.org/simple
Collecting j2cli
  Downloading https://www.piwheels.org/simple/j2cli/j2cli-0.3.11-py2.py3-none-any.whl
Requirement already satisfied: jinja2>=2.7.2 in /usr/lib/python3/dist-packages (from j2cli) (2.10)
Installing collected packages: j2cli
Successfully installed j2cli-0.3.11
Adding password for user admin
```

frennkie:
I believe this is a problem with the deployment/sync scripts.. how did you get ` ~/config.scripts/blitz.cache.sh` there? Did you run `XXsyncScripts.sh`?

If yes.. please check if it works if you manually copy the two directories:

```
# as admin
cd ~
cd raspiblitz
git status   # make sure you are on my branch
cp -a home.admin/assets/www_blitzweb/ ~/assets/
cp -a home.admin/assets/www_public/ ~/assets/
```

Finally these two dirs need to show up in `/var/www`.

geco91:
The XXsyncScripts was not copying subdirectories in assets ... committed PR  #1195

geco91:
OK HTTPS site is now running - but I dont get the HTTPS info site yet. @frennkie can you add the logs from the web server to the XXdebug output?

frennkie:
The info site is at `/info/`

geco91:
I call on my local LAN `https://192.168.178.95` or `https://192.168.178.95/info/` and I get a `connection failed`. But `http://192.168.178.95` works.

firewall should be open - ufw status gives me:
```
443                        ALLOW       192.168.0.0/16             # allow local LAN web
```

frennkie:
`sudo netstat -tlpen | grep nginx`?

The logs are in `/var/log/nginx/`

frennkie:
As discussed here the default HTTPS will listen on localhost only.

Run `~/config.scripts/blitz.web.sh listen any` to change.

geco91:
Ah missed the `listen any` that needed to be switched on first. Now its working :) Looks cool. 

Some feedback: The page just workd after the LND was unlocked, before I got a `Access Forbidden` info even I entered the correct Basic Auth password. Thats OK just wanna let you know. 

Question: So when I turn on `listen any` the NGINX will allow any client - but now the local firewall will keep it t the local network, correct?

frennkie:
> Some feedback: The page just workd after the LND was unlocked, before I got a Access Forbidden info even I entered the correct Basic Auth password.

That makes no sense to me.. can you reproduce this behaviour?

> Question: So when I turn on listen any the NGINX will allow any client - but now the local firewall will keep it t the local network, correct?

Correct.

frennkie:
@rootzoll feel free to merge.. :-D

Issue #1142, opened 2020-04-27
[Feature Request] Balance of Satoshis


openoms:
The Balance of Satoshis by Alex Bosworth is  feature rich tool to work with LND balances: https://github.com/alexbosworth/balanceofsatoshis

* [Install instructions for the RaspiBlitz](https://gist.github.com/openoms/823f99d1ab6e1d53285e489f7ba38602)  
 
* See how to use the rebalance command with:  
    `bos help rebalance`

![image](https://user-images.githubusercontent.com/43343391/80386661-950c2480-889f-11ea-848a-a5aa8f18baf0.png)

openoms:
Current version is shown in:
https://registry.npmjs.org/balanceofsatoshis Raw Data tab
or
https://github.com/alexbosworth/balanceofsatoshis/blob/master/package.json#L69

Once installed update with:
`sudo -u bos npm i -g balanceofsatoshis`

Can be reinstalled from the menu to reset to the tested version included in the release.

openoms:
new version 5.37.4
best to keep updated closer to the release since updated quite frequently

rootzoll:
TODO: Check if BoS is installed correctly on upadte/recovery

rootzoll:
OK BoS now also makes a clean reinstall on recover/update. Will be part of v1.6 final. Closing issue.

PR #1191, opened 2020-05-18
services: add balance of satoshis

openoms:
Related: https://github.com/rootzoll/raspiblitz/issues/1142

* kept under separate user to simplify uninstall and avoid running with sudo access

TODO:
- [x] pin version to latest tested


rootzoll:
sweet :)

Issue #255, opened 2019-01-08
Research: NGINX Webserver


rootzoll:
The next steps after version 1.0 release should offer a WebUI and further HTTP based services.

To manage them all on port 80 and put them behind pathes - take a look at:
https://www.raspberrypi.org/documentation/remote-access/web-server/nginx.md

Also check about HTTPS:
https://blog.filippo.io/mkcert-valid-https-certificates-for-localhost/

And how to tun RTL and Setup behind NGINX.
Further links for research are welcome.

joernroeder:
is this going to be a public node info ui or private one for node management purposes? i could contribute some vuejs based interface. what do you have in mind here?

rootzoll:
There will be multiple scenarios to have HTTP communication. 
One will be hosting the RTL Node Management WebUI: https://github.com/ShahanaFarooqui/RTL
Another will be a SetUp-Process that is closer to what the casaHODL provides.

joernroeder:
all right, sounds like a plan.

badokun:
Here's a guide on how to setup a reverse proxy with ssl included
https://dev.to/domysee/setting-up-a-reverse-proxy-with-nginx-and-docker-compose-29jg

geco91:
see also #275 .. now that the casa node is open source check how they did the server

openoms:
I was setting up Nginx to be used with electrs, but now RTL works over https too.
Setup guide: https://github.com/ShahanaFarooqui/RTL/pull/98/files#diff-cb8c8513a661a194d788306a2bab32e7

The nginx.conf addition:

```
stream {
            upstream RTL {
                    server 127.0.0.1:3000;
            }

            server {
                    listen 3002 ssl;
                    proxy_pass RTL;

                    ssl_certificate /etc/letsencrypt/live/YOUR_DOMAIN/fullchain.pem;
                    ssl_certificate_key /etc/letsencrypt/live/YOUR_DOMAIN/privkey.pem;
                    ssl_session_cache shared:SSL:1m;
                    ssl_session_timeout 4h;
                    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
                    ssl_prefer_server_ciphers on;
            }
    }
```

I think installing Nginx (and certbot) could be an option in the service menu. 
It needs multiple ports to be forwarded and to have a domain or dynamicDNS set up already.

[...]

PR #1208, opened 2020-05-23
Update nginx and LNbits

frennkie:
Introduce the nginx directive `include` and update `LNbits` accordingly.

This is related to #255.


frennkie:
If we accept this setup, then I would something similar for RTL and so on..

PR #1209, opened 2020-05-24
ufw: remove redundant enable commands

openoms:
The firewall is enabled in: https://github.com/rootzoll/raspiblitz/blob/master/home.admin/90finishSetup.sh#L49

So all other `sudo ufw --force enable` commands are redundant,

As discussed with @frennkie


openoms:
the second PR fixes that the internet.hiddenservice.sh now displays the second forwarded port correctly.

frennkie:
LGTM

frennkie:
I resolved the conflict and merged the commits.

PR #1214, opened 2020-05-24
update menu info for lnbits

frennkie:
As `LNbits` is now only listening to `localhost` on Port **5000** the info on the screen needs to be updated. 

Access through TOR (with HTTP over TOR and HTTPS over TOR) is also updated.

@arcbtc and @eillarra please take note.

### DIFF
--- a/home.admin/config.scripts/bonus.lnbits.sh
+++ b/home.admin/config.scripts/bonus.lnbits.sh
@@ -21,24 +21,27 @@ if [ "$1" = "menu" ]; then
   # get network info
   localip=$(ip addr | grep 'state UP' -A2 | tail -n1 | awk '{print $2}' | cut -f1 -d'/')
   toraddress=$(sudo cat /mnt/hdd/tor/lnbits/hostname 2>/dev/null)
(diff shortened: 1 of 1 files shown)


frennkie:
@openoms will this simply work..? Even with existing installations?

```
/home/admin/config.scripts/internet.hiddenservice.sh lnbits 80 5002 443 5003
```

openoms:
> @openoms will this simply work..? Even with existing installations?
> 
> ```
> /home/admin/config.scripts/internet.hiddenservice.sh lnbits 80 5002 443 5003
> ```

On existing installations, where there is already an lnbits service it wouldn't add the new ports, before the service is manually removed from.the torrc first.
But any new installation (with a mew sdcard (/etc/tor/torrc) should work.
it will end up exposing the service from localhost 5002 on the Toraddress:80
and from 5003 on Toraddress:443.

frennkie:
The TOR config `/etc/tor/torrc` is not on the external drive, right? So an update/restore will also end up in a clean state. The only case that is not handled is the XXsyncScript.sh way.

openoms:
@frennkie yes, correct.

frennkie:
ok.. merging.

PR #1215, opened 2020-05-24
Add nginx config for RTL

frennkie:
Related to #255


frennkie:
Is/was there a reason that LNbits and RTL are/were not started after activation?

openoms:
> Is/was there a reason that LNbits and RTL are/were not started after activation?

RTL was causing errors during recovery when it was activated in the install script and started before the LND wallet finished the setup.

Issue #1027, opened 2020-02-12
HTTPS Proxy & LetsEncrypt


rootzoll:
Extra App like BTCPayServer and LNBits (for a working LNURL extension) need a HTTPS with an letsencrypt cert that either works with SSH-Reverse Tunnel or dynDNS&PortForwarding. For this it makes sense to have a HTTPS proxy server (nginx) with letsencrypt that other services can share in the future.

rootzoll:
Because this is a bigger task .. moving to v1.6 so that v1.5 is not delayed

frennkie:
As #255 already suggests using `nginx` is a very good solution.

We should focus *this* issue on the "getting a certificate" part. 

Let's Encrypt has (at least) two different ways to get cert https://letsencrypt.org/docs/challenge-types/

* `HTTP-01` (requires inbound Port 80 connectivity)
* `DNS-01` (requires control over a DNS (sub) domain with possibility to set `TXT` records)

Especially `DNS-01` looks very promising! :-D

rootzoll:
@frennkie do you want to push the Lets Encrypt Part to v1.7

frennkie:
I would like to include the script to install (and restore) the client in v1.6 if possible. This would include the entry in 'Services'.

But everything else still needs a lot of documentation and deciding on use cases that make sense.

frennkie:
I merged the `acme.sh` installation which also includes an update for the FAQ. While this gives (advanced) users the option to get LE certificates there is no integration with e.g. Nginx yet.

Regarding the Apps I have already merged the Nginx integration for `LNbits` and for `RTL`. But the other Apps that host web services are still open. I think having dedicated issues for each would be the best way to track it.

Keeping this issue open for a integration between the LE certificates and Nginx.

rootzoll:
Is there anything that needs to be recovered on a update?

frennkie:
The recovery should check `/mnt/hdd/raspiblitz.conf` for `letsencrypt` and if it is **enabled** it should run `config.scripts/bonus.letsencrypt.sh on`.

The certificates and the configuration info for renewals is stored on the external disk (`app-data`).

frennkie:
@rootzoll I noticed that the `Let's Encrypt` client is now in the `Node Settings & Options` menu. I first looked in the `Additional Services`. 

Selecting the entry triggers the installation - but not more. What is our plan forward? 
[...]

PR #1216, opened 2020-05-25
Add Let's Encrypt ACME client

frennkie:
This PR adds the acme.sh client for Let's Encrypt. 

An entry in the `SERVICES` menu is made for installation and deinstallation of the client. Installing the client does not trigger any outbound communication (except for the download from GitHub) - it only installs the software **with** a crontab entry. 

Advanced users can use this to setup (automatic) creation of officially signed Let's Encrypt certificates. For all other users a menu integration needs to follow later on.

Related to #1027


frennkie:
I slightly modified the UX on the "SERVICES" menu.. a prompt (that disappears automatically) confirms successful (or failed) installation.

PR #1229, opened 2020-05-31
IP2Tor subscriptions

rootzoll:
OK this is the prototype implementation - ready for review.


home.admin/config.scripts/blitz.subscriptions.ip2tor.py
```diff
@@ -20,34 +21,38 @@
...
 session = requests.session()
 if Path("/mnt/hdd/raspiblitz.conf").is_file():
-    print("# blitz.ip2tor.py")
+    cfg = RaspiBlitzConfig()
     cfg.reload()
+    ENV="PROD"
     #DEFAULT_SHOPURL="shopdeu2vdhazvmllyfagdcvlpflzdyt5gwftmn4hjj3zw2oyelksaid.onion"
     DEFAULT_SHOPURL="shop.ip2t.org"
     LND_IP="127.0.0.1"
     LND_ADMIN_MACAROON_PATH="/mnt/hdd/app-data/lnd/data/chain/{0}/{1}net/admin.macaroon".format(cfg.network,cfg.chain)
```
frennkie: This doesn't work for me. It should be `cfg.network.value` and `cfg.chain.value`. Can you re-check? The `BlitzPy` version should be `0.2.0`.

frennkie (commented):
As mentioned.. please check cfg.network.value and cfg.chain.value.

frennkie:
Ran into this issue:

```
#### Check if extension duration is as advertised ...
# EXCEPTION on subscriptions-renew of subscription index 1
time data '2020-05-31T16:22:19Z' does not match format '%Y-%m-%dT%H:%M:%S.%fZ'
```

I think there really is no need for micro seconds precision.. I will round the suspend_after value so that it's second precision only. 

```
$ grep -n "%Y-%m-%dT%H:%M" config.scripts/*
config.scripts/blitz.subscriptions.ip2tor.py:81:    return datetime.datetime.strptime(datestr,"%Y-%m-%dT%H:%M:%S.%fZ")
config.scripts/blitz.subscriptions.py:35:    return datetime.datetime.strptime(datestr,"%Y-%m-%dT%H:%M:%S.%fZ")
```

rootzoll:
@frennkie the dates form the server are different format than the ones on our local ... so just let me know what the standard is an I will work with that.

home.admin/config.scripts/blitz.subscriptions.ip2tor.py
rootzoll: Its BlitzPy 2.0 ... not sure why it was working on my side - last try the value fileds dint work ... maybe that was before 2.0. Will try again.

frennkie:
I would  prefer: `2020-05-31T16:22:19Z`

home.admin/config.scripts/blitz.subscriptions.ip2tor.py
rootzoll: Bildschirmfoto 2020-05-31 um 19 02 34

rootzoll: @frennkie so what todo now?

frennkie:
I made the change on the timeformat.. it will always be in this format:     

```
'DATETIME_FORMAT': '%Y-%m-%dT%H:%M:%SZ'
```

rootzoll:
> I would prefer: `2020-05-31T16:22:19Z`

II will commit parsing without microseconds ... so all dates from the server with be now this format, correct? Will it always be "Z" at the end?

frennkie:
> so all dates from the server with be now this format, correct?

Correct.

> Will it always be "Z" at the end?

As the shop can't know which timezone the API clients are using the values will always be UTC.

home.admin/config.scripts/blitz.subscriptions.ip2tor.py
rootzoll: Even tho that BlitzPy is version 2.0 in the code on my testing raspiblitz `pip3 list` shows that verison is just 1.0 ... so XXsyncScripts was not updating :(

rootzoll: OK there should only be one dist version in the directory ... now XXsync script is installing the correct 0.2.0 and its working

Issue #1169, opened 2020-05-12
Deserialize or I/O error after downloading blocks from Torrent


adolfosrs:
Hey,

I have just finished syncing up my node but I'm getting the following data corruption error message every now and then. Sometime the node just restarts in a previous state (something like 100 blocks in the past).

I'm wondering if thats something that comes from the torrent download or it was my node that fucked up the data.

```
ERROR: ReadBlockFromDisk: Deserialize or I/O error - ReadCompactSize(): size too large: iostream error at FlatFilePos(nFile=1334, nPos=120886144)
```

rootzoll:
This looks like a disk corruption ... see similar error reported on Bitcoind: https://github.com/bitcoin/bitcoin/issues/5668

Please run from the main menu "REPAIR" > "SOFTWARE" or call `debug` on command line and check the value of `powerWARN` if that on is bigger then 0 - it might be a weak power supply. Otherwise you could format your HDD fresh on another computer and give it another try (or replace it).

vmavromatis:
Similar problem here https://termbin.com/9ojp my power seems OK... I literally just finished downloading the torrents....does that mean I need to download the torrents again? :/

rootzoll:
Please dont use the torrent anymore  - will get removed from next RaspiBlitz version (because its too unreliable in multiple ways). If you have a RP4 use SYNC .... and if on old RP3 download and selfvalidate the blockchain on your laptop thru Bitcoin Core and then use the COPY option to copy it over LAN to your RaspiBlitz.

vmavromatis:
Cheers @rootzoll I update the README, if that's OK, so that other won't run into this until it's deprecated.

rootzoll:
@vmavromatis thanks - merged.

nyxnor:
Pls close this.

lcharles123:
seem just data corruption at a given blockchain block. 
Happened to my litecoind node when it was syncing basic filters, that reads old blk files.
`2025-01-09T05:55:17Z ERROR: ReadBlockFromDisk: Deserialize or I/O error - ReadCompactSize(): size too large: iostream error at FlatFilePos(nFile=379, nPos=26597467)
`
I noted that there is many corrpted blk files, so i restored the corrupted file from my local backup using rsync:
`rsync -lt --no-perms --no-owner --no-group   --checksum --progress --recursive LTC_08_2024/blocks/blk00* ~/.litecoin/blocks` 

[...]

PR #1232, opened 2020-06-01
Torrent deprecation

vmavromatis:
https://github.com/rootzoll/raspiblitz/issues/1169

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,7 +263,7 @@ This is also the best option if you don't like to run your RaspberryPi 3 with a 
 
 More details: [I have the full blockchain on another computer. How do I copy it to the RaspiBlitz?](FAQ.md#i-have-the-full-blockchain-on-another-computer-how-do-i-copy-it-to-the-raspiblitz)
 
-#### 3. Torrent Fallback
+#### 3. Torrent Fallback (NOTE: to-be-deprecated in 1.6)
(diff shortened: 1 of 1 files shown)


bitsam:
Can you please leave the torrent option in 1.6 release. I still run an rpi 3B, and when I tried 1.6 rc3 the blockchain got corrupted.

rootzoll:
Exactly for those cases its better to have a copy of the blockchain on your laptop ... it will repair much faster than torrent with "COPY OVER LAN" option.

bitsam:
Ok, I am downloading to my laptop now, looks like it will download in about 10 hours.

rootzoll:
The problem with the torrent: It got very slow over the last month (no constant good seeded and some clients had problems with trackers), its hard to maintain (needs to be updated constantly, eats into support/dev time) and it went also often wrong (data corruptions during download, etc) and people had to begin over and over again ... and its a not self-validated blockchain anyway.

So for a RaspberryPi 3 ... its really much more ressilient to download and self-validate the blockchain thru bitcoin-core on your laptop (keep that as a backup there just in case) and the use the "COPY OVER LAN" option.

"COPY over LAN" will take between 4 and 6 hours once you have the Blockchain on your Laptop.

PR #1238, opened 2020-06-03
readme: add details to shopping list 

openoms:
* added the part details in the shopping list to avoid confusion when the referral link is broken (or out of stock)
* clarified that the amazon links are referral links

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,13 +66,13 @@ The cheapest way is to buy and assemble the single parts yourself. There are two
 
 *The "Standard Package" is most tested and recommended if you can effort it. It aims to give you the best economic value to run all the RaspiBlitz features at a good performance and even allows you to self-validate your blockchain in under 3 days.* 
 
-* RaspBerry Pi 4 2GB (or 4GB) [amazon](https://geni.us/raspiblitz-4-2gb)
-* RaspBerry Power Supply [amazon](https://geni.us/raspiblitz-ps)
-* 1TB SSD: [amazon](https://geni.us/raspiblitz-1000gb-san)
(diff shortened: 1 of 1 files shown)


rootzoll:
@openoms your reffer there a RaspberryPi 4GB (or 8GB) ... but the link goes to a 2GB version.

In general: Do we need to a 4GB minimum for the standard package? 2GB is still around 15 USD cheaper.

openoms:
My 2GB Odroid node is running for about a year with 20-40 channels and needed to tweak the mem usage at times of high mempool and everything installed.
As the channel.db grows it takes more RAM. I will migrate to 4GB to avoid possible problems. Running out of RAM is nasty on a routing node.
2GB is ok for minimal, but would not recommend it as the best choice.

rootzoll:
OK I can go with a 4GB as standard - then we have enough headroom for the future.

LightRider5:
Looks good! The problem is that some parts are not available in the US atm. So it looks like the link is broken.

rootzoll:
@animalmother1988 we would need a new link for the 4GB, would we?

LightRider5:
We can use this link: https://geni.us/raspiblitz-4gb-new

It is has matches in all Amazon stores in the world except India. But it is not affiliated like most of the products in the shopping lists. So it is not 100% correct to call all links "affiliate links" cause most products on Amazon are not part of any affiliate program. But for the shoppers it is the easiest solution to link them to the right product in the right Amazon store. But if have no problem to call them all "affiliate links", it is the maximum package of transparency.

PR #1242, opened 2020-06-04
Update _background.sh

frennkie:
The system wide Python is missing some libs (e.g. gRCP) so use the lndlibs virtualenv instead).

### DIFF
--- a/home.admin/_background.sh
+++ b/home.admin/_background.sh
@@ -260,7 +260,7 @@ do
   recheckSubscription=$((($counter % 600)+1))
   if [ ${recheckSubscription} -eq 1 ]; then
     # IP2TOR subscriptions (that will need renew in next 20min = 1200 secs)
-    /home/admin/config.scripts/blitz.subscriptions.ip2tor.py subscriptions-renew 1200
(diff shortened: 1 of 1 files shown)


rootzoll:
Those libraries are part of the v1.6 sd card build ... in general I would recommend to get rid of the virtualenv for standard raspiblitz scripts. Better lets clean up on that side. 

The goal should be: that every raspi standard script can be excuted on the shell by just calling it without "python" before it and all dependencies/libs (that should be kept small) are available on the systemwide python.

frennkie:
Fine with me..  also now I'm aware of this and can keep this in mind for future PRs.

frennkie:
Then this line is no longer needed, right?! https://github.com/rootzoll/raspiblitz/blob/v1.6/build_sdcard.sh#L590

rootzoll:
Yes .. we would need to check the ither scripts still a bit , but we should be able to get rid of that virtualenv.

frennkie:
Virtualenvs are actually useful if you have different apps that have conflicting dependencies (e.g. one needs newer than x.y and then other needs older than x.y). But I don't think we currently have something like that.

rootzoll:
For bonus apps that run python virtual env make sense - but they sould get created once that bonus app gets installed.

Issue #1117, opened 2020-04-15
 tlsextraip settings for LND


openoms:
Currently if I have there are more than 1 lines of `tlsextraip` is given in the lnd.conf there is a warning on the LCD:
```
While reading from '<string>' [line 18]: option 'tlsextraip' in section 'Application Options' already exists
```
Also the default setting `tlsextraip=0.0.0.0` is not useful. 
Any IP address or domain used to connect vis REST or GRPC needs to be added as a new 
`tlsextraip=` or `tlsextradomain=` line.

See: https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf#L31

frennkie:
Ah.. My bad. So you would suggest to not do any configuration validation from Raspiblitz side..?

openoms:
LND would not start with an erroneous lnd.conf, so I don't see the merit of checking it again.

rootzoll:
@openoms when is that happening ... the nodes I am running all just have one tlsextraip line and all is good.

openoms:
I needed to use multiple lines due to connecting through ZeroTier and using the Tor2IP tunnel too.

frennkie:
I agree. The `tlsextraip=0.0.0.0` line makes no sense and should be removed.

frennkie:
Interesting (re)discovery.. run this on your Node: `openssl x509 -in /mnt/hdd/lnd/tls.cert -noout -text | grep -A 1 "Subject Alternative Name"`

frennkie:
I opened https://github.com/lightningnetwork/lnd/issues/4271 for the "finding" in my last comment here.

frennkie:
> when is that happening ... the nodes I am running all just have one tlsextraip line and all is good.

@rootzoll The point isn't really when this is happening.. the line makes no sense to me in any scenario. There is no possibility to connect to a RaspiBlitz using `0.0.0.0` as the destination IP address.

My suggestion would be to remove the lines that add this to `lnd.conf` from `lnd.autounlock.sh`. But I'm not sure whether we should also actively remove it from `/mnt/hdd/lnd/lnd.conf`. People who care can do this themselves. And new installations will not longer have this setting added.

Let me know if I missing something.

openoms:
​Found an interesting setting here: https://github.com/alexbosworth/run-lnd
```
#​ Delete and recreate RPC TLS certificate when details change or cert expires​
​tlsautorefresh​=1
```

[...]

PR #1253, opened 2020-06-10
Add initial commit for bonus zerotier installation

21isenough:
This bonus script adds the functionality to join a ZeroTier network via CLI for remote connection to the RaspiBlitz. 

Things TODO:
- [x] Add menu entry to `00mainMenu.sh`
- [x] Add to `00settingsMenuServices.sh`
- [x] Remove dialog from install script 
- [X] Saving the Network ID to /mnt/hdd/raspiblitz.conf for restoration/bootstrapping
- [X] Add restoration of setup to `_bootstrap.provision.sh`
- [ ] ~~Detect and deny uninstall if SSH connected via ZeroTier~~
- [ ] ~~Error message if incorrect/invalid Network ID entered~~


21isenough:
I'll be adding the above mentioned features over the weekend during the #LightningHacksprint if my "proposal" to add ZeroTier is welcome :wink:

rootzoll:
ZeroTier looks like a nice addition if you want to run your RaspiBlitz in a more complex overlay network setup.

@21isenough so sure :) happy to see an integration for RaspiBlitz. What is your personal use case?

21isenough:
> happy to see an integration for RaspiBlitz. What is your personal use case?

Very cool @rootzoll  :+1: Yes, happy to integrate this properly. 

I literally couldn't be without ZeroTier anymore. It has almost completely replaced my "normal" LAN. Every device I own is hooked up into my ZeroTier network and it makes the devices completely location and network independent. Some of the things:

- Connecting mobile wallets to my Lightning nodes
- Connecting Electrum and other wallets to a remote electrs server
- Accessing web applications like RTL/ThunderHub/BTC RPC Explorer from wherever I am. 
- SSH access to all my headless machines. 
- Syncing and backing up between devices (e.g. regular backup of `channel.backup` to a remote location)
- Running my own cloud for file sharing, contact, todo lists, calendar
- Running my own DNS server (Pihole) and routing the DNS traffic of all my devices over it

While much of this could be solved over TOR as well, ZeroTier is much less complex. My traffic is also encrypted over ZeroTier and there's no need for NAT - all my devices can talk to each other regardless of where I am. 

Here you go - that was my little love letter to ZeroTier :heart:  :smile: :laughing:

rootzoll:
If it works well for you, it will also for others :) With v1.6 there will also be IP2TOR tunnles that can be booked as subscription to make it very easy to connect to your node running behind TOR thru a public IP. ZeroTier could be a good alternative if people want to keep it all in a private overlay LAN.

Some more integration points to consider:

If you connect your mobile apps you should also consider to edit the `bonus.lndconnect.sh` script to set the ZeroTier IP as the host IP (you can do it similar to the `# tunnel thru SSH-Reverse-Tunnel if activated for that port` part).

And once you turn ZeroTier on and you have your overlay IP you should prob add it to the LND TLS cert - you can call the `lnd.tlscert-sh ip-add` for that.

21isenough:
Perfect. Looking forward to that IP2TOR tunnel :fire: 

Yes, will take care of this too :+1: Question regarding the LND TLS cert:
I'm running on the default RaspiBlitz `tlsextraip=0.0.0.0`settings and didn't need to add my ZeroTier IP. Connecting ZAP was smooth and easy. Is that needed for certain other wallets?

rootzoll:
> I'm running on the default RaspiBlitz tlsextraip=0.0.0.0 settings and didn't need to add my ZeroTier IP. 

@openoms & @frennkie is Zerotier working because of the "0.0.0.0" (that we planned to remove) or it LND TLS in general is OK with accepting requests from local networks like zerotier?

openoms:
`tlsextraip=0.0.0.0` does exactly nothing. @21isenough you should be able to connect to LND the same when deleting that line: https://github.com/rootzoll/raspiblitz/issues/1117#issuecomment-615433741

Not sure about this, but isn't connecting through zerotier similar to Tor or and ssh tunnel  that from the point of view from LND it comes from localhost?

21isenough:
Ok, I see. Yes, I believe it's the same as with Tor or/and an SSH tunnel, works all fine with me without changing anything. 
I'll be adding it to `bonus.lndconnect.sh`though.

21isenough:
This current version of the ZeroTier integration works. The install and uninstall scripts are functioning properly and the menu is enhanced to displayed ZeroTier items in the same way as other additional services. 

To further improve I would need to save the ZeroTier Network ID into `raspiblitz.conf` as well as integrate it into `_bootstrap.provision.sh`

rootzoll:
The "SERVICES" menu is now split into two parts "SETTINGS" (which should ne the palce for ZeroTier because its part of the infrastructure) and "SERCIVES" (which are additional Apps).

rootzoll:
@21isenough I merged and added the missing features. Would be great if you then can do filal testing once the v1.6RC2 is out.

21isenough:
Oh, wow :heart: Yes, can absolutely do that. Nice addition with the [?networkid] as an additional parameter :+1: 
You think, a way of joining the network via a simple GUI (through `bonus.zerotier.sh menu`) is not needed? Would have some spare time next week...

rootzoll:
Always keep in mind that the all data once given to join a network (which can be with dialogs) then on a update/recovery needs to run automatically with the stored credentials. So the `bonus.zerotier.sh on` will ask about the network id if not given so it can then be run with that data on e recovery.

I think that works for prublic zerotier networks, but private once are a bit more difficult because it needs "manual acceptence" from the zerotier admin. To further support this process you can build something into `bonus.zerotier.sh menu` but keep in mind that the user might to "redo" this after a update manually evertime when using a private network. Or is there anyway to perserve some zerotier auth data once a client joind a network that can be persisted in the hdd app-data directory so that it can survive updates?

21isenough:
Yes, your concerns are valid and I thought about this too. This is why I'm suggesting to implement a "GUI" through the `bonus.zerotier.sh menu` option. This wouldn't interfere with an update/recovery where `bonus.zerotier.sh on` is being called. 

I'm not exactly sure what you mean by "private" and "public" ZeroTier networks. But it is clear that the user will have to manually confirm his device and accept it back into his network through the ZeroTier admin panel after a restore process.

There is indeed an option to preserve the installation and settings (app-data directory). BUT I discussed this option with @openoms a while ago and we agreed that it is not a good idea to save that data to the hdd because you can potentially get very easy access to a users ZeroTier network with a stolen RaspiBlitz by just running the recovery process. We think it is better that the user has to manually accept a device back into his network instead of automating it and introducing a vulnerability into your virtual private network (of course, if the attacker knows what they are doing, they can just preserve that directory from the SD card but it still doesn't happen automatically). What's your thoughts on this?

openoms:
Agreed @21isenough. The issue with automatically restoring the zerotier connection is indeed that with the recovery process one can reset the ssh password and can get back into to the zerotier VPN without needing to be accepted again.

Needing to enter the networkID through the menu after an update would be an acceptable inconvenience and would avoid sacrificing security. There could be a a few lines about needing to login to the zerotier account and accepting the connection for the first time.
Once the `networkID` is known the link can  be given as: 
`https://my.zerotier.com/network/$networkID`

rootzoll:
> I'm not exactly sure what you mean by "private" and "public" ZeroTier networks

When I created a network there was the option to switch it from a private to a public network where every device can join by network-id without manual adding.

> Needing to enter the networkID through the menu after an update would be an acceptable inconvenience and would avoid sacrificing security. There could be a a few lines about needing to login to the zerotier account and accepting the connection for the first time.

Yes. The zerotier can check in the status (if ACCESS DENIED still) and let the user know to still add the devices and that would ne redone after every update.

PR #1250, opened 2020-06-09
ThunderHub nginx config and blitz.web.sh fixes

openoms:
- [x] add Thunderhub nginx config https://github.com/rootzoll/raspiblitz/issues/1233
- [x]  fixing blitz.web install:  https://github.com/rootzoll/raspiblitz/issues/255
- [x] add BTCRPCExplorer Nginx config
- [x] run blitz.web.sh automatically during setup
- [x] add the HTTPS ports to UFW during setup and closed on uninstall for RTL, THub and Explorer

PR #1254, opened 2020-06-11
Electrs nginx integration 

openoms:
Based on: https://github.com/rootzoll/raspiblitz/pull/1250


openoms:
Getting:
```
$ sudo nginx -t
nginx: [emerg] "stream" directive is not allowed here in /etc/nginx/sites-enabled/electrs_ssl.conf:4
```

Will try with the same syntax as with the other services.

openoms:
Serves on LAN_IP:50002 over HTTPS using the self signed certificate used by all other services.

Issue #1265, opened 2020-06-18
RESEARCH: Detect if a screen is detected or not


rootzoll:
During Setup it would be very helpfull to auto-detect if a LCD screen is connected to the RaspiBlitz or not .. for example see: #860

Maybe a python script can test thru the GPIO library to check if a screen is connected.
Try during LCD driver are on/off and also while LCD Hat is on/off.

CandleHater:
I'd like to propose the following approaches regarding this topic:
```
# touchscreen?
if [ ! -z "$(dmesg | grep 'touchscreen')" ]; then
    echo "touchscreen detected (high probability to be a TFT/LCD screen)"
fi

# second framebuffer?
if [ -d "/sys/class/graphics/fb1" ]; then
    echo "secondary framebuffer found (TFT/LCD screen would be fb1, unless more than one monitor is connected)"
fi

# HDMI?
if [ ! -z "$(tvservice -s | grep 'HDMI')" ]; then
    echo "HDMI connection detected"
else
    echo "no HDMI connection found"
fi
```

I've tested this on three Pis with TFT, HDMI and no display at all. Can others confirm that their screen is detected correctly with the first two if-statements?

rootzoll:
@CandleHater thats also an interesting approach .. much thanks.

adevoss:
On my Raspberry Pi 4 Model B Rev 1.1 used as a desktop computer (HDMI monitor and Raspberry Pi 7" touchscreen, PiOS Buster):
false/false/false

lrwxrwxrwx 1 root root 0 jul  8 23:17 fb0 -> ../../devices/platform/soc/soc:gpu/graphics/fb0
lrwxrwxrwx 1 root root 0 jul  8 23:17 fbcon -> ../../devices/virtual/graphics/fbcon
tvservice -s:
state 0x400000 [LCD], 800x480 @ 60.00Hz, progressive

On a Raspberry Pi 3 Model B Rev 1.2 (no monitor or screen, PiOS Buster Lite):
false/false/false

lrwxrwxrwx 1 root root 0 Jun 17 21:00 fb0 -> ../../devices/platform/soc/soc:fb/graphics/fb0
lrwxrwxrwx 1 root root 0 Jun 17 21:00 fbcon -> ../../devices/virtual/graphics/fbcon
tvservice -s:
state 0x40000 [NTSC 4:3], 720x480 @ 60.00Hz, interlaced

[...]

Issue #1184, opened 2020-05-16
Test Raspbian Buster Lite as base image to reduce bloat


openoms:
Most sotftware included in the Raspbian Desktop in not used in the RaspiBlitz.

Will attempt to build the SDCard image with the minimal image and identify what is missing if anything.

For the Odroid build I am using a 149 MB minimal Buster image ( https://www.armbian.com/odroid-xu4/) vs Raspbian desktop 1136 MB and even the Lite is 434 MB (https://www.raspberrypi.org/downloads/raspbian/).

frennkie:
If I recall correctly then the switch from Lite to Desktop happened when I added BlitzTUI as this uses the XServer, Display / Window manager and such things. 

Getting it to work was a lot simpler by choosing the big image and selectively removing or disabling stuff.

Build_sd also already removes many unnecessary packages.

openoms:
RaspiBlitz is running with the Desktop version from the beginning.

I think it is worth trying to identify the required extra, seems simpler to me then finding all the unnecessary packages.

frennkie:
Okay.. Then I mixed this up.. But I definitely started on creating a PyQT5 GUI from a reduced image and abandoned the approach.

> I think it is worth trying to identify the required extra, seems simpler to me then finding all the unnecessary packages

Agreed.

openoms:
Successfully built the SDcard manually on  Raspbian Buster Lite.
It has half the installed packages compared to the desktop version (~500 vs ~1100) ! 
The LCD is working already so will continue with testing the individual services (especially the TUI).

Only needs some extra packages for the xserver and display drivers:
```
The following NEW packages will be installed:
  libdrm-amdgpu1 libdrm-common libdrm-nouveau2 libdrm-radeon1 libdrm2 libegl-mesa0 libegl1 libegl1-mesa libepoxy0 libevdev2 libfontenc1 libgbm1 libgl1 libgl1-mesa-dri libglapi-mesa
  libglvnd0 libglx-mesa0 libglx0 libice6 libllvm9 libmtdev1 libpciaccess0 libpixman-1-0 libsensors-config libsensors5 libsm6 libunwind8 libwayland-client0 libwayland-server0
[...]

PR #1281, opened 2020-06-22
Extra parameters and checks in build script

adevoss:
Extra parameters added to the script. To get boot from USB drive working I added an extra parameter.
When I was reading the script I added some more parameters.

Prevent duplicate code in ~/.bashrc., cmdline.txt and config.txt.
When the script is executed more than once duplicate code is inserted.
By checking first this will be prevented.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 #########################################################################
 # Build your SD card image based on:
-# Raspbian Buster Desktop (2020-05-27)
-# https://www.raspberrypi.org/downloads/raspbian/
-# SHA256: b9a5c5321b3145e605b3bcd297ca9ffc350ecb1844880afd8fb75a7589b7bd04
+# Raspberry PiOS Buster Lite (2020-05-27)
+# https://www.raspberrypi.org/downloads/raspberry-pi-os/
+# SHA256: f5786604be4b41e292c5b3c711e2efa64b25a5b51869ea8313d58da0b46afc64
 ##########################################################################
 # setup fresh SD card with image above - login per SSH and run this script:
 ##########################################################################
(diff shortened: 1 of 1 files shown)


build_sdcard.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash
 #########################################################################
 # Build your SD card image based on:
-# Raspbian Buster Desktop (2020-05-27)
-# https://www.raspberrypi.org/downloads/raspbian/
-# SHA256: b9a5c5321b3145e605b3bcd297ca9ffc350ecb1844880afd8fb75a7589b7bd04
+# Raspberry PiOS Buster Lite (2020-05-27)
+# https://www.raspberrypi.org/downloads/raspberry-pi-os/
+# SHA256: f5786604be4b41e292c5b3c711e2efa64b25a5b51869ea8313d58da0b46afc64
```
openoms: We do need to us the Desktop version of the Raspberry Pi OS in order for the display to work.

See this issue where we attempted to build on the Lite image: https://github.com/rootzoll/raspiblitz/issues/1184

build_sdcard.sh
```diff
@@ -104,12 +159,12 @@ if [ "${baseImage}" = "raspbian" ] || [ "${baseImage}" = "dietpi" ] ; then
 fi
 
 # remove some (big) packages that are not needed
-sudo apt-get remove -y --purge libreoffice* oracle-java* chromium-browser nuscratch scratch sonic-pi minecraft-pi plymouth python2
+sudo apt-get remove -y --purge python2 python2-minimal
```
openoms: This presumes the Desktop image

build_sdcard.sh
```diff
@@ -172,30 +261,32 @@ echo ""
...
+      # requires AUTO_SETUP_AUTOSTART_TARGET_INDEX=7 in the dietpi.txt
+      # /DietPi/dietpi/dietpi-autostart overwrites /etc/systemd/system/getty@tty1.service.d/dietpi-autologin.conf on reboot
+      sudo sed -i 's/agetty --autologin root %I $TERM/agetty --autologin pi --noclear %I 38400 linux/' /DietPi/dietpi/dietpi-autostart
+   fi
+
+   if [ "${baseImage}" = "ubuntu" ] || [ "${baseImage}" = "armbian" ]; then
+      sudo bash -c "echo '[Service]' >> /lib/systemd/system/getty@.service"
+      sudo bash -c "echo 'ExecStart=' >> /lib/systemd/system/getty@.service"
+      sudo bash -c "echo 'ExecStart=-/sbin/agetty --autologin pi --noclear %I 38400 linux' >> /lib/systemd/system/getty@.service"
+   fi
```
openoms, line 289: The assumption here is that the user should be able to attach the LCD anytime afterwards and it would work - otherwise these should be included in an LCD install script.

This is not only true for the GPIO LCD but HDMI too.

build_sdcard.sh
```diff
@@ -288,36 +379,36 @@ sudo apt-get install -y autossh telnet
 sudo apt-get install -y fbi
 
 # prepare for powertest
-sudo apt install -y sysbench
+sudo apt-get install -y sysbench
```
openoms: Any particular reason for using `apt-get`?

build_sdcard.sh
adevoss: Didn't see that. So Desktop version is the default for official images.
The point for me to add the parameters is when you build an image yourself to easily include/exclude things.
When Desktop is the default it doesn't prevent people to use the Lite version as a start.
Fine with me to undo.

build_sdcard.sh
adevoss: See my comment on Desktop version above.

build_sdcard.sh
adevoss: When you build an (official) image you still can include LCD software if you want or need to.
I don't use an LCD screen and wifi so this way I can exclude these.

build_sdcard.sh
adevoss: Using apt I came across this:
WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
man apt:
SCRIPT USAGE:
...
Please prefer using these commands in your scripts.

build_sdcard.sh
openoms: since the desktop should be the recommendation these can be left in. Makes no problem if the dependencies are not present.

build_sdcard.sh
openoms: makes sense, since installing the LCD related settings is opt out this should be fine

build_sdcard.sh
openoms: Apt and apt-get does the same. Apt is a newer command, but we are using latest releases.
I would say don't fix what is not broken.

adevoss: Just took the advice from the man page. If it's your choice to use apt I will undo this.

openoms:
not tested, but looks good (utACK)

adevoss:
Also related to #1265.
If LCD can be detected, the parameter can be deleted.

Issue #1194, opened 2020-05-20
New: Tor2IP (ip2tor) Bridge Subscriptions


rootzoll:
As part of the Lightning Hackathons a tor2ip bdrige concept was developed to make it easy for a RaspiBlitz running behind TOR to buy a tunnel bridge giving it a clearnet IP+port to make services like the REST API reachable for Mobile Apps, POS systems or other apps.  

The goal if this issue is to make it possible to make a subscription to a automated shop offering such bridges - running the open shop software: https://github.com/frennkie/django-ip2tor

- [x] retest when LND moved and because of https://github.com/rootzoll/raspiblitz/pull/1385

rootzoll:
To create a subscription script its needed to easily parse JSON data from such a shop. This may get complicated with bash .. so I looked at python and to make user interaction using the following UI dialogs: http://pythondialog.sourceforge.net

This python dialog wrapper can be installed from the debian repos with:
`apt-get install python3-dialog`
To be tested.

rootzoll:
Consider a new config valaue format for storing the list of subscription to bridges - see: #1185
The data for the subscriptions can be in a seperate file next to raspiblitz.conf - so we can maybe test a new config value format here.

frennkie:
I noticed yesterday that a key part was missing in the API.. but now this workflow should work:

* Retrieve Host list: `GET` https://shop.ip2t.org/api/v1/public/hosts/
* Place Purchase Order: `POST` https://shop.ip2t.org/api/v1/public/order/ (store the resulting ID/URL)
* wait a few ms
* Retrieve Purchase Order: `GET` https://shop.ip2t.org/api/v1/public/pos/22a942b3-89de-48e4-841c-f15d4d21e69f/ (store both `item_details[0]` (e.g. in order to extend life time of bridge later) and `ln_invoices[0]` <- if empty repeat until a value shows up)
* wait a few seconds (this is a looping script running every 5-30 seconds)
* Retrieve LN Invoice : `GET` https://shop.ip2t.org/api/v1/public/po_invoices/1b7fe1ce-0ba6-4c74-81de-bbd304261fb4/
* Pay to `payment_request`
* `status` of LN Invoice should change to 2.
* Check implementation status (PO: item_details[0].object_id  (ToDo: flatten this ..!): `GET` https://shop.ip2t.org/api/v1/public/tor_bridges/0f25a0b7-e261-44eb-a01b-0b2b25981c68/  `status` should change to "A".

[...]

PR #1292, opened 2020-06-26
ip2tor: add https:// to link to make clickable

openoms:
as discussed: https://github.com/rootzoll/raspiblitz/issues/1194#issuecomment-650048682

### DIFF
--- a/home.admin/config.scripts/blitz.subscriptions.ip2tor.py
+++ b/home.admin/config.scripts/blitz.subscriptions.ip2tor.py
@@ -695,7 +695,7 @@ You payed {0} sats for the first {1} hours.
 Next AUTOMATED PAYMENTS will be {2} sats each.
 
 Your service '{3}' should now publicly be reachable under:
-{4}:{5}
+https://{4}:{5}
 
 Please test now if the service is performing as promised.
 If not - dont forget to cancel the subscription under:


rootzoll:
I think https:// can make sense ... but only when the type of the Bridge is "CUSTOM". The same dialog is shown when the user creates a IP2TOR for the gRPC API ... where the https does make no sense and may even be confusing.

openoms:
in that case would need to show the message with whiptail instead of dialog because currently the link cannot be highlighted and copied either.

rootzoll:
whiptail is fine with me :)

openoms:
Looking into it whiptail is not available in the pythondialog.
image

I did not find any reference on clickable links or text (http://pythondialog.sourceforge.net/doc/), but it does happen with the `http://` or `https://` prefixes.
I think it would be great help to the user (even myself) if all the `IP_ADRESS:PORT` entries and `.onion` addresses would be possible to be opened from the dialog messageboxes.

frennkie:
Isn't rendering (e.g. underlining) and being able to click on a link something that is also affected by the terminal application being used...? So LXTerminal or the default Gnome terminal from which you would ssh into the RaspiBlitz... ?

rootzoll:
I think we have for now live with it that the python scripts can just call dialog and not whiptail and so in standard terminal apps there are no klickable links or text that can be easy cut&pasted. Its a bit of a pain ... but that all will be good once we have the webUI.

@openoms I would propose to close this PR without merge.

openoms:
@rootzoll if you think showing the prefix would be more confusing than not being able to click and copy the numbers than close this of course. It is not too long to type, only inconvenient.

@frennkie I've only tried in the Gnome terminal, but the behaviour of the bash dialog is similar and consistent in other OS-es / terminal apps as well.

frennkie:
Do we agree that this is not targeted for v1.6?

rootzoll:
I will close this .. if someone finds an easy way to make links clickable in dialog or use whiptail from python then we can reiterate on this.

PR #1294, opened 2020-06-26
Fix typo

dennisreimann:


rootzoll:
it was originally witchcraft ... but I think you are right using "switch" will be more easy for people to process ;)

PR #1295, opened 2020-06-27
Thub update

openoms:
- introduced `.env.local` file in `/mnt/hdd/app-data/thunderhub` to persist on updates as discussed with @apotdevin
- update to v0.8.4
- change password when Password_B is changed


apotdevin:
Thanks for the quick change ⚡

PR #1296, opened 2020-06-27
Mail notification improvements

dennisreimann:
Some things I found while working on the integration with my mail server:

- Added the `notifyMailHostname` because my mailserver required a fully qualified domain for authentication. It falls back to the hostname (as before) when initializing the config option. This is an optional flag, so it should not hurt users who already activated notifications.
- Allowed to pass through additional flags from the `config.scripts/blitz.notify.sh send` call to `XXsendNotification.py`, which is useful for configuring individual mails one might want to send. (e.g. with different subjects or with custom from addresses)
- Allow [setting a custom `From` header](https://tosbourn.com/allowing-your-own-from-header-with-ssmtp/)
- Uppercased the sSMTP config options for consistency

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -772,8 +772,8 @@ If you want to use the HDMI output you can place a file called `hdmi` on the `bo
 
 If you are already logged in you can use on the console the commands:
 
-- `hdmi` --> to witch to HDMI
-- `lcd` --> to witch to LCD
+- `hdmi` --> switch to HDMI
+- `lcd` --> switch to LCD
 
 ## How do I set up VNC?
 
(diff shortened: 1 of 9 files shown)


rootzoll:
@frennkie this one would be for you to check
@dennisreimann when contributing bigger scripts updates it would be best to make PR to the v1.6 branch

dennisreimann:
Ah, ok, didn‘t know about the branch. Can adjust it if the changes are wanted and @frennkie gives a tACK :)

frennkie (commented):
Looks good to me

PR #1300, opened 2020-06-29
Add bonus script: stacking-sats-kraken

dennisreimann:
Added [my Auto-DCA](https://github.com/dennisreimann/stacking-sats-kraken) script. Let me know if this needs more documentation or integration.


rootzoll:
Interesting idea - why not :) There would still some things missing to integrate it better:
- Add an entry like "stackingstatskranken=on" to the "/mnt/hdd/raspiblitz.conf" file and when on update/recovery the `_bootstrap.provision.sh` sees this entry it will install it again.
- Store the configutarion in the path `/mnt/hdd/app-data` so that it will survive a change of sd card and also will be part of the migration backup file

dennisreimann:
Alright, done. Also improved the script to make it fully customizable via the `stacking-sats-kraken/.env` file.

dennisreimann:
Did some more finetuning and now consider it ready to go. Would appreciate someone who could verify it though :)

rootzoll:
Just got the idea, that it can make sense to add the stacking stats script as a "subscription" - because its a recurring event and it would fit the subscription UI that gets introduced with the v1.6. ... i may adapt the scripts for that.

dennisreimann:
yeah, sure. go ahead :)

Siberator:
The DCA script works perfectly. Just tested it. Thanks for contributing @dennisreimann :+1: 
Would be great to have the script integrated pretty soon! In any case a nice enrichment for the raspiblitz.

rootzoll:
I can add the script already into the v1.6 release ... making a subscription item in the menu from it can then be a goal for a future release.

dennisreimann:
@rootzoll awesome, thanks!

PR #1305, opened 2020-06-30
SCB: Local backup and timestamping

### LINKED ISSUE #1302: SCB: Local backup and timestamping
Just a question if these things are wanted, because I'd like to extend the background checks with the following:

* A `localBackupTarget` option (in addition to `scpBackupTarget`) to save the SCB file to a USB drive/stick
* Timestamp the backups, so that they do not get overwritten (e.g. `channel-20200630-061500.backup`)

dennisreimann:
Addresses and closes #1302. Currently the restore part is missing, I'll keep it as a draft until that is implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -939,6 +939,14 @@ The script `/home/admin/config.scripts/internet.sshpubkey.sh` helps on init, sho
 
 To test it - open or close a channel and check if you find a copy of `channel.backup` on your remote server. You can check the background-script logs to see details on errors: `sudo journalctl -f -u background`
 
+#### C) Local Backup Target
+
+*You can also backup the SCB to another local drive, e.g. an USB stick:*
+
+In the `/mnt/hdd/raspiblitz.conf` the parameter `localBackupTarget='[DIRPATH-WITHOUT-ENDING-/]'` can be set to activate this feature.
+
+To test it - open or close a channel and check if you find a copy of `channel.backup` in the specified location. You can check the background-script logs to see details on errors: `sudo journalctl -f -u background`
+
 ## Updating RaspiBlitz to new Version
 
 If you have a RaspiBlitz v1.2 or higher - just follow the `UPDATE` option from the main menu (choose `RELEASE` if asked) and follow the instructions.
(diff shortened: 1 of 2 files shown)


rootzoll:
I think when the there is always a "channel.backup" AND a timestamped version (like I can see in your script) then i see no problem on the upload side because its like before. Just of a user reports "my channel.backup file is corrupt or not working" then we can say, OK then take one of the old ones and rename it to "channel.backup" and give it try.

rootzoll:
For the `localBackupTarget` also think of the recovery/update situation. After an update the USB drive will not be auto-mounted and the user would need to redo this manually. 

Maybe also a `localBackupDeviceUUID` (the device UUID needed for fstab entry/mounting).

Or we even standardize the localBackupTarget-path (mountpoint) to for example to `/mnt/backup/` and just the `localBackupDeviceUUID` needs to be set in the `raspiblitz.conf` to activate the feature and then the RaspiBlitz will mount the device to the standardize the localBackupTarget-path on every boot with the `_bootstrap.sh` script. This way it will also continue to work after a recovery/update.

dennisreimann:
Yes, I thought it to be good for backwards compatibility to leave the standard channel.backup file there. I wasn't sure whether or not other custom scripts are maybe using it and it doesn't hurt to keep it.

> then we can say, OK then take one of the old ones and rename it to "channel.backup" and give it try.

Good point, as the timestamped version can be considered a fallback and this approach allows us to not complicate the restore code.

> Maybe also a localBackupDeviceID (the device UUID needed for fstab entry/mounting).

Sounds good! I don't have an idea how to implement that as I'm not knowledgable in that regard yet, but I could see how to make it work or would gladly accept contributions to this PR.

dennisreimann:
Ok, took a look at it and right now this is beyond me. There seem to be multiple things involved (e.g. the datadrive script) and I'd appreciate help or even someone taking over that part.

rootzoll:
Merging to develop the rest in the v1.6 branch

dennisreimann:
Thanks for marging and taking over that part! 👍

rootzoll:
OK I finished it so that its even available now from the new settings menu. @dennisreimann hope that you will do extensive testing once v1.6RC2 is out :)

dennisreimann:
Yeah, for sure 😀

dennisreimann:
Can confirm now: works for me 👍

PR #1316, opened 2020-07-05
kindle-display: fix and improvements

dennisreimann:
- Use latest app version (handles missing data.json gracefully)
- Improve cron command
- Stop service before removing


openoms:
Excellent thanks, reinstall with this branch fixed it!
image

EDIT: now just need to look into putting the USD and EUR symbol in the right place.

PR #1321, opened 2020-07-06
Kindle Display: Finetuning

dennisreimann:
In case the value `kindleDisplay=` isn't in the raspiblitz.conf it didn't get added. Now that works.

Also updated the kindle-display to v0.4.0, which allows to choose the exchange rates and require outside connections to be done via Tor.


openoms:
It will fetch the price and quotes over Tor now if it is active. Huge win!

dennisreimann:
Great work, @openoms 🚀

PR #1337, opened 2020-07-12
exclude docker0/ virtual eth from IP check and DHCP check

phenomeus:
for users that do more with their raspiblitz ;)


rootzoll:
@phenomeus I am rejecting the #1337 PR because it was targeted to master branch instead to v1.6 branch and changing the base to v1.6 for that PR already had too many merge conflicts. But no worry - I made the changes to the local IP detection again on the v1.6 branch and with this commit: https://github.com/rootzoll/raspiblitz/commit/9c99093574ecb3106dbdec933e31c4388e8c44ea ... so consider this merged :)

Issue #1258, opened 2020-06-12
BTCPay over HTTPS


rootzoll:
In former releases BTCPay Server was just available tru TOR. Now with #1256 it should also be available over HTTPS using the new nginx server proxy.

- [X] Test merged PR
- [X] add https://LAN_IP:23001 login to menu
- [ ] note change in FAQ

rootzoll:
The selfsigned HTTPS seems to work, but the http at port 23000 is not responding and also the TOR looks still wrong.

openoms:
The port 23000 is bound to localhost which is probably right, it cannot be port forwarded without encryption this way. 
```
    echo "
### Global settings ###
network=mainnet

### Server settings ###
port=23000
bind=127.0.0.1
externalurl=https://$BTCPayDomain

### NBXplorer settings ###
BTC.explorer.url=http://127.0.0.1:24444/
BTC.lightning=type=lnd-rest;server=https://127.0.0.1:8080/;macaroonfilepath=/home/btcpay/admin.macaroon;certthumbprint=$FINGERPRINT
" | sudo -u btcpay tee -a /home/btcpay/.btcpayserver/Main/settings.config
```

rootzoll:
@openoms I merged the PR, but now even the HTTPS self-signed is not working anymore and on TOR I get this:
![Bildschirmfoto 2020-06-26 um 14 29 53](https://user-images.githubusercontent.com/2472225/85857199-ac8d5e00-b7b9-11ea-9147-75e28e7c599b.png)

openoms:
that is weird. It is working for me.

torrc:
```
$ sudo cat /etc/tor/torrc
...
# Hidden Service for btcpay
HiddenServiceDir /mnt/hdd/tor/btcpay
HiddenServiceVersion 3
HiddenServicePort 80 127.0.0.1:23002
HiddenServicePort 443 127.0.0.1:23003
```
ports:
```
$ sudo netstat -tulpn | grep 230
tcp        0      0 127.0.0.1:23000         0.0.0.0:*               LISTEN      15727/BTCPayServer  
tcp        0      0 0.0.0.0:23001           0.0.0.0:*               LISTEN      20603/nginx: master 
tcp        0      0 127.0.0.1:23002         0.0.0.0:*               LISTEN      20603/nginx: master 
tcp        0      0 127.0.0.1:23003         0.0.0.0:*               LISTEN      20603/nginx: master 
[...]

PR #1339, opened 2020-07-14
switch nginx proxy header Host

frennkie:
Nginx is acting as a reverse proxy. There it needs to take care to set/change some headers during the communication. I ran into an issue with `LNbits` that the port was stripped (while using IP2Tor). Changing the setting to use the `$http_host` should include both the `hostname` and the `port` from the initial request: https://stackoverflow.com/questions/15414810/whats-the-difference-of-host-and-http-host-in-nginx

This should not break any of the apps when using the Nginx port (either with or without Tor).

As mentioned in: https://github.com/rootzoll/raspiblitz/issues/1258#issuecomment-657225926

### DIFF
--- a/home.admin/assets/nginx/snippets/ssl-proxy-params.conf
+++ b/home.admin/assets/nginx/snippets/ssl-proxy-params.conf
@@ -1,7 +1,7 @@
 # ssl-proxy-params.conf
 
 proxy_redirect off;
-proxy_set_header Host $host;
+proxy_set_header Host $http_host;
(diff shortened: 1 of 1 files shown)


rootzoll:
OK I think I ran into something similar with BTCPay Server ... it gave me this warning after I tried to login:
Bildschirmfoto 2020-07-15 um 02 34 17

rootzoll:
It fixed it :) hooray :D

Issue #1354, opened 2020-07-17
Feature request: Confirm Reboot/PowerOff


adevoss:
When I wanted to reboot my Pi, in the menu I choose 'Reboot' and it just rebooted.
Exactly what it should do of course.

What about a confirmation: Are you sure? A mistake is easily made.

Same with 'Off' probably. Did not try that.

frennkie:
Feel free to give this a try - sounds like a good "first issue" :-D

If you have the Touch User Interface enabled - I did implement a confirm dialog there. Might make sense to look at the wording and use something similar.

adevoss:
My first issues were #1281 and #1297 but I will give it a try this weekend. Good third issue.
I don't have a screen but the Update/Patch menu has a lot of confirmation questions. So I'll take a look at that.

rootzoll:
- [X] merge and adapt PR

rootzoll:
tested and is working. closing issue.

PR #1359, opened 2020-07-18
confirmation for reboot and shutdown (issue #1354)

adevoss:
Extra dialog box to confirm reboot or shutdown


frennkie (commented):
This is a modal/two choice dialog. Ever since Windows 95 if have disliked 'Yes' / 'No' buttons.. 

We can do better..! 😜

Can you please make it that the confirm button says what will happen when it is pressed (here reboot or shutdown)?

adevoss:
Sure.
My point to make this pr is to prevent accidentally shutdown the Pi. What buttons or text there should be is only about preference.

Made it general for reuse. Try to keep it that way. I'll change the text.
Or do you mean there should be three buttons (Shutdown/Reboot/Cancel)? Then the main menu has one option less.

rootzoll:
Thanks. Will be part of v1.6 final release.

rootzoll:
OK tested and worked :)

PR #1361, opened 2020-07-19
ip2tor: add blitz error + refactor

frennkie:
Still testing

@rootzoll Please note that `BlitzPy` is updated to 0.3.0.


frennkie:
I had to fix a bug in django-ip2tor that cause HTTP 500 errors since ~11 pm yesterday evening.

Now this looks mostly OK.. I tested listing; extending, creating new (both with ok and reject).

I did not fix every issue in subscription.letsencrypt.. needs to be done separately.

Issue #1364, opened 2020-07-19
IP address not detected in amd64 VM


adevoss:
When running _bootstrap.sh the IP address is not detected and the loop is endless.
gotLocalIP never becomes 1

OS: Debian Buster amd64
Network interface: ens3
IP address: 192.x.x.x

I don't know a solution right now that works for all setups.

adevoss:
Files involved:
_bootstrap.sh and blitz.statusscan.sh

rootzoll:
Is this still with the RC3?

adevoss:
Yes. Tested yesterday. Working on a pr.
ens3 isn't a common network interface name.

rootzoll:
PR merged ... please test if it now works for you.

adevoss:
Already tested before submitting pr. :smiley:

nyxnor:
Pr merged and tested as mentioned above, pls close this.

PR #1372, opened 2020-07-22
detect more network interface names for localip (issue #1364)

adevoss:
On other hardware/OS network names are different.


rootzoll:
Ah sorry just changed stuff in the same area - so we have a merge conflict. But will add your changes manually. Wait a sec.

rootzoll:
Thanks. Its merged.

Issue #1319, opened 2020-07-05
Issues with v1.6 RC2/RC3 (reportings)


rootzoll:
- [X] test recovery when BTC-RPC-Explorer is on

adevoss:
Downloaded build_sdcard.sh v1.6

No download for lnd version v0.10.2-beta.rc4.
Version v0.10.2-beta downloads the requested files.

Jestopher-BTC:
The prompt after installing Thunderhub instructs the user to use port 3011, but it's actually on port 3010. 3011 returns a bad gateway error.

openoms:
> The prompt after installing Thunderhub instructs the user to use port 3011, but it's actually on port 3010. 3011 returns a bad gateway error.

It should show a HTTPS link. Did you try 3011 with HTTPS? 3010 is the cleartext (HTTP) port.

openoms:
Trying Testnet:

- [x] wallet=wallet.dat needs to be specific: main.wallet=wallet.dat
- [x] LND autounlock leads to errors if the testnet wallet has a different password (needs to be disabled before switching to testnet)

Jestopher-BTC:
> > The prompt after installing Thunderhub instructs the user to use port 3011, but it's actually on port 3010. 3011 returns a bad gateway error.
> 
> It should show a HTTPS link. Did you try 3011 with HTTPS? 3010 is the cleartext (HTTP) port.

Ah, yes. That works as HTTPS. It gives me a bunch of warnings about an invalid certificate though.

openoms:
> Ah, yes. That works as HTTPS. It gives me a bunch of warnings about an invalid certificate though.

the Thub info screen warns about needing to accept the self signed certificate, also shows you the fingerprint to compare. Welcome to the world of encryption (unapproved by a third a party).

Since both are kept available we could show the the link for the cleartext port too (+ the Hidden Service :) ). Would you think that would be a better experience?

frennkie:
LNbits over TOR

`/etc/tor/torrc`

```
# Hidden Service for lnbits
HiddenServiceDir /mnt/hdd/tor/lnbits
HiddenServiceVersion 3
HiddenServicePort 80 127.0.0.1:5000
```

This should be
```
# Hidden Service for lnbits
HiddenServiceDir /mnt/hdd/tor/lnbits
HiddenServiceVersion 3
HiddenServicePort 80 127.0.0.1:5002
HiddenServicePort 443 127.0.0.1:5003
```

[...]

PR #1404, opened 2020-08-02
Update lnd.autounlock.sh

frennkie:
Should fix this: https://github.com/rootzoll/raspiblitz/issues/1319#issuecomment-667676250


rootzoll:
good catch!

PR #1415, opened 2020-08-05
add gettx and watchtx command

frennkie:
What to you think about this?

I especially like this in combination with `tmux` and blitz notify:

```
watchthistx "f4184fc596403b9d638783cf57adfe4c75c605f6356fbc91338530e9831e9e16" && ~/config.scripts/blitz.notify.sh send "TX confirmed."
```

Unless `txindex=on` the RPC `getrawtransaction` can not find all transactions. But the main idea here is to see when a mempool TX is confirmed so this use case should work all the time.

### DIFF
--- a/home.admin/_commands.sh
+++ b/home.admin/_commands.sh
@@ -170,3 +170,15 @@ function watchtx() {
 
     done
 }
+
+# command: notifyme
+# A wrapper for blitz.notify.sh that will send a notification using the configured 
+# method and settings. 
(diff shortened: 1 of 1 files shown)


frennkie:
Do we need the 'this'  part? I used it as we already have it for torthistx...

Could also be

* gettx
* torthistx
* watchtx

frennkie:
Changed to gettx and watchtx.

PR #1448, opened 2020-08-13
Fix touchscreen rotation on new LCD driver fork

phenomeus:
The logic before needed to be turn around


frennkie:
Duplicate #1490

PR #1465, opened 2020-08-19
Loop: enhanced privacy, security and update

openoms:
* adding dedicated user for Loop
* will communicate with Loop server through Tor (if Tor running)
* update to latest v0.8.0

tested ok on RaspiBlitz v1.6 node

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -443,6 +443,11 @@ echo "*** ADDING GROUPS FOR CREDENTIALS STORE ***"
 sudo /usr/sbin/groupadd --force --gid 9700 lndadmin
 sudo /usr/sbin/groupadd --force --gid 9701 lndinvoice
 sudo /usr/sbin/groupadd --force --gid 9702 lndreadonly
+sudo /usr/sbin/groupadd --force --gid 9703 lndinvoices
+sudo /usr/sbin/groupadd --force --gid 9704 lndchainnotifier
+sudo /usr/sbin/groupadd --force --gid 9705 lndsigner
+sudo /usr/sbin/groupadd --force --gid 9706 lndwalletkit
+sudo /usr/sbin/groupadd --force --gid 9707 lndrouter
 
 echo ""
 echo "*** SWAP FILE ***"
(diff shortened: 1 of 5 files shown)


openoms:
@frennkie part of this PR we give permission to all macaroons for `lndadmin`.
Do you see any concerns with this?
Loopd v0.8.0 requires all.

Also I needed to reset the macaroons to create these new ones on an old node. 
Will  need to make that part of the script as well otherwise `loopd` fails lacking macaroons.

Not sure do we have the option to create the new ones without deleting all?

frennkie:
> Loopd v0.8.0 requires all. 

Are you sure about this? I know and have seen that loop also uses `invoices.macaroon`. Do you have a link to the documentation stating that?

```
copy_mac_set_perms walletkit.macaroon lndadmin "${network}" "${chain}"
copy_mac_set_perms router.macaroon lndadmin "${network}" "${chain}"
``` 

Shouldn't `router` then be added to a (new) group called `lndrouter` (and so on) - if we follow our agreed approach.

frennkie:
Is there (already) any kind of authentication between loop and loopd? When loopd has the admin access and can spent all funds and any user on the RaspiBlitz can initiate loop out... then this is not the separate I am looking for, right?

If there are no credentials yet for loop to loopd - is it possible to use e.g. a socket that could have restrictive permissions?

openoms:
@frennkie
Tested and needed the create all the macaroons added for loop otherwise it would fail with an error stating the missing macaroon.
I can only refer to the Loop docs https://github.com/lightninglabs/loop#lnd and
and this comment: https://github.com/lightninglabs/loop/issues/274#issuecomment-677397468

Seems that Loop needs the specific macaroons to access the RPC interface. Since it connects to the localhost LND RPC there will be no other channel to communicate.
The separation to other user is to clearly define the permissions and access to files.

My line of thinking was that `lndadmin` could have access to all the rpc-s, but clearly the `admin.macaroon` does not seem to be enough for Loop, so would make sense to add new group for all the default macaroons.

openoms:
Tested ok with an old node (with missing macaroons).
`lnd.credentials.sh` now has `check` + `reset keepold` options.

In this PR we are fixing:
lightninglabs/loop#274
lightninglabs/loop#253
rootzoll#1350

PR #1471, opened 2020-08-21
Update bonus.electrs.sh

kevin-libertarian:
Fixed a typo that was causing electrum not to start through TOR


rootzoll:
Much thanks. Will be part of v1.6.1 service release.

PR #1473, opened 2020-08-21
rework infoBlitz and add VPN color

ChuckNorrison:
- fixed bugs with wrong variable syntax
- replaced 1 space intend with 2 spaces
- call ip route 8.8.8.8 to make public check
- skip netcat checks for VPN setup (tun0)
- new color cyan for VPN setup

### DIFF
--- a/home.admin/00infoBlitz.sh
+++ b/home.admin/00infoBlitz.sh
@@ -10,6 +10,7 @@ color_amber='\033[0;33m'
 color_yellow='\033[1;93m'
 color_gray='\033[0;37m'
 color_purple='\033[0;35m'
+color_cyan='\e[0;36m'
 
 ## get basic info
 source /home/admin/raspiblitz.info 2>/dev/null
@@ -39,11 +40,13 @@ if [ ${#hostname} -eq 0 ]; then hostname="raspiblitz"; fi
 if [ ${#network} -eq 0 ]; then
   network="bitcoin"
   litecoinActive=$(sudo ls /mnt/hdd/litecoin/litecoin.conf 2>/dev/null | grep -c 'litecoin.conf')
+  
   if [ ${litecoinActive} -eq 1 ]; then
     network="litecoin"
   else
     network=`sudo cat /home/admin/.network 2>/dev/null`
   fi
+
   if [ ${#network} -eq 0 ]; then
     network="bitcoin"
   fi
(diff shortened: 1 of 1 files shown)


ChuckNorrison:
the ip route 8.8.8.8 could replaced with any ip, it has not to be reachable, just match the routing table. so TOR should not have any problems with that check. maybe a fallback check to the old method could make sense. 

i dont like the tun0 hardcoded string but it should be fine

ChuckNorrison:
some findings here i want to add soon

ChuckNorrison:
got some issues with bash variable checks and syntax for data types. public_check seems fine for that data type expected. Please double check this :+1: 

```
Gathering Information (please wait) ...
INFO: active interface found: tun0
INFO: public port: 8333
INFO: VPN Setup was found
INFO: check VPN setup for lightning network...
INFO: ... succeeded
```

home.admin/00infoBlitz.sh
```diff
@@ -84,8 +87,17 @@ else
   color_ram=${color_green}
 fi
 
-# get name of active interface (eth0 or wlan0)
-network_active_if=$(ip addr | grep -v "lo:" | grep 'state UP' | tr -d " " | cut -d ":" -f2 | head -n 1)
+public_vpn_state="False"
+# get name of active interface with route to internet
+network_active_if=$(ip route get 8.8.8.8 | awk  -- '{print $5}')
```
frennkie, line 92: Looks a bit awkward to have a hard coded IP here... how about this:

`ip addr | awk '/state UP/ {print $2}' | sed 's/.$//' | head -n 1`

Or will that fail if multiple interfaces are up (e.g. eth0 and tun0 ) ? If it fails.. would 255.255.255.255 work instead of 8.8.8.8?

frennkie:
> fixed bugs with wrong variable syntax
> replaced 1 space intend with 2 spaces

It's hard to only identify these changes. Would you mind putting these in to a separate PR?

> call ip route 8.8.8.8 to make public check

I left a alternative suggestion in the review.

> skip netcat checks for VPN setup (tun0)
> new color cyan for VPN setup

I'm not entirely sure we want/need this (also why I'm asking to keep the changes separated).

frennkie (changes requested):

home.admin/00infoBlitz.sh
ChuckNorrison: i will test this with my setup, thanks

ChuckNorrison:
VPN Setup is not a use case, close

home.admin/00infoBlitz.sh
ChuckNorrison: `ip addr | awk '/state UP/ {print $2}' | sed 's/.$//' | head -n 1`

results in eth0 than tun0.

ChuckNorrison: `ip route get 255.255.255.255 | awk  -- '{print $5}'`

works for me

ChuckNorrison:
After setup my LCD, i observed my INFO echos. That was not intended :) But i like the additional DEBUG echo. I will rework that solution after https://github.com/rootzoll/raspiblitz/pull/1478 was accepted

PR #1477, opened 2020-08-23
use ip route over ip addr

ChuckNorrison:
should be a more generic, more safe and faster solution without any port forwarding or firewall dependencies. Will support virtual and tunneled network interfaces (tun0 for OpenVPN)


ChuckNorrison:
this was not as expected...

PR #1478, opened 2020-08-23
use 'ip route get' over 'ip addr' for interface check

ChuckNorrison:
should be a more generic, more safe and faster solution without any port forwarding or firewall dependencies. Will support virtual and tunneled network interfaces (tun0 for OpenVPN)

### DIFF
--- a/home.admin/00infoBlitz.sh
+++ b/home.admin/00infoBlitz.sh
@@ -85,7 +85,7 @@ else
 fi
 
 # get name of active interface (eth0 or wlan0)
-network_active_if=$(ip addr | grep -v "lo:" | grep 'state UP' | tr -d " " | cut -d ":" -f2 | head -n 1)
+network_active_if=$(ip route get 255.255.255.255 | awk -- '{print $5}')
 
 # get network traffic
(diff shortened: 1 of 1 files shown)


home.admin/00infoBlitz.sh
```diff
@@ -85,7 +85,7 @@ else
 fi
 
 # get name of active interface (eth0 or wlan0)
-network_active_if=$(ip addr | grep -v "lo:" | grep 'state UP' | tr -d " " | cut -d ":" -f2 | head -n 1)
+network_active_if=$(ip route get 255.255.255.255 | awk -- '{print $4}')
```
frennkie: missing a `| head -n 1` right..?

frennkie (changes requested):

ChuckNorrison:
I tested this call on some systems with tunneled adapters, with wlan and with lan only. I did not tested this with multiple physical adapters. 

Did not tested this with TOR or IP2TOR solution. 

`ip route get 255.255.255.255 | awk -- '{print $4}' | head -n 1`

This should always result the correct interface on all systems.

frennkie (approved):

PR #1479, opened 2020-08-23
info display should visualize VPN connection

ChuckNorrison:
currently it is visualized red even everything is working great.

(fixed some bad code from last try. netcat is great and should be used!)


ChuckNorrison:
only useful if https://github.com/rootzoll/raspiblitz/pull/1478 is merged

ChuckNorrison:
ok i did not get that merge with conflicting states here....

Issue #1436, opened 2020-08-11
White Screen after every Boot, maybe because of newest Kernel? 


phenomeus:
Hi everybody, 

I have a white screen that is not going away, i think its dead (but still powered on?) or maybe that came after kernel updates etc. can‘t say for sure, i noticed it too late. 

`uname -a:
Linux raspberrypi 5.4.51-v7l+ #1327 SMP Thu Jul 23 11:04:39 BST 2020 armv7l GNU/Linux `

Any idea how to debug and get it to work again? I already tried to power down the pi and removed the power cord. 

Thanks in advance

phenomeus:
It looks like a kernel and driver issue 

https://github.com/goodtft/LCD-show/issues/223

phenomeus:
There is a fork that is updated to work with newest kernel, trying right now myself ~~, will report back.~~ its working. Reapplying the screen rotation again  

```
sudo rm -rf LCD-show
git clone https://github.com/MrYacha/LCD-show.git
chmod -R 755 LCD-show
cd LCD-show/
sudo ./LCD35-show
```

phenomeus:
Rotation activation by menu brings back white screen. Will debug further

phenomeus:
I created two pull requests. My screen is working again and is rotated correctly. 

(Sorry for 2 patches, I’m still learning how to use github ;) )

phenomeus:
I have to rotate the touchscreen also, 180 degree

phenomeus:
Touch screen rotation also fixed and tested. Logic had to be turned around. PR is online

rootzoll:
Thanks .. hope to find time to review your changes in the next days.

frennkie:
I have this issue too.

frennkie:
@phenomeus wouldn't the three PRs better be put into one..? Would be great if you could look into that.

Would also be nice to have a more descriptive branch name... Then it would be easier for me to quickly check out yand test your proposed changes.

phenomeus:
Totally agree but I’m a GitHub newb :) seriously no idea how to merge my 3 prs and work with branches. 

Open to any guidance.

frennkie:
But you do use the command line, right?!

In general I would recommend you go have a look some tutorials.. but here are the essential steps:

```
git remote add phenomeus https://github.com/phenomeus/raspiblitz.git
git remote add upstream https://github.com/rootzoll/raspiblitz.git
git fetch --all --prune
[...]

PR #1490, opened 2020-08-25
Fix white screen

frennkie:
After updating v1.6 to a newer kernel (running `apt-get upgrade`) my screen stayed white.

This fixes it for the newer kernel. Might not work on the old kernel.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -883,7 +883,7 @@ if [ "${lcdInstalled}" == "true" ]; then
      echo "*** LCD DRIVER ***"
      echo "--> Downloading LCD Driver from Github"
      cd /home/admin/
-     sudo -u admin git clone https://github.com/goodtft/LCD-show.git
+     sudo -u admin git clone https://github.com/MrYacha/LCD-show.git
      sudo -u admin chmod -R 755 LCD-show
      sudo -u admin chown -R admin:admin LCD-show
      cd LCD-show/
--- a/home.admin/config.scripts/blitz.lcdrotate.sh
+++ b/home.admin/config.scripts/blitz.lcdrotate.sh
@@ -24,7 +24,7 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
     echo "lcdrotate=1" >> /mnt/hdd/raspiblitz.conf
   fi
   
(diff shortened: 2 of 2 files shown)


ChuckNorrison:
i got this working on my system manually with that referenced repo. i played a lot with the common way, ended up copiying the code to a script and after the reboot i just have to enter `lcd` and wait for the reboot.

Here are the parts which worked

```
#!/bin/bash
echo "*** LCD DRIVER ***"
echo "--> Downloading LCD Driver from Github"
cd /home/admin
sudo -u admin git clone https://github.com/MrYacha/LCD-show.git
sudo -u admin chmod -R 755 LCD-show
sudo -u admin chown -R admin:admin LCD-show
cd LCD-show/
# set comit hard to old version - that seemed to run better
#
#sudo -u admin git reset --hard ce52014

# install xinput calibrator package
echo "--> install xinput calibrator package"  
sudo apt install -y libxi6 
sudo dpkg -i xinput-calibrator_0.7.5-1_armhf.deb                                  
                                                                                    

sudo sed -i 's/^lcd2hdmi=.*/lcd2hdmi=on/g' /home/admin/raspiblitz.info 2>/dev/null
sudo sed -i 's/^lcd2hdmi=.*/lcd2hdmi=on/g' /mnt/hdd/raspiblitz.conf 2>/dev/null
cd /home/admin/LCD-show
./LCD-hdmi
```

After reboot i have to execute `lcd`. now my screen is back :+1: 

// the last line should be working with ./LCD35-show, but i got it not working this way.

rootzoll:
Its a bit more complicated if we want to include the fix into the v1.6.1 ... because the base image will stay the same for such a service release. This PR is more for a v1.7 release ... but that might a bit long for people wanna upgrade their kernel.

So I will take parts of this PR and also scripts from @ChuckNorrison and turn this into a solution that RaspiBlitz will check on boot time if kernel has been updated and then might correct the LCD driver. This should then worl for the v1.6.1

rootzoll:
closing this PR because I needed to fix this with more detail - but used parts of this PR. See: 
https://github.com/rootzoll/raspiblitz/issues/1436#issuecomment-685847430

frennkie:
@rootzoll this is really confusing..! Please use PRs for the things you do too!

PR #1493, opened 2020-08-26
Fix lnd.setaddress.sh usage message

nolith:
I suppose this script was renamed, but the usage message still refers to the old name.


ChuckNorrison (approved):

rootzoll:
@nolith good catch @ChuckNorrison thanks for review - I will merge and it will be part of v1.6.1 service release.

PR #1494, opened 2020-08-27
Update lndmanage to v0.11.0

bitromortac:
Updates lndmanage to the latest [release](https://github.com/bitromortac/lndmanage/releases/tag/v0.11.0), which is compatible with LND v0.11.0. The update includes a bunch of dependency updates. I tested installation on a raspberry pi, however not on the raspiblitz. The installation process through pip should be able to find precompiled dependencies and the process should be faster. I removed the atlas and python-dev packages, as they should not be needed. Also lndmanage should be able to run on nodes without any channels now.

### DIFF
--- a/home.admin/config.scripts/bonus.lndmanage.sh
+++ b/home.admin/config.scripts/bonus.lndmanage.sh
@@ -17,8 +17,8 @@ fi
 # show info menu
 if [ "$1" = "menu" ]; then
   dialog --title " Info lndmanage " --msgbox "\n\
-Usage: https://github.com/bitromortac/lndmanage/blob/master/README.md
-Have at least one channel active to run it without error.\n
+Usage: https://github.com/bitromortac/lndmanage/blob/master/README.md or
+lndmanage --help.\n
 To start type: 'manage' in the command line.
 " 9 75
   exit 0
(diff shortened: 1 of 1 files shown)


ChuckNorrison (commented):
python3 -m pip is more safe than pip3. Some issues can happen on paths and installation of different python software environments. What are the benefits to change that behaviour to pip3?

bitromortac:
Ok, yes, never thought about it as I'm always running in virtual environments, thanks! Related article: https://snarky.ca/why-you-should-use-python-m-pip/
Updated to `python3 -m pip`.

rootzoll:
@bitromortac sounds great - does this update of lndmanage also work with lnd version below v0.11.0 ... our next update v1.6.1 of raspiblitz will just be a service update to the v1.6 (the vasic sd card build will be the same - just updated scripts) so lnd will not get updated yet to 0.11.0 with v1.6.1 ... that will come later with RaspiBlitz v1.7. Let me know so that I can assign your PR to the correct Release Milestone.

bitromortac:
> @bitromortac sounds great - does this update of lndmanage also work with lnd version below v0.11.0 ... our next update v1.6.1 of raspiblitz will just be a service update to the v1.6 (the vasic sd card build will be the same - just updated scripts) so lnd will not get updated yet to 0.11.0 with v1.6.1 ... that will come later with RaspiBlitz v1.7. Let me know so that I can assign your PR to the correct Release Milestone.

Hey, I ran the integration tests against lnd v0.10.4-beta and it should support this version too.

rootzoll:
@bitromortac thanks. Will be part of v1.6.1 service release.

Issue #1467, opened 2020-08-19
bitcoind service not running error on each reboot


Jestopher-BTC:
Every time I restart my node, I get prompted with the following error. After a few moments , I get notified that my raspiblitz is running. When I enter 'raspiblitz' instead of debugging, everything works as expected.

*****************************************
* The bitcoind service is not running.
*****************************************
If you just started some config/setup, this might be OK.

More Error Detail:
error code: -28 error message: Loading block index...

POSSIBLE OPTIONS:
-> Use command 'repair' and then choose 'DELETE-INDEX' to try rebuilding transaction index.
-> Use command 'repair' and then choose 'RESET-CHAIN' to try downloading new blockchain.
-> Use command 'debug' for more log output you can use for getting support.
-> Use command 'menu' to open main menu.
-> Have you tried to turn it off and on again? Use command 'restart'

bitcoin error: error code: -28
error message:
Loading block index...

Debug log for reference:
https://termbin.com/spge4

bitsam:
Looking at the log, the bitcoind service is still verifying the blocks between 16-32%. You can keep an eye on this by re-running the command. sudo tail -n 20 /mnt/hdd/bitcoin/debug.log

and see if the progress increases. I also notice that things get a little out of order and sometime the system wants you to unlock your wallet. so you can also force this via the command prompt with the following command, lncli unlock

I would let it run for a while and keep and eye on the bitcoin/debug.log and see if it progresses, it may also have to download several blocks if you have been offline for a while. BTW, were you running 1.6 for a while before this happened, or did this happen immediately after upgrading to v1.6?

hermanlim:
Try waiting a while (5-10 mins should be sufficient) and then running raspiblitz from command line.

rootzoll:
This might be a timing issue also @CandleHater reported .. normally waiting like @hermanlim mentioned should be a workaround for now. Will take a look to see if we can optimze here for the v1.6.1 release.

CandleHater:
I guess a good solution to this is checking if `bitcoinErrorFull` has Error -28 (means RPC_IN_WARMUP, see https://github.com/bitcoin/bitcoin/blob/master/src/rpc/protocol.h#L49) around here https://github.com/rootzoll/raspiblitz/blob/aef8b3a1da5c0df5916ba828f5adc808fd2de624/home.admin/80scanLND.sh#L47

thecharacter:
[...]

Issue #1466, opened 2020-08-19
error code: -28 error message: Loading block index...


grnqrtr:
I get this on every reboot and and I've seen a number of other people get this on 1.6. Seems like it is giving up to early or something, doesn't give enough time to let the block index load.
```
*****************************************
* The bitcoind service is not running.
*****************************************
If you just started some config/setup, this might be OK.

More Error Detail:
error code: -28 error message: Loading block index...

POSSIBLE OPTIONS:
-> Use command 'repair' and then choose 'DELETE-INDEX' to try rebuilding transaction index.
-> Use command 'repair' and then choose 'RESET-CHAIN' to try downloading new blockchain.
-> Use command 'debug' for more log output you can use for getting support.
-> Use command 'menu' to open main menu.
-> Have you tried to turn it off and on again? Use command 'restart'

bitcoin error: error code: -28
error message:
Loading block index...
```

ChuckNorrison:
hey i get this too after power off and on. last reboot it worked. Everytime if that happens, i have to update public ip manually with the dynupdate url

Jestopher-BTC:
Just got this error with a completely fresh SD.

bitsam:
run, sudo tail -n 20 /mnt/hdd/bitcoin/debug.log

See if it is slowly verifying the blocks or downloading blocks. Once the chain is sync'd then you can run , raspiblitz to launch the menu and go back through the normal loading without a reboot. 

I also find running the lnd wallet unlock to help with the command, lncli unlock

bitsam:
I had some teething problems with 1.6, and decided to blow away the chain data. Which was probably an over-reaction, as once I had finally copied it over I still got the above errors, and it was caused by the the slow verification of blocks or need to download the last few remaining blocks while the node had been offline. You can check all this in the /bitcoin/debug.log

ChuckNorrison:
so its just a race condition between the bitcoind and lnd. Wait for start lnd while bitcoind is syncing. Not that easy for a script to get that state i think, if it depends a lot on sync time... can be seconds, days or weeks. At setup from scratch that all worked well.

Further in my cases there was no need to interact... the LND waits in a suspend state or something as i observed. So only wait for the bitcoind sync progress. that could be visualized a bit more

rootzoll:
This seems to be a timing issue .. marked for v1.6.1 to change timing on this.

thecharacter:
[...]

PR #1497, opened 2020-08-27
Fix: "service is not running" error upon boot

CandleHater:
As mentioned in https://github.com/rootzoll/raspiblitz/issues/1467 and https://github.com/rootzoll/raspiblitz/issues/1466 there's an issue regarding a falsely reported error (service is not running) upon boot if the bitcoind service hasn't finished warming up within 10min. This PR fixes that problem by checking if the variable `bitcoinErrorFull` contains the `error code: -28`, after the 10min threshold has been surpassed.

This has been tested on two devices (Pi4/P3) several times now. In fact it is now possible to lower the 10min uptime check: https://github.com/rootzoll/raspiblitz/blob/80de4054c66e4083055a1f7148eda4b7e831c8db/home.admin/80scanLND.sh#L66 In my experience there shouldn't be a problem halving (yay another one) it down to 5min.

### DIFF
--- a/home.admin/80scanLND.sh
+++ b/home.admin/80scanLND.sh
@@ -63,9 +63,11 @@ if [ ${bitcoinActive} -eq 0 ] || [ ${#bitcoinErrorFull} -gt 0 ] || [ "${1}" == "
   else
 
     # output when user login in as admin and bitcoind is not running
-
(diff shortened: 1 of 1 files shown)


rootzoll:
Thanks - looks good. I think 10minutes is a good time window before displaying an error .. bitcoind is really not a quick starter in general.

CandleHater:
Lowerd to 5min it wouldn't show an error, it would display `The bitcoind service is warming up: Loading block index Please wait ...` after 5min due to my check regarding "error code -28". It would offer users more info about what it is actually doing, then just "wait at least 10min".

rootzoll:
@CandleHater used your commited fix - tested good :)

Issue #1411, opened 2020-08-05
Update to: Specter v0.8.1


rootzoll:
see https://github.com/rootzoll/raspiblitz/issues/1171#issuecomment-669027062 from #1171

ben-kaufman:
Hi, in case you haven't started with this yet, there's version 0.6.1 which contains a few bug fixes and would be nice to have instead.

https://github.com/cryptoadvance/specter-desktop/releases/tag/v0.6.1

rootzoll:
@ben-kaufman thx, havent yet 0.6.1 sounds good :)

rootzoll:
Also merged PR #1481 to add missing library

ChuckNorrison:
Here we go :)
https://github.com/cryptoadvance/specter-desktop/releases/tag/v0.7.2

rootzoll:
Update PR merged - moving to final testing.

openoms:
Specter 0.8.0 is out:
https://github.com/cryptoadvance/specter-desktop/releases/tag/v0.8.0

Update successfully with the UPDATE menu (available in the latest script for RaspiBlitz v1.6.1)

I suggest we should ship with the latest version by default.

openoms:
​Update from Specter v0.5.5 in RaspiBlitz v1.6 paste these command to the RaspiBlitz terminal:
```
#uninstall old version
config.scripts/bonus.cryptoadvance-specter.sh off
# install the missing dependency
sudo apt install -y libffi-dev
# change script to use the latest version
sed -i "s#cryptoadvance.specter==*#cryptoadvance.specter==0.8.1#g"  /home/admin/config.scripts/bonus.cryptoadvance-specter.sh
```
Install Specter from the SERVICES

Edit: updated to v0.8.1

rootzoll:
Update Bug Reported: https://github.com/rootzoll/raspiblitz/pull/1566#issuecomment-699681734

rootzoll:
@openoms merged #1619 ... did you found any hint of the bug mentioned above while testing.

openoms:
Nope, but I am way past of updating from 0.5.5. 
There were no errors installing v0.8.0 or 0.8.1. Update works too.

openoms:
We could probably add this entry to the bitcoin.conf:
![image](https://user-images.githubusercontent.com/43343391/95376083-77c09500-08d8-11eb-83f2-8596aa04bbb0.png)

[...]

PR #1502, opened 2020-09-01
Specter update to v0.7.2

openoms:
installs and runs well on RaspiBlitz v1.6 

Related: https://github.com/rootzoll/raspiblitz/issues/1411


openoms:
To update manually from Specter 0.5.5 (in raspiblitz v1.6):
```
#uninstall old version
/config.scripts/bonus.cryptoadvance-specter.sh off
# install the missing dependency
sudo apt install -y libffi-dev
# change script to use the latest version
sed -i "s#cryptoadvance.specter==0.5.5#cryptoadvance.specter==0.7.2#g"  /home/admin/config.scripts/bonus.cryptoadvance-specter.sh
```
and install from SERVICES

k9ert:
Sorry to have not spotted that earlier. Here one have to change the version explicitely in the pip-command. Why have that been introduced in this commit at all @rootzoll  :
https://github.com/rootzoll/raspiblitz/commit/a5864f71243e3f8eb2222814dcb349edf6584292
The beauty and the difference to mynode was that the latest version got installed. If one wants to pin it, he can still add the version but the default should imho be to install the latest version, no?
Specter is doing very quick releases and is fixing a lot o bugs rolling forward. So i can't see many arguments pinning the version.

ChuckNorrison:
Stick to a specific revision is a basic question of usage. For the RaspiBlitz project, i like it "the sticky" way, for a more safe first time user experience and make room to grow with the system and get some ideas how it works. It should not be hard to override the sticky revisions manually after basic knowledge of the setup without much untested things in to get interrupted on this way.

But great idea to get some more comfort on this topic. This could be extended via the Patch menu to take control of the repos in use

openoms:
@k9ert our aim is to know exactly what is running on each raspiblitz version and if someone made manual changes can always fall back to a reinstall or tested sdcard image.

Specter can make it to the UPDATE menu as well if we find a reliable way to.do it. Can be simple pip upgrade. That would add an update option to the script too.

ChuckNorrison:
Hey, i tested the upgrade successful with this:

> sudo -u bitcoin /home/bitcoin/.specter/.env/bin/python3 -m pip install --upgrade cryptoadvance.specter==0.7.2
> sudo systemctl restart cryptoadvance-specter

Great update here

// fixed

dennisreimann:
Can confirm this works too. Used the upgrade command proposed by @ChuckNorrison above, though for me it was `sudo systemctl restart cryptoadvance-specter` to restart the service. (dash instead of period in the service name)

home.admin/config.scripts/bonus.cryptoadvance-specter.sh
```diff
@@ -287,5 +287,15 @@ if [ "$1" = "0" ] || [ "$1" = "off" ]; then
   exit 0
 fi
 
+# update
+if [ "$1" = "update" ]; then
+  echo "#    --> UPDATING Cryptoadvance Specter"
+  sudo -u bitcoin /home/bitcoin/.specter/.env/bin/python3 -m pip install --upgrade cryptoadvance.specter
```
openoms, line 293: Added the option to upgrade to the latest from https://pypi.org/project/cryptoadvance.specter/#history once the pinned Specter version is installed.

PR #1509, opened 2020-09-05
update Bitcoin to v0.20.1

frennkie:
should be as straight forward as this.


rootzoll:
This merge is a bit confusing ... because v1.6.1 was not planned to have any changes on the build sd card script. But in release v1.6.1 would be merged into master. @frennkie I will try to revert and merge into v1.7.

PR #1523, opened 2020-09-09
Add vagrant support

nolith:
This will allow building a raspiblitz base image with packer and spin up
a development environment in a VM.

This is designed to work on top of #1297

PR #1524, opened 2020-09-09
Scripts should be executable

nolith:
This is a very minor PR.

It's a pre-requirement for https://github.com/rootzoll/raspiblitz/pull/1523 as it will link the git workdir on the development machine into admin home inside the virtual machine.

If scripts are not executable, nothing will work :D


rootzoll:
@nolith I dont understand the changes in this PR. Sure scripts need to be executable ... but thats done by the build sd card script after the scripts get copied to the target system - see: https://github.com/rootzoll/raspiblitz/blob/2a66fba8e6cc3236073497014fa815220002945e/build_sdcard.sh#L732

nolith:
@rootzoll yes I noticed that scripts are marked as executable when building the sd card.

This changes are needed for https://github.com/rootzoll/raspiblitz/pull/1523 where is possibile to run a development version of raspiblitz in a virtualbox VM with vagrant.

What vagrant do is providing a shared disk between the host and the VM, with this we can edit scripts locally and have them ready to use in the VM. But if they are not executables on the repo, they cannot be executed inside the VM.

I wrote this as a separate PR to avoid overloading https://github.com/rootzoll/raspiblitz/pull/1523

I thought this could be merged in the meantime.

rootzoll:
I didnt knew that Git was tracking file executable permissions - like mentioned here:
https://medium.com/@tahteche/how-git-treats-changes-in-file-permissions-f71874ca239d

OK lets see if this works ... will merge into v1.6 branch

frennkie:
Should be no problem..

Issue #1488, opened 2020-08-25
Update: JoinMarket v0.7.1 (with joininbox menu v0.1.11) 


openoms:
working to refactor the update script
The challenge is to use system-site-packages in the jmvenv virtual env to keep the QT GUI working from the blitz.

rootzoll:
merged PR - setting for final testing

rootzoll:
Install on final testing is working. Will be part of v1.6.1 release. Closing issue.

openoms:
JoinMarket v0.7.1 is released, just testing installation. Could not find significant differences in dependencies.

openoms:
Installation is tested ok. 
The changes with payjoin will need modifications in the joininbox menu: https://github.com/openoms/joininbox/issues/15
but basic functionality (running the Yield generator and walllet generation and info) is working.

openoms:
To update JoinMarket separately can use this much improved script:
```
# download from this repo
wget https://raw.githubusercontent.com/rootzoll/raspiblitz/v1.6/home.admin/config.scripts/bonus.joinmarket.sh
# uninstall old version (wallets are preserved on the disk)
sudo bash bonus.joinmarket.sh off
# install with the latest updates
sudo bash bonus.joinmarket.sh on
```

rootzoll:
OK install seems to work - rest is up to you @openoms to test on Release Candidate. Will be part of v1.6.1 - closing this issue.

PR #1527, opened 2020-09-09
JoinMarket install fix and update to JoininBox v0.1.4

openoms:
related: https://github.com/rootzoll/raspiblitz/issues/1488


rootzoll:
merged into v1.6 and dev

PR #1531, opened 2020-09-10
Checkout a PR thru PATCH-MENU

rootzoll:
To make it easier to test pending PRs on RaspiBlitz this adds a menu item `PR Checkout RullRequest` to the PatchMenu. The. the user just enters the number of the PR and it gets checkout as a local branch for the user to test.

### DIFF
--- a/home.admin/99updateMenu.sh
+++ b/home.admin/99updateMenu.sh
@@ -125,7 +125,8 @@ patch()
   # Patch Options
   OPTIONS=(PATCH "Patch/Sync RaspiBlitz with GitHub Repo" \
            REPO "Change GitHub Repo to sync with" \
-           BRANCH "Change GitHub Branch to sync with"
+           BRANCH "Change GitHub Branch to sync with" \
+           PR "Checkout a PullRequest to test"
 	)
 
   CHOICE=$(whiptail --clear --title "GitHub-User: ${activeGitHubUser} Branch: ${activeBranch}" --menu "" 10 55 3 "${OPTIONS[@]}" 2>&1 >/dev/tty)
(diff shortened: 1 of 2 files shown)


frennkie:
@rootzoll Shouldn't we add some "safe-guards" to make `XXsyncScript.sh` more robust? E.g. what if I have multiple remotes. What if I have unstage/uncommited changes on `~/raspiblitz` ?

frennkie:
What's the purpose of `-run` in `XXsyncScript.sh`. 

```
./XXsyncScripts.sh -run v1.6 rootzoll
```

```
./XXsyncScripts.sh v1.6 rootzoll
```

These two are exactly the same, right?

rootzoll:
> @rootzoll Shouldn't we add some "safe-guards" to make XXsyncScript.sh more robust? E.g. what if I have multiple remotes. What if I have unstage/uncommited changes on ~/raspiblitz ?

In the normal setup the script is build for is for one remote repo and just for "receiving changes" (not comitting). The commits are done usually from a laptop and then synced to the RaspiBlitz for testing (one-way). We could develop that further ... but is there anybody that is "in need" of this.

rootzoll:
In a future version the `./XXsyncScript.sh` should migrate to a something like a `blitz.patch.sh` config-script and it would make sense to then "cleanup" the parameter structure.

frennkie:
Would you mind changing it so that PR is in the front..?! So `pr1501` instead of `1501pr` ?

openoms:
> > @rootzoll Shouldn't we add some "safe-guards" to make XXsyncScript.sh more robust? E.g. what if I have multiple remotes. What if I have unstage/uncommited changes on ~/raspiblitz ?
> 
> In the normal setup the script is build for is for one remote repo and just for "receiving changes" (not comitting). The commits are done usually from a laptop and then synced to the RaspiBlitz for testing (one-way). We could develop that further ... but is there anybody that is "in need" of this.

agreeing here. 
I usually develop in the IDE on my laptop and SCP the scripts in place for testing and only committing to github when works.
Would be different of course for someone more hardcore using vim in place for example.

Issue #1482, opened 2020-08-24
Canceling out of 'Change Passwords' fails [v1.6]


shawnyeager:
I selected `Change Passwords` from the raspiblitz menu, then selected `RPC Password`, as I intended to change from a non-alphanumeric-only password. I then double-checked, realized it was alphanumeric, and selected `<Cancel>`. This fails, insisting that the password field be filled. The only way out is to close the terminal window.

rootzoll:
Like with #1458 .. will check the CANCEL option again for v1.6.1

rootzoll:
found the problem in the auto-unlock cancel. Will be fixed in v1.6.1

rootzoll:
Tested working. Will be part of v1.6.1 release - closing issue.

Issue #1458, opened 2020-08-17
Cancelling autounlock from menu fails [v1.6]


shawnyeager:
I was looking through the config options in menu, selected lnd autounlock, then, after being prompted to store the password on-device, selected `cancel`. The process continued anyway, confirming that autounlock is now active.

rootzoll:
Good find - will check the "CANCEL" option on that dialog for the next v1.6.1 service release.

ChuckNorrison:
cancel button in reboot dialogue will also result in a reboot and not cancel

rootzoll:
@ChuckNorrison checked the reboot dialog on SSH main menu and it respects my cancel.
@shawnyeager found the problem in the auto-unlock cancel. Will be fixed in v1.6.1

rootzoll:
OK Testing works. Will be part of v1.6.1 release .. closing issue.

PR #1548, opened 2020-09-15
Fix problems with dialog CANCEL

rootzoll:
fixes for #1458 and  #1482


rootzoll:
also cherry-picked into dev branch

Issue #1410, opened 2020-08-04
Fix: lndmanage 0.11.0


rootzoll:
https://github.com/rootzoll/raspiblitz/issues/1319#issuecomment-668064164

rootzoll:
@frennkie got the install/recovery of lndmanage fixed (will merge) ... but I get this error when I run it: 

```
admin@raspberrypi:~ $ manage
NOTICE: Needs at least one active channel to run without error.
to exit (venv) enter ---> deactivate
Running lndmanage for the first time.
Creating configuration folder at /home/admin/.lndmanage.
The default path can be overridden by setting the LNDMANAGE_HOME environment variable.
Detected a local lnd configuration folder /home/admin/.lnd.
Will use admin.macaroon and tls.cert from this directory.
Config file was written to /home/admin/.lndmanage/config.ini.
Traceback (most recent call last):
  File "/home/admin/lndmanage/venv/lib/python3.7/site-packages/numpy/core/__init__.py", line 22, in <module>
    from . import multiarray
  File "/home/admin/lndmanage/venv/lib/python3.7/site-packages/numpy/core/multiarray.py", line 12, in <module>
    from . import overrides
  File "/home/admin/lndmanage/venv/lib/python3.7/site-packages/numpy/core/overrides.py", line 7, in <module>
    from numpy.core._multiarray_umath import (
ImportError: libf77blas.so.3: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/admin/lndmanage/venv/bin/lndmanage", line 5, in <module>
    from lndmanage.lndmanage import main
  File "/home/admin/lndmanage/venv/lib/python3.7/site-packages/lndmanage/lndmanage.py", line 11, in <module>
    from lndmanage.lib.listchannels import ListChannels
  File "/home/admin/lndmanage/venv/lib/python3.7/site-packages/lndmanage/lib/listchannels.py", line 10, in <module>
    from lndmanage.lib.forwardings import get_forwarding_statistics_channels
  File "/home/admin/lndmanage/venv/lib/python3.7/site-packages/lndmanage/lib/forwardings.py", line 4, in <module>
    import numpy as np
  File "/home/admin/lndmanage/venv/lib/python3.7/site-packages/numpy/__init__.py", line 140, in <module>
    from . import core
[...]

PR #1552, opened 2020-09-16
#1410 fix lndmanage install

rootzoll:
updated the install process (now based on releases and checking signature) and added needed os dependency


rootzoll:
+cp dev branch

PR #1557, opened 2020-09-17
FAQ: add table of contents

openoms:
Proposing a simple autogenerated Table of Contents for the FAQ.
Can be extended and grouped later on.
Initial result:
https://github.com/openoms/raspiblitz/blob/FAQ-table-of-contents/FAQ.md

Method in VSCode (from https://stackoverflow.com/questions/11948245/markdown-to-create-pages-and-table-of-contents):

For the Visual Studio Code users the best option to use today (2020) is the Markdown All in One plugin.
To install it, launch the VS Code Quick Open (Control/⌘+P), paste the following command, and press enter.
`ext install yzhang.markdown-all-in-one`
And to generate the TOC, open the command palette (Control/⌘+Shift+P) and select the Select Markdown: Create Table of Contents option.
The TOC is automatically updated on file save. To disable please change the toc.updateOnSave option.
The indentation type (tab or spaces) of TOC can be configured per file (find it in the right bottom corner)
Note: be sure to also check the list.indentationSize option
To make TOC compatible with GitHub or GitLab, set option slugifyMode accordingly

More here: https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one


frennkie:
I think this is definitely a valuable improvement for these big markdown pages.

I just don't like the dependency on a locally installed (VS Code) tool. 

How about using GitHub actions for this? https://dev.to/fredericlemay/generate-table-of-contents-with-github-actions-3a0b I haven't tried it - but that should be able to do pretty much the same, right? This would then even work if somebody does a quick edit on the Github editor.

openoms:
@frennkie would be cool to outsource it to an automatic GitHub action.
The final outcome should be the same since the TOC is generated in a very simple format.
In any case a link can be added manually as well.
I would happy to have either, just would a like have a TOC with the FAQ.

rootzoll:
> How about using GitHub actions for this? https://dev.to/fredericlemay/generate-table-of-contents-with-github-actions-3a0b 

TOC with GitHub Actions also seems to have some setup with docker, etc ... if @openoms like to contribute the TOC on the regular generated from his local VSCode plugin thats fine with me. I will try to add now a TOC-Entry manually when extending the FAQ. So I dont see any hard dependency on a VSCode plugin.

I will merge this PR for the v1.6.1.

rootzoll:
+cp into dev branch

Issue #1554, opened 2020-09-16
Export the Connection String for BTCPayserver


LightRider5:
Would it be possible to implement an easy to use function for exporting the "Connection String" needed for any BTCPayserver to use the RaspiBlitz as underlying Lightning Node?

For example I use a "Third Party Hosted" BTCPayserver and want to accept Lightning payments.

This is the needed Connection String:
![Bildschirmfoto 2020-09-16 um 14 52 30](https://user-images.githubusercontent.com/48626097/93339887-67168500-f82c-11ea-91c9-0fdcf3e24f3e.jpg)

rootzoll:
@Animalmother9000 Does the macaroon needs to be admin? Will just invoice-macaroon work? If you just want to "receive" payments having a "not admin" macaroon would be the best security option.

rootzoll:
@Animalmother9000 this should now work - give it a try.
Run on a v1.6.0 RaspiBlitz MENU > UPDATE > PATCH > PATCHMENU > PATCH and reboot
Then go: MENU > LNDCRED > EXPORT > BTCPAY Server

This should give you the connection string.for Lightning Payments. For the xPub on on-chain payments with BTC_pAy server its still recommended to use the xpub of a hardware wallet here (more security).

Also its advised to loop-out funds regularly from the RaspiBlitz ... because the BTCPay server has still the admin macaroon and if someone hacks that server then could steal the funds from RaspiBlitz. 

I tried using the "invoices.macaroon" but that did not worked with BTCPay. If you can find out more (what macaroon or what macaroon to bake) that can be used to receive but not to send with BTCPay ... that would be a great impovement.

LightRider5:
I subscribed for a Ip2Tor bridge which worked smoothly. After that I could export the string and tried to connect the node with the BTCPayserver. I tried it several times and mostly I got an error. I attached the message below. 

![Bildschirmfoto 2020-09-19 um 10 16 31](https://user-images.githubusercontent.com/48626097/93662767-d37ac980-fa62-11ea-914f-6320739d2d59.png)

But sometime it worked and the node connected to the BTCPS. 

After that I made some test orders and the BTCPS correctly issued a Lightning invoice. I tried to pay it with three different wallets but was not able to pay it. I have to figure out the reason for that and will do more tests with different nodes and wallets. 

Now I also tried it with another BTCPS but was not able again to connect. I always get the timeout error. Maybe it is because of delays in the Tor network.

[...]

PR #1559, opened 2020-09-17
#1554 export BTCPay Connection String

rootzoll:
Adds the export of local and IP2TOR BTCPay connection string.


rootzoll:
+cp into dev branch

Issue #869, opened 2019-11-19
RaspiBlitz IPv6 support


PatrickScheich:
Is there a way to (re)configure an existing/already running RaspiBlitz to an IPv6 only configuration?

IPv4 wise I'm behind an ISP NAT so I get an external IPv4 address but will never be reachable by this.
I already configured my RaspiBlitz to use IPv6 and a DynDNS (also IPv6 only), but my Bitcoin- and Lightning-Node is only listed by the (unreachable) IPv4 address.

frennkie:
The idea is to go with TOR-by-default. IPv6 is would kind-of be the opposite direction. 

Do you still feel a need for this?

PatrickScheich:
> 
> 
> The idea is to go with TOR-by-default. IPv6 is would kind-of be the opposite direction.
ok, I had not looked into the TOR option yet

But as I "suffer" from the CGNat my ISP is giving me I had recently looked into the raspiblitz source code during my update to v1.6.0.

I have a "hardcoded" solution which is surely not desirable but may be useful for reference 

Perhaps a Setting/Option may be feasible

I would like to contribute but I'm not very keen with the way to collaborate on a public github project
So I may need some advice here

rootzoll:
@PatrickScheich If its just some additional/modified lines just paste them here as example and then we can see how complex it would be to add that as an option.

PatrickScheich:
> 
> 
> @PatrickScheich If its just some additional/modified lines just paste them here as example and then we can see how complex it would be to add that as an option.

great, I attached a Zip containing the 4 files with changes
[IPv6 changes.zip](https://github.com/rootzoll/raspiblitz/files/5057374/IPv6.changes.zip)

- _background.sh
- _bootstrap.sh
- 00infoBlitz.sh
- config.scripts/bonus.electrs.sh

basically I changed 

```
freshPublicIP=$(curl -s http://v4.ipv6-test.com/api/myip.php 2>/dev/null)
````

by 

```
freshPublicIP=$(curl -s http://v6.ipv6-test.com/api/myip.php 2>/dev/null)
```
and added the square brackets around the new IPv6 address as this is needed in `lnd.service` 

unfortunately there are usages of `publicIP` with the netcat tool which does not like the square backets, so I introduces an additional (internal) variable to get an un-brackified-version at those points.

rootzoll:
@PatrickScheich will take a look at your proposed changes and see if we can use them for v1.6.1 in any kind of way.

[...]

PR #1561, opened 2020-09-17
#869 Adding IPv6 option to config

rootzoll:


rootzoll:
+cp into v1.6 branch

frennkie:
@rootzoll ~This would have been a candicate for the "squash-merge"..!~ (you squash-merged it.. that just isn't reflect in this PR.. but it's correct in the commit history - I think I was confused about this before... need to keep it in mind).

PR #1565, opened 2020-09-19
add a 'last refresh' line to infoBlitz

### LINKED ISSUE #1161: status display has no date or time on it, data displayed is of unknown freshness
There's no way to know if the status display is current or not.   It could have frozen; a date and time display (with seconds) should be incorporated into the LCD layout.

frennkie:
Fix #1161

The change is pretty simple. I'm just nor sure about the exact wording and placement. Any better ideas?

Updated Proposal:

RPi_RefreshR

Old Proposal

RPi_Refresh

### DIFF
--- a/home.admin/00infoBlitz.sh
+++ b/home.admin/00infoBlitz.sh
@@ -308,7 +308,6 @@ ${color_yellow}               ${color_gray}${ln_channelInfo} ${ln_peersInfo}
 ${color_yellow}               ${color_gray}${ln_feeReport}
 ${color_yellow}
 ${color_yellow}${ln_publicColor}${ln_external}${color_gray}
-
 " \
 "RaspiBlitz v${codeVersion}" \
 "-------------------------------------------" \
--- a/home.admin/00infoBlitz.sh
+++ b/home.admin/00infoBlitz.sh
@@ -282,6 +282,11 @@ else
 fi
 
 sleep 5
+
+## get uptime and current date & time
+uptime=$(uptime --pretty)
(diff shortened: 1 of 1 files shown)


openoms:
Good idea!
The placement is ok, but would prefer to use a clearer time format eg.:
```
$ date +"%X %F %Z"
09:19:54 2020-09-20 BST
```

frennkie:
Yeah.. let's have that discussion..! :-D 

Seriously.. I agree that the default date string which I use here is pretty verbose. But I would say it also makes it very clear. Day of week (Sat) might also be handy. Writing out the month (Sep) will avoid anybody from ask whether 2020-10-09 is Oct. 9th or Sep. 10th.. ;-)

We could also use `date -R`: `Sun, 20 Sep 2020 10:47:26 +0200`

Anyway - for me the perfect date is definitely ISO-8601 (https://xkcd.com/1179/) ;-P

openoms:
we agree on the perfect date format since that is the only one usable for sorting.

Anyway, verbosity might make it more straighforward as you write. 
I am ok with the default format too.

frennkie:
I changed it to the RFC 5322 (email) format because I find the "distance" between the Day+Month and the Year in the default very awkward (Day, Month, Time, TZ, Year).

rootzoll:
LGTM merging for v1.6.1 release

rootzoll:
fixes #1161

rootzoll:
+cp into v1.6 branch

Issue #998, opened 2020-01-24
Tor [WARN] fixes


openoms:
Ths message in Nyx (`sudo -u bitcoin nyx`) is a returning discussion point:
```
 [WARN] Before Tor can create a control socket in "/run/tor/control", the directory "/run/tor" needs to exist, and to be accessible only by the user and group account that is running Tor. (OnOn
some Unix systems, anybody who can list a socket can connect to it, so Tor is being careful.) [406 duplicates hidden]n]
 [WARN] /run/tor is not owned by this user (bitcoin, 1002) but by debian-tor (108). Perhaps you are running Tor as the wrong user? [405 duplicates hidden]n]
```

It can be fixed by modifying the torrc (`sudo nano /etc/tor/torrc`)
to use the user debian-tor 
```
User debian-tor
```
and making the /mnt/hdd/tor dir to be owned by the user  `debian-tor`
`sudo chown -R debian-tor:debian-tor /mnt/hdd/tor`

```
 [WARN] Your log may contain sensitive information - you disabled SafeLogging, and you're logging more than "notice". Don't log unless it serves an important reason.
Overwrite the log afterwards.
```
This is fixed in torrc:
change SafeLogging 0 to:
```
SafeLogging 0 to 1
```
[WARN] Skipping obsolete configuration option "PortForwarding".
In torrc delete 
```
PortForwarding 1
```

All about torrc in: https://helpmanual.io/man5/torrc/

openoms:
Also this line: https://github.com/rootzoll/raspiblitz/blob/v1.4/home.admin/config.scripts/internet.tor.sh#L315  does nothing a  the directory changes back to be owned by debian-tor on Tor restart.

See the test:
```
admin@raspberrypi:~ $ ls -la /var/run/tor/
total 0
drwxr-sr-x  2 debian-tor debian-tor  60 Jan 24 21:40 .
drwxr-xr-x 28 root       root       880 Jan 22 13:35 ..
srw-rw----  1 debian-tor debian-tor   0 Jan 24 21:40 control

admin@raspberrypi:~ $ sudo chown -R bitcoin:bitcoin /var/run/tor/

admin@raspberrypi:~ $ ls -la /var/run/tor/
total 0
drwxr-sr-x  2 bitcoin bitcoin  60 Jan 24 21:40 .
drwxr-xr-x 28 root    root    880 Jan 22 13:35 ..
[...]

PR #1571, opened 2020-09-20
Run tor with the default user

nolith:
This PR run tor as the default user and make use of hashed password for LND and bitcoind.

It fixes some of concerns in https://github.com/rootzoll/raspiblitz/issues/998

### DIFF
--- a/home.admin/00mainMenu.sh
+++ b/home.admin/00mainMenu.sh
@@ -184,7 +184,7 @@ case $CHOICE in
             fi
             ;;
         TOR)
-            sudo -u bitcoin nyx
+            sudo -u debian-tor nyx
             ;;
         SCREEN)
             dialog --title 'Touchscreen Calibration' --msgbox 'Choose OK and then follow the instructions on touchscreen for calibration.\n\nBest is to use a stylus for accurate touchscreen interaction.' 9 48
--- a/home.admin/config.scripts/internet.tor.sh
+++ b/home.admin/config.scripts/internet.tor.sh
@@ -402,9 +402,8 @@ EOF
   fi
 
   echo "Setup logrotate"
-  if ! grep -Eq "^/mnt/hdd/tor/" /etc/logrotate.d/tor; then
-    # add logrotate config for modified Tor dir on ext. disk
-    cat << EOF | sudo tee -a /etc/logrotate.d/tor >/dev/null
+  # add logrotate config for modified Tor dir on ext. disk
+  sudo tee /etc/logrotate.d/raspiblitz-tor >/dev/null <<EOF
 /mnt/hdd/tor/*log {
         daily
         rotate 5
@@ -412,7 +411,7 @@ EOF
         delaycompress
         missingok
         notifempty
-        create 0640 bitcoin bitcoin
+        create 0640 debian-tor debian-tor
         sharedscripts
         postrotate
                 if invoke-rc.d tor status > /dev/null; then
@@ -421,7 +420,8 @@ EOF
         endscript
 }
 EOF
-  fi
+
+  sudo systemctl restart tor@default
 
   echo "OK - TOR is now ON"
   echo "needs reboot to activate new setting"
@@ -451,6 +451,10 @@ if [ "$1" = "0" ] || [ "$1" = "off" ]; then
 
   sudo systemctl enable lnd
   echo "OK"
+  echo ""
+
+  echo "*** Stop TOR service ***"
+  sudo systemctl stop tor@default
   echo ""
 
   echo "needs reboot to activate new setting"
--- a/home.admin/config.scripts/internet.tor.sh
+++ b/home.admin/config.scripts/internet.tor.sh
@@ -285,7 +285,7 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
     sudo chmod -R 700 /mnt/hdd/tor
     sudo chown -R debian-tor:debian-tor /mnt/hdd/tor
     PASSWORD_B=$(sudo cat /mnt/hdd/${network}/${network}.conf | grep rpcpassword | cut -c 13-)
-    HASHED_PASSWORD=$(tor --hash-password "$PASSWORD_B")
+    HASHED_PASSWORD=$(sudo -u debian-tor tor --hash-password "$PASSWORD_B")
 
     cat > ./torrc <<EOF
 ### See 'man tor', or https://www.torproject.org/docs/tor-manual.html
@@ -354,11 +354,6 @@ EOF
 [Service]
 ReadWriteDirectories=-/mnt/hdd/tor
 EOF
-
-    sudo -u bitcoin mkdir -p /home/bitcoin/.nyx
-    sudo -u bitcoin touch /home/bitcoin/.nyx/config
-    sudo -u bitcoin sed -i "s/^password .*//g" /home/bitcoin/.nyx/config
-    echo "password $PASSWORD_B" | sudo -u bitcoin tee -a /home/bitcoin/.nyx/config >/dev/null
 
   else
     echo "TOR package/service is installed and was prepared earlier .. just activating again"
--- a/home.admin/config.scripts/internet.tor.sh
+++ b/home.admin/config.scripts/internet.tor.sh
@@ -115,11 +115,6 @@ activateBitcoinOverTOR()
       fi
       sudo chmod 444 /home/bitcoin/.${network}/${network}.conf
 
-      sudo -u bitcoin mkdir -p /home/bitcoin/.nyx
-      sudo -u bitcoin touch /home/bitcoin/.nyx/config
-      sudo -u bitcoin sed -i "s/^password .*//g" /home/bitcoin/.nyx/config
-      echo "password $PASSWORD_B" | sudo -u bitcoin tee -a /home/bitcoin/.nyx/config >/dev/null
-
       # copy new bitcoin.conf to admin user for cli access
       sudo cp /home/bitcoin/.${network}/${network}.conf /home/admin/.${network}/${network}.conf
       sudo chown admin:admin /home/admin/.${network}/${network}.conf
@@ -287,13 +282,6 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
     #sudo rm -r -f /mnt/hdd/tor 2>/dev/null
     sudo mkdir -p /mnt/hdd/tor
     sudo mkdir -p /mnt/hdd/tor/sys
-    sudo mkdir -p /mnt/hdd/tor/web80
-    sudo mkdir -p /mnt/hdd/tor/lnd9735
-    sudo mkdir -p /mnt/hdd/tor/lndrpc9735
-    sudo mkdir -p /mnt/hdd/tor/lndrest8080
-    sudo mkdir -p /mnt/hdd/tor/lndrpc9735fallback
-    sudo mkdir -p /mnt/hdd/tor/lndrest8080fallback
-    sudo mkdir -p /mnt/hdd/tor/bitcoin8332
     sudo chmod -R 700 /mnt/hdd/tor
     sudo chown -R debian-tor:debian-tor /mnt/hdd/tor
     PASSWORD_B=$(sudo cat /mnt/hdd/${network}/${network}.conf | grep rpcpassword | cut -c 13-)
@@ -366,6 +354,11 @@ EOF
 [Service]
 ReadWriteDirectories=-/mnt/hdd/tor
 EOF
+
+    sudo -u bitcoin mkdir -p /home/bitcoin/.nyx
+    sudo -u bitcoin touch /home/bitcoin/.nyx/config
+    sudo -u bitcoin sed -i "s/^password .*//g" /home/bitcoin/.nyx/config
+    echo "password $PASSWORD_B" | sudo -u bitcoin tee -a /home/bitcoin/.nyx/config >/dev/null
 
   else
     echo "TOR package/service is installed and was prepared earlier .. just activating again"
--- a/home.admin/config.scripts/internet.tor.sh
+++ b/home.admin/config.scripts/internet.tor.sh
@@ -362,6 +362,7 @@ EOF
 
     sudo mkdir -p /etc/systemd/system/tor@default.service.d
     sudo tee /etc/systemd/system/tor@default.service.d/raspiblitz.conf >/dev/null <<EOF
+    # DO NOT EDIT! This file is generate by raspiblitz and will be overwritten
 [Service]
 ReadWriteDirectories=-/mnt/hdd/tor
 EOF
--- a/home.admin/config.scripts/internet.tor.sh
+++ b/home.admin/config.scripts/internet.tor.sh
@@ -93,15 +93,17 @@ activateBitcoinOverTOR()
       # clean all previous added nodes
       sudo sed -i "s/^main.addnode=.*//g" /home/bitcoin/.${network}/${network}.conf
       sudo sed -i "s/^test.addnode=.*//g" /home/bitcoin/.${network}/${network}.conf
-    
+
       echo "Addding TOR config ..."
       sudo chmod 777 /home/bitcoin/.${network}/${network}.conf
       echo "onlynet=onion" >> /home/bitcoin/.${network}/${network}.conf
       echo "proxy=127.0.0.1:9050" >> /home/bitcoin/.${network}/${network}.conf
       echo "main.bind=127.0.0.1" >> /home/bitcoin/.${network}/${network}.conf
       echo "test.bind=127.0.0.1" >> /home/bitcoin/.${network}/${network}.conf
-      echo "dnsseed=0" >> /home/bitcoin/.${network}/${network}.conf 
-      echo "dns=0" >> /home/bitcoin/.${network}/${network}.conf 
+      echo "dnsseed=0" >> /home/bitcoin/.${network}/${network}.conf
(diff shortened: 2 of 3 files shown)


nolith:
@openoms do you mind a preliminary review? 🙏

home.admin/config.scripts/internet.tor.sh
```diff
@@ -155,7 +163,19 @@ activateLndOverTOR()
...
+
+    # check if "tor.password" exists
+    valueExists=$(sudo cat /mnt/hdd/lnd/lnd.conf | grep -c 'tor.password=')
+    if [ ${valueExists} -eq 0 ]; then
+        echo "Adding tor config defaults to /mnt/hdd/lnd/lnd.conf"
+        PASSWORD_B=$(sudo cat /mnt/hdd/${network}/${network}.conf | grep rpcpassword | cut -c 13-)
+        sudo -u bitcoin tee -a /mnt/hdd/lnd/lnd.conf >/dev/null <<EOF
+
+[Tor]
+tor.password=$PASSWORD_B
```
nolith, line 170: This is fine to remain here in case of disactivation.

But it will not work if someone changes password b

nolith: Maybe instead of password b we should generate a password when we activate the feature?

Using password B has the advantage that a user can use it also from other software, while using a random one will require reading the cleartext password from `bitcoin.conf`, `lnd.conf`, or `.nyx/config`

home.admin/config.scripts/internet.tor.sh
```diff
@@ -338,13 +357,14 @@ EOF
...
     sudo chmod 644 $torrc
-    sudo chown -R bitcoin:bitcoin /var/run/tor/ 2>/dev/null
+    sudo chown -R debian-tor:debian-tor /var/run/tor/ 2>/dev/null
     echo ""
 
     sudo mkdir -p /etc/systemd/system/tor@default.service.d
-    if ! grep -Eq "^ReadWriteDirectories=" /etc/systemd/system/tor@default.service.d/raspiblitz.conf; then
-      echo -e "[Service]\nReadWriteDirectories=-/mnt/hdd/tor" | sudo tee -a /etc/systemd/system/tor@default.service.d/raspiblitz.conf >/dev/null
-    fi
+    sudo tee /etc/systemd/system/tor@default.service.d/raspiblitz.conf >/dev/null <<EOF
```
nolith, line 352: I think this is more clear and less error prone than the `grep` version.

This file is owned by `raspiblitz`.

Maybe we can add something like this:

```suggestion
    sudo tee /etc/systemd/system/tor@default.service.d/raspiblitz.conf >/dev/null <<EOF
    # DO NOT EDIT! This file is generate by raspiblitz and will be overwritten
```

rootzoll:
marked for v1.7 release @openoms if you think this is good to go already for v1.6.1 let me know

home.admin/config.scripts/internet.tor.sh
```diff
@@ -113,6 +115,11 @@ activateBitcoinOverTOR()
       fi
       sudo chmod 444 /home/bitcoin/.${network}/${network}.conf
 
+      sudo -u bitcoin mkdir -p /home/bitcoin/.nyx
+      sudo -u bitcoin touch /home/bitcoin/.nyx/config
+      sudo -u bitcoin sed -i "s/^password .*//g" /home/bitcoin/.nyx/config
+      echo "password $PASSWORD_B" | sudo -u bitcoin tee -a /home/bitcoin/.nyx/config >/dev/null
+
```
openoms, line 122: I don't think we should run nyx with the `bitcoin` user any more.

home.admin/config.scripts/internet.tor.sh
openoms: the solution is to change the Tor password when Password B is being changed just like with other services relying on it.
See the list around here: https://github.com/rootzoll/raspiblitz/blob/v1.6/home.admin/config.scripts/blitz.setpassword.sh#L262

home.admin/config.scripts/internet.tor.sh
```diff
@@ -265,17 +285,20 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
...
-    sudo mkdir /mnt/hdd/tor/bitcoin8332 2>/dev/null
+    sudo mkdir -p /mnt/hdd/tor
+    sudo mkdir -p /mnt/hdd/tor/sys
+    sudo mkdir -p /mnt/hdd/tor/web80
+    sudo mkdir -p /mnt/hdd/tor/lnd9735
+    sudo mkdir -p /mnt/hdd/tor/lndrpc9735
+    sudo mkdir -p /mnt/hdd/tor/lndrest8080
+    sudo mkdir -p /mnt/hdd/tor/lndrpc9735fallback
+    sudo mkdir -p /mnt/hdd/tor/lndrest8080fallback
+    sudo mkdir -p /mnt/hdd/tor/bitcoin8332
```
openoms:
```
    sudo mkdir -p /mnt/hdd/tor/web80
    sudo mkdir -p /mnt/hdd/tor/lnd9735
    sudo mkdir -p /mnt/hdd/tor/lndrpc9735
    sudo mkdir -p /mnt/hdd/tor/lndrest8080
    sudo mkdir -p /mnt/hdd/tor/lndrpc9735fallback
    sudo mkdir -p /mnt/hdd/tor/lndrest8080fallback
    sudo mkdir -p /mnt/hdd/tor/bitcoin8332
```
these folders are autogenerated by Tor on restart when the HIdden Service is configured in `torrc` so creating them is redundant (and it was also before)

home.admin/config.scripts/internet.tor.sh
openoms: go for it, more comments are a good thing

openoms:
ok looking good.
tested on a testnet node.
PR 1571 checked out
deactivated Tor
```
sudo apt remove tor tor-arm
sudo rm -rf /etc/tor
```
reactivated Tor from the menu

Tor is running well, LND and bitcoin is functional.
Only nyx is not working with the `bitcoin` user for some reason.
`sudo nyx` works - shows warning about the unnecessary root permission
`sudo -u bitcoin nyx` is not showing the log despite the correct `password` given in `/home/bitcoin/.nyx/config`

If we can figure out Nyx would be happy to see this in v1.6.1 so we are not running tor as the `bitcoin` user any more.

home.admin/config.scripts/internet.tor.sh
openoms: In case of a passwordB change:
* torrc would need a new HASHED_PASSWORD entry
* bitcoin.conf and lnd.conf would need the new Tor password (the RPC password is being changed there anyway)

home.admin/config.scripts/internet.tor.sh
nolith: Thanks, removed

home.admin/config.scripts/internet.tor.sh
nolith: Thank you. Done

nolith:
@openoms I think the problem with the log was related an issue I fixed in the last commit.

Logrotate was creating the files as `bitcoin`

home.admin/config.scripts/internet.tor.sh
```diff
@@ -265,17 +280,13 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
...
-    sudo mkdir /mnt/hdd/tor/lndrpc9735fallback 2>/dev/null
-    sudo mkdir /mnt/hdd/tor/lndrest8080fallback 2>/dev/null
-    sudo mkdir /mnt/hdd/tor/bitcoin8332 2>/dev/null
+    sudo mkdir -p /mnt/hdd/tor
+    sudo mkdir -p /mnt/hdd/tor/sys
     sudo chmod -R 700 /mnt/hdd/tor
-    sudo chown -R bitcoin:bitcoin /mnt/hdd/tor 
+    sudo chown -R debian-tor:debian-tor /mnt/hdd/tor
+    PASSWORD_B=$(sudo cat /mnt/hdd/${network}/${network}.conf | grep rpcpassword | cut -c 13-)
+    HASHED_PASSWORD=$(tor --hash-password "$PASSWORD_B")
```
openoms: the script is run with sudo, so here we need to specify the user to run Tor otherwise it will be root.
use:
`HASHED_PASSWORD=$(sudo -u admin tor --hash-password "$PASSWORD_B")`


got this funny one in my torrc:
```
$ PASSWORD_B=test
admin@tBlitz:~ $ sudo -u admin tor --hash-password "$PASSWORD_B"
16:762FA966A54EDA5B6085909EA26BFB6BFCAF00576779DB34368CD3E36E
admin@tBlitz:~ $ sudo tor --hash-password "$PASSWORD_B"
Sep 22 16:15:49.895 [warn] You are running Tor as root. You don't need to, and you probably shouldn't.
16:0CAE80B3D7B1D63E60D7722CB576FA1AB1A4C73E03EEEE747F486C7634
```

openoms (changes requested):
`sudo -u debian-tor nyx`

now works well.

Just need to make sure that `nyx` is called with that user form the `00mainMenu.sh`

home.admin/config.scripts/internet.tor.sh
nolith: It's expected to see differente outputs because there should be a salt in the hashed format that changes whern you generate it.

```
nolith@kitsune ~> tor --hash-password "foobar"
16:1F0EE7F6F5CD8F21600DD862B1662B21772ED6CCF0F0A87A4230EBDCB3
nolith@kitsune ~> tor --hash-password "foobar"
16:93D47C0497ADC30E609ACD5727A8BC27AA9C71C9F76127F80D9D616E80
nolith@kitsune ~> tor --hash-password "foobar"
16:03032718E1D0406660359658F467DC6FF146A5844BB2BDCD0D57288CC5
nolith@kitsune ~> tor --hash-password "foobar"
16:889188205C83B649608AF74E2E1EF742EEF05E07AA50840C3E18F0EB2A
```

> Sep 22 16:15:49.895 [warn] You are running Tor as root. You don't need to, and you probably shouldn't.

This should be part of stderr as it didn't ended up on my config files when I tested renaming.

So from a functional point of view, it works. But I do agree it will be better to run it with the proper user

nolith:
> `sudo -u debian-tor nyx`
> 
> now works well.
> 
> Just need to make sure that `nyx` is called with that user form the `00mainMenu.sh`

Maybe I got it wrong here. I made it so that we can run `nyx` with the `bitcoin` user, as it is today in `00mainMenu.sh`.

If I had to follow my personal taste, I would have preferred not using `bitcoin` user but make it work with `admin` only. `nyx` is clearly an admin monitoring tool, if we have to change `00mainMenu.sh` I'll suggest we make it so it's the `admin` that runs it.

WDYT @openoms ?

openoms:
`debian-tor` is perfect to run `nyx` since it is telling about Tor.
`admin` has passwordless sudo access so avoid it as much as possible. It is the user facing user.
`bitcoin` has the RPC password and access to wallets so would not use it for `nyx` either. It was only used because Tor was running with `bitcoin`.
Segrating these is a very welcome change 👍

openoms:
Since testing on restart I get the Tor error:
```
 $ sudo systemctl status tor@default
● tor@default.service - Anonymizing overlay network for TCP
   Loaded: loaded (/lib/systemd/system/tor@default.service; enabled-runtime; vendor preset: enabled)
  Drop-In: /etc/systemd/system/tor@default.service.d
           └─raspiblitz.conf
   Active: failed (Result: exit-code) since Wed 2020-09-23 11:28:00 BST; 11s ago
  Process: 25190 ExecStartPre=/usr/bin/install -Z -m 02755 -o debian-tor -g debian-tor -d /run/tor (code=exited, status=0/SUCCESS)
  Process: 25191 ExecStartPre=/usr/bin/tor --defaults-torrc /usr/share/tor/tor-service-defaults-torrc -f /etc/tor/torrc --RunAsDaemon 0 --verify-config (code=exited, status=1/FAILURE)

Sep 23 11:28:00 tBlitz systemd[1]: tor@default.service: Service RestartSec=100ms expired, scheduling restart.
Sep 23 11:28:00 tBlitz systemd[1]: tor@default.service: Scheduled restart job, restart counter is at 5.
Sep 23 11:28:00 tBlitz systemd[1]: Stopped Anonymizing overlay network for TCP.
Sep 23 11:28:00 tBlitz systemd[1]: tor@default.service: Start request repeated too quickly.
Sep 23 11:28:00 tBlitz systemd[1]: tor@default.service: Failed with result 'exit-code'.
Sep 23 11:28:00 tBlitz systemd[1]: Failed to start Anonymizing overlay network for TCP.
```
not sure what it is about, will need to investigate. Would need a clean install, but have no time currently.
Can someone have a look if it can be reproduced?

Edit: known problem:  https://github.com/rootzoll/raspiblitz/pull/1571#discussion_r494593467

home.admin/config.scripts/internet.tor.sh
openoms: Still having this issue causing the failure:
```
 $ sudo -u debian-tor /usr/bin/tor --defaults-torrc /usr/share/tor/tor-service-defaults-torrc -f /etc/tor/torrc --RunAsDaemon 0 --verify-config
Sep 24 21:29:18.940 [notice] Tor 0.4.2.7 running on Linux with Libevent 2.1.8-stable, OpenSSL 1.1.1d, Zlib 1.2.11, Liblzma 5.2.4, and Libzstd 1.3.8.
Sep 24 21:29:18.940 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://www.torproject.org/download/download#warning
Sep 24 21:29:18.940 [notice] Read configuration file "/usr/share/tor/tor-service-defaults-torrc".
Sep 24 21:29:18.940 [notice] Read configuration file "/etc/tor/torrc".
Sep 24 21:29:18.942 [warn] Failed to parse/validate config: Unknown option '16:F2A2286AC63B1D3760D6E9F2B6C3C1E1ED3684CD7F87947317BD33E006'.  Failing.
Sep 24 21:29:18.942 [err] Reading config failed--see warnings above.
```
Not surprised since my torrc has this:
```
...
ExitRelay 0

HashedControlPassword Sep 24 21:24:23.090 [warn] You are running Tor as root. You don't need to, and you probably shouldn't.
16:F2A2286AC63B1D3760D6E9F2B6C3C1E1ED3684CD7F87947317BD33E006

# Hidden Service for WEB ADMIN INTERFACE
...
```

openoms, line 288: Still need to use:
```
HASHED_PASSWORD=$(sudo -u admin tor --hash-password "$PASSWORD_B")
``` 
in line 288

openoms (changes requested):
Still need to use:
```
HASHED_PASSWORD=$(sudo -u debian-tor tor --hash-password "$PASSWORD_B")
```
in line 288

Edit: changed to `debian-tor` instead of `admin` for consistency.

home.admin/00mainMenu.sh
```diff
@@ -184,7 +184,7 @@ case $CHOICE in
             fi
             ;;
         TOR)
-            sudo -u bitcoin nyx
+            sudo -u debian-tor nyx
```
nolith, line 187: Changing this allowed us to avoid creating a nyx config file 🎉

nolith (commented):
Now everything should run as `debian-tor`

openoms (approved):

openoms:
ok, works great `tACK`

image

@rootzoll what do you think of including it to v1.6.1?

openoms:
Tested the password change too, works OK.
`apt upgrade` should not be a problem any more either.

rootzoll:
> @rootzoll what do you think of including it to v1.6.1?

LGTM lets do this ... including into my v1.6.1 test-runs.

rootzoll:
Then recovering/updating a RaspiBlitz with this new PR - LND is not starting anymore. I get:
Bildschirmfoto 2020-09-28 um 23 05 23

openoms:
it should have the tor password in lnd.conf and not look for the cookie :/

openoms:
https://github.com/lightningnetwork/lnd/blob/master/docs/configuring_tor.md
```
 --tor.password=  The password used to arrive at the HashedControlPassword for the control port. If provided, the HASHEDPASSWORD authentication method will be used instead of the SAFECOOKIE one.
```

Check for `HashedControlPassword` in `/etc/tor/torrc` and `tor.password=PASSWORD_B` in `/mnt/hdd/lnd/lnd.conf`

rootzoll:
OK the problem is when I update/recover a node running on old Tor settings with the new script. Then I have the following situation:

- lnd.conf has no Tor password in it
- bitcoin.conf has no password in it
- tor conf has the password set

I will check how to make sure this is also working in recovery/update from old state.

rootzoll:
OK now also on recovery/update from the old configs this works with my latest merge.

nolith:
Thanks for fixing this. 

So the problem was that finding an old conf on recovery it was not updating it with the tor password?

rootzoll:
> So the problem was that finding an old conf on recovery it was not updating it with the tor password?

Yes. With LND that was due to an outside script thing. Should be working now.

theotherside:
LTND: unable to initialize tor controller: HASHEDPASSWORD authentication method not supported

openoms:
> LTND: unable to initialize tor controller: HASHEDPASSWORD authentication method not supported

you need at least lnd v0.10.0. Uodate form the UPDATE menu
https://github.com/lightningnetwork/lnd/releases/tag/v0.10.0-beta

PR #1575, opened 2020-09-22
Fix broken 40addHDD.sh

nolith:
Testing https://github.com/rootzoll/raspiblitz/pull/1523 I found that the `40addHDD.sh` is broken.


adevoss:
Oops. Sorry. Should have seen that one. Thank for fixing.

rootzoll:
@nolith Good catch :)

rootzoll:
+cp into v1.6 branch

Issue #1484, opened 2020-08-24
Update: Loop v0.8.1


rootzoll:
Updated Loop to v0.8.0 wit PR #1465 

TODO: Testing before v.1.6.1

rootzoll:
OK .. looks good on testing. Will be part of v1.6.1 release. Closing issue.

PR #1576, opened 2020-09-22
Loop update to v0.8.1

openoms:
Related:
https://github.com/rootzoll/raspiblitz/issues/1484

An exisiting issue is that the changes in the latest loop break the RTL integration, so would be best to wait for RTL 0.9.1 with this:
https://github.com/Ride-The-Lightning/RTL/issues/472

### DIFF
--- a/home.admin/_commands.sh
+++ b/home.admin/_commands.sh
@@ -114,7 +114,7 @@ function status() {
 # command: balance
 # switch to the bos user for Balance of Satoshis
 function balance() {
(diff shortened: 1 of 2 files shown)


openoms:
Downgraded to v0.8.1 which is confirmed to work with RTL v0.9.0.
This could be in the  Raspiblitz v1.6.1

rootzoll:
+cp into dev branch

PR #1577, opened 2020-09-22
Faraday update to v0.2.1

openoms:
Not implemented the chain backend because it would involve storing the password be directly in the systemd service file. 
It is there as an option to fill manually.
https://github.com/lightninglabs/faraday#chain-backend

### DIFF
--- a/home.admin/_commands.sh
+++ b/home.admin/_commands.sh
@@ -133,6 +133,29 @@ function jmarket() {
   fi
 }
 
+
+# command: jmarket
+# switch to the joinmarket user for the JoininBox menu
+function jmarket() {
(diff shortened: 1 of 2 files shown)


openoms:
`frcli audit ` is quite interesting

Can output a .csv:
`$ frcli audit --csv_path .`

CAn be viewed in the CLI with:
` $ column -s, -t < node_report.csv | less -#2 -N -S`

rootzoll:
+cp into dev branch

rootzoll:
faraday 0.2.1-alpha needs at least lnd v0.11.0 ... will check for optional lower release if lnd is below like in v1.6.1 default

PR #1580, opened 2020-09-22
Check if /dev/fb1 exists and select /dev/fb0 when not.

holgern:
When using the hdmi, /dev/fb0 is available and not /dev/fb1. The change check if /dev/fb1 exists and when not , switch to /dev/fb0

### DIFF
--- a/home.admin/_bootstrap.sh
+++ b/home.admin/_bootstrap.sh
@@ -201,7 +201,11 @@ fi
 # see https://github.com/rootzoll/raspiblitz/issues/195#issuecomment-469918692
 # see https://github.com/rootzoll/raspiblitz/issues/647
 randnum=$(shuf -i 0-7 -n 1)
-sudo fbi -a -T 1 -d /dev/fb1 --noverbose /home/admin/raspiblitz/pictures/startlogo${randnum}.png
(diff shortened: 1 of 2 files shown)


rootzoll:
Interesting find. Does "using the hdmi" means that a HDMI display is "connected"? Because this could also then be used to "auto-detect" what is at the moment a "manual switch": https://github.com/rootzoll/raspiblitz/blob/v1.6/FAQ.md#can-i-run-the-raspiblitz-without-a-displaylcd

holgern:
I have connected the hdmi to a 7zoll LCD.

The hdmi is always /dev/fb0 and the spi display is always /dev/fb1,

I think it is possible to use fb1 and fb0 for autodetect.

rootzoll:
@holgern did you tested this also with the LCD hat on and Touchscreen on? In my tests with the PR it does not show the images/qr anymore and instead shows the pi-terminal. Will push this PR to v1.7 milestone.

rootzoll:
@holgern I moved your code into a new PR because it needed to be rebased on dev branch -->  #2064

So this fix got merged by the PR #2064 and will be part of the v1.7 release. Closing this one.

PR #1586, opened 2020-09-23
Adds a new service Mempool Space to the raspiblitz

holgern:
* Installs the master of https://github.com/mempool/mempool, as there are not specific releases
* Mempool needs electrs and txindex
* Ports chosen: 4080 for http api, 4081, https, 4082 tor and 4083 ssl tor
* New scripts for installing typescript and angular cli

### DIFF
--- a/home.admin/00mainMenu.sh
+++ b/home.admin/00mainMenu.sh
@@ -83,6 +83,9 @@ fi
 if [ "${loop}" == "on" ]; then
   OPTIONS+=(LOOP "Loop In/Out Service")
 fi
+if [ "${Mempool}" == "on" ]; then
+  OPTIONS+=(MEMPOOL "Mempool Space")
+fi
 if [ "${specter}" == "on" ]; then
   OPTIONS+=(SPECTER "Cryptoadvance Specter")
 fi
@@ -210,6 +213,9 @@ case $CHOICE in
             ;;
         LOOP)
             /home/admin/config.scripts/bonus.loop.sh menu
+            ;;
+        MEMPOOL)
+            /home/admin/config.scripts/bonus.mempool.sh menu
             ;;
         SPECTER)
             /home/admin/config.scripts/bonus.cryptoadvance-specter.sh menu
--- a/home.admin/00settingsMenuServices.sh
+++ b/home.admin/00settingsMenuServices.sh
@@ -372,7 +372,6 @@ if [ "${Mempool}" != "${choice}" ]; then
       sudo sytemctl start mempool
       whiptail --title " Installed Mempool Space " --msgbox "\
 The txindex may need to be created before Mempool can be active.\n
-An active electrs server needs to be running which may take also some time to be finished.\n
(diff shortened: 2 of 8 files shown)


holgern:
After testing, the backend seems not able to call the electrs server, as he do not understand tcp. I'm getting:
```
getRawMempool error. Error: socket hang up
    at connResetException (internal/errors.js:608:14)
    at Socket.socketOnEnd (_http_client.js:453:23)
    at Socket.emit (events.js:322:22)
    at endReadableNT (_stream_readable.js:1187:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21) {
  code: 'ECONNRESET'
}
```

holgern:
It seems to work with version 1.0.0, I will adapt the script.

holgern:
After playing around, I'm not so sure this is a useful extentension for raspiblitz.
* v1.0.0 does not show fees estimation and is limited in comparison to V2 (https://mempool.space)
* v1.0.0 results in a quite high load (1 cpu with around 30 % load)
* V2 needs the esplora electrs server, https://github.com/Blockstream/electrs which needs at least 300 GB storage space

I had some fun in writing the scripts, you have to decide if this is useful or not :).

rootzoll:
This looks like a sweet addition anyways. On my testing ...

I got two times asked for user interaction during install ..
Bildschirmfoto 2020-09-24 um 15 55 13
Bildschirmfoto 2020-09-24 um 16 19 32
.. the script need to work without interaction so that they can run thru during recovery. But I think that needs just some parameter somewhere.

And when finished (my txindex was already activated) I get this when calling the website:
Bildschirmfoto 2020-09-24 um 16 32 01

Any last step that I am missing? So close :)

holgern:
After activating, you need to wait 8 blocks (around 80 minutes) before something is shown.

You have to start the service by hand, when not using the menu (./config.scripts/bonus.mempool.sh on)
```
sudo systemctl start mempool
```
You can check if the service is running by:
```
sudo journalctl -f -u mempool
```

holgern:
https://github.com/wiz/mempool is also usefull,

client/client.go line 58 needs to be changed to conn, _, err := dialer.Dial("ws://localhost:8999", nil)
After building with
```
go build
```
it can be used to view the blocks and the mempool in the terminal.
After copying it to /usr/bin/mempool-cli, the following can be added to 00infoBlitz.sh line 304
```
( cmdpid=$BASHPID; (sleep 30; kill $cmdpid) & exec mempool-cli )
```
in order to show the block+mempool info to the LCD alternating to the normal screen.

grnqrtr:
> it can be used to view the blocks and the mempool in the terminal.

This would be cool to connect to one of the buttons on the touch display.

Also, one thing that I think would be beneficial even in v1.0.0 is the ability to track where a transaction is in the mempool. I like to input the transaction id and then quickly be able to see how many blocks away it is from getting mined.

rootzoll:
+cp into dev branch

PR #1594, opened 2020-09-26
fixed bad reference "/home/admin/system" that causes btcexplorer restart each 10minutes

PatrickScheich:
a bad reference to `/home/admin/system` instead of `/home/admin` causes btc-explorer to be restarted every 10 minutes

some minor changes I added for my own system are added to this PR feel free to exclude them

Note: I'm not so experienced with github. 
Maybe someone can teach me how to exclude those "private" changes from an previous PR.
Is it possible to assign lated chages to another PR?

### DIFF
--- a/home.admin/00infoLCD.sh
+++ b/home.admin/00infoLCD.sh
@@ -78,9 +78,18 @@ chain=""
 while :
     do
 
+    # save some CPU / log clutter on a "headless" installation (no LCD attached)
+    # by cutting out all the LCD stuff that otherwise will be run every 5 seconds
+    # which will spamm the DNS requests as well
+    isHeadless=$(cat ${configFile} | grep -Ec "headless=1|headless=on")
+    if [ ${isHeadless} -gt 0 ]; then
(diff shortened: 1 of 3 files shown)


rootzoll:
I moved the start script to `/home/admin/system` by purpose (in `/home/admin` its causes permission errors on script update). Everything should run clean when people do a update/recover to v1.6.1 or you uninstall/install btcrpcexplorer.

Also the loop for the LCD screen is producing a file that is needed for status report. This is to be changed in future versions - but for now I would also keep the LCD loop running for headless mode.

Closing this PR.

PatrickScheich:
@rootzoll oh, my bad...I did not get it that the `/home/admin/system` dir with the `btc-rpc-explorer.run.sh` script is automatically created. 
For some reason my `/home/admin/raspiblitz.info` had not contained the `setupStep=100` entry...so the `00infoLCD.sh` has constantly executed the "check for online" section with pinging which heavy-spammed my pihole.

PR #1595, opened 2020-09-27
Ignore docker veth networks during local ip extraction

### LINKED ISSUE #1593: "Missing DHCP" false positives when running docker
## Issue

It seems like the command used for local ip detection `ip addr | grep 'state UP' -A2 | egrep -v 'docker0' | grep 'eth0\|wlan0' | tail -n1 | awk '{print $2}' | cut -f1 -d'/'` sometimes matches `veth` networks created by docker and forces the device to reboot.

Running `ip addr` gives the following output (I've only left the interesting parts)
```
$ ip addr
1: lo: ...
2: eth0: ...
    inet 192.168.1.14/24 brd 192.168.1.255 scope global dynamic noprefixroute eth0
       valid_lft 86288sec preferred_lft 75488sec
3: wlan0: ...
    inet 192.168.1.20/24 brd 192.168.1.255 scope global dynamic noprefixroute wlan0
       valid_lft 86287sec preferred_lft 75487sec
...
5: docker0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:42:2e:6d:88:5b brd ff:ff:ff:ff:ff:ff
    inet 172.17.0.1/16 brd 172.17.255.255 scope global docker0
       valid_lft forever preferred_lft forever
...
19: veth05eaad1@if18: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-2ee85fbbdec0 state UP group default 
    link/ether 5e:eb:e2:8d:65:f5 brd ff:ff:ff:ff:ff:ff link-netnsid 1
    inet 169.254.32.251/16 brd 169.254.255.255 scope global noprefixroute veth05eaad1
       valid_lft forever preferred_lft forever
...
```

Filters used for local ip extraction will match `eth0`, `wlan0` but also `veth05eaad1`:
```
$ ip addr | grep 'state UP' -A2 | egrep -v 'docker0' | grep 'eth0\|wlan0'
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    inet 192.168.1.14/24 brd 192.168.1.255 scope global dynamic noprefixroute eth0
3: wlan0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    inet 192.168.1.20/24 brd 192.168.1.255 scope global dynamic noprefixroute wlan0
19: veth05eaad1@if18: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-2ee85fbbdec0 state UP group default 
    inet 169.254.32.251/16 brd 169.254.255.255 scope global noprefixroute veth05eaad1
```

Since the last result is assumed to be the local ip it would cause the device restart due to the following check passing https://github.com/rootzoll/raspiblitz/blob/v1.6/home.admin/_background.sh#L58

## Possible solutions

Apart from ignoring `docker0` network we may want to ignore all `veth` networks as well

## Software info

This happens on raspiblitz 1.6

glowacki-dev:
## Changes
- Update local IP extraction to ignore `veth` networks - Fixes #1593

### DIFF
--- a/home.admin/00mainMenu.sh
+++ b/home.admin/00mainMenu.sh
@@ -37,8 +37,8 @@ confirmation()
   return $answer
 }
 
-# get the local network IP to be displayed on the lCD
-localip=$(ip addr | grep 'state UP' -A2 | egrep -v 'docker0' | grep 'eth0\|wlan0' | tail -n1 | awk '{print $2}' | cut -f1 -d'/')
+# get the local network IP to be displayed on the LCD
+localip=$(ip addr | grep 'state UP' -A2 | egrep -v 'docker0|veth' | grep 'eth0\|wlan0' | tail -n1 | awk '{print $2}' | cut -f1 -d'/')
 
 # BASIC MENU INFO
 HEIGHT=17
--- a/home.admin/40addHDD.sh
+++ b/home.admin/40addHDD.sh
@@ -108,7 +108,7 @@ else
     freshPublicIP=""
   fi
   if [ ${#freshPublicIP} -eq 0 ]; then
-    localIP=$(ip addr | grep 'state UP' -A2 | egrep -v 'docker0' | grep 'eth0\|wlan0' | tail -n1 | awk '{print $2}' | cut -f1 -d'/')
+    localIP=$(ip addr | grep 'state UP' -A2 | egrep -v 'docker0|veth' | grep 'eth0\|wlan0' | tail -n1 | awk '{print $2}' | cut -f1 -d'/')
     echo "# WARNING: No publicIP information at all yet - working with placeholder : ${localIP}"
     freshPublicIP="${localIP}"
   fi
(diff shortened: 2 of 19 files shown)


rootzoll:
So far we avoided to introduce a "generic functions" file like `_functions.sh`- if this path continued this file would get bigger and bigger and harder to maintain. I am all for capsulating often used functions ... but then we have to find a special place for them in another more specialised config script. And they should not be "functions" they should return values .. so that tehre output can also be used by python scripts or later by the webUI interface.

Please cut this PR in single pieces - if possible one PR fixes one specific thing. And in the `IP extraction` we can discuss a better place.

glowacki-dev:
Yeah, I got a bit carried away with extracting this to a separate script. For now, here is a basic fix for docker networking. I'll open a separate PR with hdd calculation fix

PR #1297, opened 2020-06-28
more options for datadrive (ext4)

adevoss:
For hardware other then Pi with SD card the disk partitions are named different. Also possible to use not only first partition.

Edit:
Currently the datapartition is hardcoded to be the first partition on the drive.

Goal is to get more flexibility in choice of the datapartition. Now it has to be /dev/sdX1 if I remember correctly.
With this pr it can be a partition on a disk with a different name (as on linux kvm vm's) and it can also be partition 2-9.
The idea came when I wanted to test on a amd64 vm.

I don't know if there are more diskname conventions out there. I use linux/kvm vm's on amd64.
Let me know if there are. It should not be hard to add those.

The approach is replacing the hardcoded diskname with code that searches for more partitions. The minimum disk size is replaced by minimum partition size. That seems more appropriate to me.

The gain is more configurations or setups can be used.

PR #1603, opened 2020-09-30
Fixing detection of HDD/SSD without partition 

rootzoll:
fixing that after #1297 the `blitz.datadrive.sh` script was not detecting hdd/ssd without any partiotion anymore


rootzoll:
+cp into dev branch

PR #1605, opened 2020-09-30
 LCD: show JoinMarket stats if the Yield Generator is running 

openoms:
* Updating joininBox to 0.1.7
* shows the JMstats only if JM installed and YG running. Otherwise display the LND URI as before.
Example (masked):
image


rootzoll:
+cp into dev branch

PR #1612, opened 2020-10-02
Add packages to install-script

adevoss:
Package is needed for formatting datadrive


rootzoll:
As this is a change to the `build.sh` file ... this would either wait for the next major release v1.7 or if needed to fix a bug adding the install into the `datadrive` script to make sure dependency is getting installed before use.

adevoss:
Think there is no need to rush this pr. It is needed for formatting. Simple to do manually.
I don't know if it is part of standard Raspbian install. It was not installed when I installed Debian without GUI.

My preference would be build_sdcard.sh script because there is where all the other packages are installed/removed.

rootzoll:
> Think there is no need to rush this pr. It is needed for formatting. Simple to do manually.

then moving to v1.7 milestone

rootzoll:
With the emergency update of LND we needed to change the sd card build - so adding this also to fix BTRFS.

PR #1613, opened 2020-10-03
change ascii art for JoinMarket stats

grnqrtr:
This was my attempt at a simpler ascii art for the JoinMarket stats.
It's just a simple JM underlined with a chain.

joinmarket

### DIFF
--- a/home.admin/00infoBlitz.sh
+++ b/home.admin/00infoBlitz.sh
@@ -291,6 +291,7 @@ if [ $(sudo -u joinmarket pgrep -f "python yg-privacyenhanced.py $YGwallet --wal
   JMstatsL3=$(sed -n 3p < "$JMstats")
   JMstatsL4=$(sed -n 4p < "$JMstats")
   lastLine="\
+${color_gray}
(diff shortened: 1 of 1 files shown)


openoms:
Tested, but got an issue on the physical touchscreen.
The xterm does not seem to support the unicode characters.
@frennkie is there an acceptable solution to this?
image

Looks great when the touchscreen is not active:
image

holgern:
The touchscreen uses xterm: https://github.com/rootzoll/raspiblitz/blob/v1.6/home.admin/BlitzTUI/blitztui/main.py#L157 which has a problem with unicode chars:
https://unix.stackexchange.com/questions/196152/xterm-not-displaying-unicode

A fast solution is to replace xterm by uxterm, I will provide a PR.

holgern:
https://github.com/rootzoll/raspiblitz/pull/1614

openoms:
Confirmed to work with https://github.com/rootzoll/raspiblitz/pull/1614
Ready for v1.6.1 @rootzoll

rootzoll:
+cp into dev branch

PR #1614, opened 2020-10-03
replace xterm by uxterm for unicode support

holgern:
https://invisible-island.net/xterm/manpage/uxterm.html

### DIFF
--- a/home.admin/BlitzTUI/CHANGELOG.md
+++ b/home.admin/BlitzTUI/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+## [0.49.0] - 2020-10-03
+### Add
+- xterm is replaced by uxterm, so that unicode chars are correctly displayed
 
 ## [0.48.1] - 2020-05-30
 ### Add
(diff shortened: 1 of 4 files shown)


openoms:
Thank you for the PR, but did you test this?

I have changed `xterm` to `uxterm` in the 
`./python3-env-lnd/lib/python3.7/site-packages/blitztui/main.py`
and restarted with:
`config.scripts/blitz.touchscreen.sh calibrate`
and the unicode characters are still not showing.

holgern:
It seems that addionally unicode fonts need to be loaded. I will try and update the PR.

holgern:
@openoms I fixed it now. I tested it, it works for me.
You need to install `sudo apt-get install -y fonts-inconsolata` for testing. I added the installation of the package to blitz.touchscree.sh

holgern:
I switched to the Terminus font, which has a better appearance:
```
sudo apt-get install xfonts-terminus
```

openoms:
the unicode characters now show, but the physical touchscreen has completely fallen apart due to the size difference in the font. 
Will need to play with this a bit more.
I presume @holgern you don't have an LCD yourself?

holgern:
Have you set ` '-fs', '9'` after `'-fa', 'Terminus'` ? I have a LCD and when using size 9 it looks ok.

openoms:
Ok, my problem was not running :
`sudo apt-get install xfonts-terminus`
on the same node where editing the `main.py`.

Now works and looks awesome, unicode characters unlocked!
image

rootzoll:
OK my tests of this PR look also good .. will merge and be part of v1.6.1

rootzoll:
+cp into dev branch

PR #1617, opened 2020-10-03
introduce "headless" mode for H/W without LCD

PatrickScheich:
this will reduce log clutter and CPU utilization
by running the whole LCD specific stuff every 5 minutes instead of 5 seconds

### DIFF
--- a/home.admin/00infoLCD.sh
+++ b/home.admin/00infoLCD.sh
@@ -78,6 +78,16 @@ chain=""
 while :
     do
 
+    # save some CPU / log clutter on a "headless" installation (no LCD attached)
+    # by running the LCD stuff each 5 minutes that otherwise will be run every 5 seconds
+    # 
+    # just insert a line "headless=on" into your "/mnt/hdd/raspiblitz.conf" file
+    # 
+    isHeadless=$(cat ${configFile} | grep -Ec "headless=1|headless=on")
+    if [ ${isHeadless} -gt 0 ]; then
+      sleep 600
+    fi
+
     ###########################
     # CHECK BASIC DATA
     ###########################   
--- a/home.admin/00infoLCD.sh
+++ b/home.admin/00infoLCD.sh
@@ -83,8 +83,9 @@ while :
     # 
     # just insert a line "headless=on" into your "/mnt/hdd/raspiblitz.conf" file
     # 
-    isHeadless=$(cat ${configFile} | grep -Ec "headless=1|headless=on")
+    isHeadless=$(cat "${configFile}" 2>/dev/null | grep -Ec "headless=1|headless=on")
     if [ ${isHeadless} -gt 0 ]; then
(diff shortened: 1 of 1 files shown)


openoms:
Good idea. Just to note two things:
* ~~if there is `isHeadless` entry in the config this will throw an error every 5 seconds.~~
* even headless (or remote) I find having `status` in a small part of a tmux window very useful so would not opt in.

PatrickScheich:
@openoms the entry in the config is `headless=on` or `headless=off`.  `isHeadless` is an internal variable in the script.

I'm not so familiar with tmux...can you explain how to get such a tmux window with the status info in a ssh session.

openoms:
for tmux see an example description and links here: https://github.com/openoms/bitcoin-tutorials/tree/master/joinmarket#running-in-the-background-with-tmux
basicly type:
`tmux` ,then `status`
screenshot:
image

openoms:
> @openoms the entry in the config is `headless=on` or `headless=off`. `isHeadless` is an internal variable in the script.

for this I meant to write:
* if there is no `headless=` entry in the config this will throw an error every 5 seconds.

so to introduce this option the `headless=off` default value  needs to be placed in the `raspiblitz.conf` during the installation.

PatrickScheich:
> 
> 
> > @openoms the entry in the config is `headless=on` or `headless=off`. `isHeadless` is an internal variable in the script.
> 
> for this I meant to write:
> 
>     * if there is no `headless=` entry in the config this will throw an error every 5 seconds.
> 
> 
> so to introduce this option the `headless=off` default value needs to be placed in the `raspiblitz.conf` during the installation.

now I get it. I took a look into `journalctl -f` but had not been able to make out the occurrence of such error
I removed the `headless=on` from config file beforehand
the script snippet 
```
    isHeadless=$(cat ${configFile} | grep -Ec "headless=1|headless=on")
    if [ ${isHeadless} -gt 0 ]; then
      sleep 600
    fi
```
will cause an error if the configFile is not accessible, I will address this

PatrickScheich:
hopefully this will now work properly in any case

- `headless=on` in conf file  => sleep 600 seconds
- every other case (`headless=off` or no `headless=` line present) should be treated as "ignore this and carry on as normal"

rootzoll:
I just marked this to be remembered once we continue looking into `auto-detect` of LCD hat & HDMI.
https://github.com/rootzoll/raspiblitz/issues/1265#issuecomment-704513743

rootzoll:
I will merge but change from `600` to `60` seconds ... so that for example someone connects a HDMI after startup or uses a tmux also still gets some update and also status values are not getting too old. Should be a good enough delay to reduce the CPU load still.

rootzoll:
headless mode improved/changed ... see: https://github.com/rootzoll/raspiblitz/issues/2120#issuecomment-814247408

PR #1620, opened 2020-10-04
remove the square brackets from IP in a oneliner

PatrickScheich:
small change in `internet.sh` to remove the square brackets around an IPv6 address


rootzoll:
+cp into dev branch

PR #1623, opened 2020-10-04
restart several services after publicIP change

PatrickScheich:
when the public IP has changed these services should be restarted

- bitcoind in case of IPv6 only, as the eth0 IPv6 (global) is the publicIP
  and this reflects in the node configuration
- btc-rpc-explorer (IPv6 only) as it needs a restart to pickup the new IP
  in its "Node Status" page
- LND to pickup the new public IP

### DIFF
--- a/home.admin/_background.sh
+++ b/home.admin/_background.sh
@@ -94,7 +94,11 @@ do
 
   ####################################################
   # RECHECK PUBLIC IP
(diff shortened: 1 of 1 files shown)


rootzoll:
Sounds good - thanks for testing and improving.

rootzoll:
+cp into dev branch

PR #1615, opened 2020-10-03
Add 'hddRaspiData' when datadrive is mounted

adevoss:
Fixes script error in 30initHDD.sh line 29

Issue #1235, opened 2020-06-03
[Research] Boot option from SSD


openoms:
All applications with their own users could be built on the disk to spare SDcard write cycles.
In case of USB3 connection and SSD there is a significant gain on the speed as well.

For example with ThunderHub I've tested:
```
# create thunderhub user
sudo adduser --disabled-password --gecos "" thunderhub
# move the home directory to the disk
sudo usermod -m -d /mnt/hdd/app-storage/thunderhub thunderhub
# symlink to default folder
sudo ln -s /mnt/hdd/app-storage/thunderhub /home/thunderhub
```

but resulted in npm permission errors on installation.

Same when instead of the symlink I have swapped all `/home/thunderhub` paths to `/mnt/hdd/app-storage/thunderhub`

Will try with other services.

The ones with long build times like ElectRS and RTL would benefit greatly being built on the disk.

openoms:
Running completely off the SSD is now possible as @hermanlim described in https://github.com/rootzoll/raspiblitz/issues/1205#issuecomment-635095566 
Link to how to: https://www.tomshardware.com/how-to/boot-raspberry-pi-4-usb

Seems that it would need a dedicated (first) partition on the SSD so would only be realistic for new setups.

In that case the SDcard could be kept solely for firmware+OS updates and maybe for infrequent security backups for the SCB and other valuable data.
The SDcard contents could be cloned to the boot partition of the SSD during a first boot, update or recovery -> more work on the  blitz.datadrive.sh

@rootzoll what do you think?

rootzoll:
@openoms yes had the same realization. So the sd card will become as a way to insert new updates, but the OS is running from the SSD. This way we can even use cheaper sd card again, which would also lower the price :) 

We can do this together with introducting a new default SSD partition layout similar as we do for experimental BTRFS ... so we would have:

- a os-partition (around 32GBs)
- a data-data partition
- a data-storage partition
- a temp-partition

hermanlim:
USB thumb drive which is being used for the experimental BTRFS setup could also be partitioned to allow it to serve as an update vector. This would remove completely the need for the SD card once boot from USB is enabled by default on RPi4...

Even if not using BTRFS, I believe more people find it easier to access something that can write onto USB thumbdrive media as opposed to SD card since pretty much all computers come with USB drives but not all computers come with SD slot.

rootzoll:
[...]

PR #1628, opened 2020-10-06
Fix btrfs for PR #1615

holgern:
This fixes :
```
# RASPIBLITZ DATA DRIVE Status

# BASICS
isMounted=1
isBTRFS=1
ls: cannot access '/mnt/hdd/WORKINGDIR': No such file or directory
```
when using `sudo ./config.scripts/blitz.datadrive.sh status` on btrfs

### DIFF
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -137,7 +137,7 @@ if [ "$1" = "status" ]; then
         # BTRFS is working with subvolumnes for snapshots / ext4 has no SubVolumes
         subVolumeDir=""
         if [ "${hddFormat}" = "btrfs" ]; then
-          subVolumeDir="/WORKINGDIR"
+          subVolumeDir="/"
         fi
 
         # temp mount data drive
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -137,7 +137,7 @@ if [ "$1" = "status" ]; then
         # BTRFS is working with subvolumnes for snapshots / ext4 has no SubVolumes
         subVolumeDir=""
         if [ "${hddFormat}" = "btrfs" ]; then
-          subVolumeDir="/"
+          subVolumeDir="/WORKINGDIR"
         fi
 
         # temp mount data drive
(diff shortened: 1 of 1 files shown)


adevoss:
@holgern I did #1615 
I have no experience with BTRFS so I just left that part as it was.

How can I test btrfs? When I start a clean installation on my vm the datadrive is being formatted ext4.
Can't remember seeing an option during installation. I won't create an pr about btrfs right away.
Just to what it is.

holgern:
You need a USB-stick which has the same size as the SD card. The stick needs to be plugged in and then you see the option.

holgern:
WORKINGDIR is a subvolume in the btfrs and not a directory

holgern:
I changed the wrong line, now it changes the correct lines.

adevoss:
@holgern That's why I don't see the option when installing on my vm (no usb stick). Thanks.
Unfortunately I can't test/experiment with btrfs because of that.

holgern:
I should be possible to test btrfs when there is a third disc available
```
    # find the possible drives that can be used as 
    drivecounter=0
    for disk in $(lsblk -o NAME,TYPE | grep "disk" | awk '$1=$1' | cut -d " " -f 1)
    do
      devMounted=$(lsblk -o MOUNTPOINT,NAME | grep "$disk" | grep -c "^/")
      # is raid candidate when not mounted and not the data drive cadidate (hdd/ssd)
      if [ ${devMounted} -eq 0 ] && [ "${disk}" != "${hdd}" ]; then
        sizeBytes=$(lsblk -o NAME,SIZE -b | grep "^${disk}" | awk '$1=$1' | cut -d " " -f 2)
        sizeGigaBytes=$(echo "scale=0; ${sizeBytes}/1024/1024/1024" | bc -l)
        vedorname=$(lsblk -o NAME,VENDOR | grep "^${disk}" | awk '$1=$1' | cut -d " " -f 2)
        mountoption="${disk} ${sizeGigaBytes} GB ${vedorname}"
        echo "raidCandidate[${drivecounter}]='${mountoption}'"
        drivecounter=$(($drivecounter +1))
      fi
    done
    echo "raidCandidates=${drivecounter}"`
```
btrfs is possible when raidCandidates is 1 and sizeGigaBytes > 14

rootzoll:
merged and marked btrfs again for testing

adevoss:
> You need a USB-stick which has the same size as the SD card. The stick needs to be plugged in and then you see the option.

@holgern 
Thought about your answer a little bit more and read the FAQ (https://github.com/rootzoll/raspiblitz/blob/v1.6/FAQ.md#how-to-use-btrfs-on-raspiblitz). I think btrfs should be possible with a second virtual disk.
Because it's experimental and because it is the first time I will work with btrfs no pr from me soon.

@rootzoll Will btrfs be the default or a second option in the future (not hidden as mentioned in the FAQ)? If yes, I will keep that in mind when experimenting. If no, then I keep it as a personal experiment.

rootzoll:
> Will btrfs be the default or a second option in the future (not hidden as mentioned in the FAQ)? 

making btrfs an option first or and maybe the default later was the intent when starting on that development - but then realized that this needs some more work until it would really be ready (work on data recovery etc) and also it should work together with the future planned Boot option from SSD see #1235. So its just an experimental feature for now ... but would love to see it getting an option for the future.

PR #1635, opened 2020-10-08
bitcoin.conf: add peerbloomfilters for bisq

openoms:
I am running my raspiblitz with these settings for a while and my and other people's Bisq is served without any problems.

TG FAQ entry previously:
```
How to connect Bisq to your own node

Edit the bitcoin.conf:  
sudo nano /mnt/hdd/bitcoin/bitcoin.conf

add the line:
peerbloomfilters=1

Restart bitcoind:  
sudo systemctl restart bitcoind

Get the bitcoin Tor address (same as shown on the display):
bitcoin-cli getnetworkinfo | jq .localaddresses

Set Custom Bitcoin Node in Bisq

Download, Install, and run Bisq on your preferred Operating System.
Go to Bisq Settings, Network Info, and choose "Custom Bitcoin Node".
Enter your the raspiblitz bitcoind onion address and then you will be asked to restart Bisq.

Example:  
weburlxxx123.onion:8333

Source:
wiki.ronindojo.io/Bisq
```

### DIFF
--- a/home.admin/assets/bitcoin.conf
+++ b/home.admin/assets/bitcoin.conf
@@ -8,6 +8,7 @@ server=1
 daemon=1
 txindex=0
 disablewallet=1
+peerbloomfilters=1
 
 # Connection settings
 rpcuser=raspibolt


openoms:
This setting has been deactivated by default in Bitcoin Core: https://bitcoincore.org/en/releases/0.19.0.1/
>To eliminate well-known denial-of-service vectors in Bitcoin Core, especially for nodes with spinning disks, the default value for the -peerbloomfilters configuration option has been changed to false. This prevents Bitcoin Core from sending the BIP111 NODE_BLOOM service flag, accepting BIP37 bloom filters, or serving merkle blocks or transactions matching a bloom filter. Users who still want to provide bloom filter support may either set the configuration option to true to re-enable both BIP111 and BIP37 support or enable just BIP37 support for specific peers using the updated -whitelist and -whitebind configuration options described elsewhere in these release notes. For the near future, lightweight clients using public BIP111/BIP37 nodes should still be able to connect to older versions of Bitcoin Core and nodes that have manually enabled BIP37 support, but developers of such software should consider migrating to either using specific BIP37 nodes or an alternative transaction filtering system. (#16152)

In case of a personal node the benefits to be able to serve Bisq outweighs the risk of a DOS attack in my opinion.
Don't think that this would be needed to be optional.

rootzoll:
Having this as part of the asset would mean that just fresh-setuped RaspiBlitzes would run with that ... and those mostly even run with SSD. Having this as an option in settings menu would be even better - also for people to see if its activated or not ... but as a quick default change I agree to have this now.

rootzoll:
+cp into dev branch

PR #1636, opened 2020-10-08
specter: add blockfilterindex

openoms:
as discussed in:
https://github.com/rootzoll/raspiblitz/issues/1411


rootzoll:
I think its good to turn the blockfilterindex on with activating specter ... but I would not turn blockfilterindex off with deactivatinf because this would also turn something off that the user (or another app in the future) would have turned on manually.

openoms:
ok, so should we leave that couple of GB blockfilter data as well when deleting all the Specter data? 
Sounds like you would prefer having a similar script to the blockfilterindex as there is for the txindex?
Could be just activating now and push the two-way script to v1.7.

rootzoll:
> ok, so should we leave that couple of GB blockfilter data as well when deleting all the Specter data?

your right - for now the v1.6.1 release it might be best to just delete everything on switching off ... but I open a new ticket #1639 for v1.7 where it should become a seperate setting and the we can make it optional on uninstall.

So thats merge this like it is for v1.6.1

rootzoll:
+cp into dev branch

Issue #1581, opened 2020-09-22
[Feature Request] Add Circuit Breaker - it is to Lightning what firewalls are to the internet


openoms:
A project by Joost Jager announced in this thread: https://twitter.com/joostjgr/status/1308414374789427200
https://github.com/lightningequipment/circuitbreaker

Needs at least lnd v0.11.0

rootzoll:
Looks like a good fit for v1.7 when >=lnd 0.11.0 will be standard ... lets keep it on that milestone list.

openoms:
Starting off with some notes:
```
# create dedicate user (or just use admin to test first)
# check Go and PATH
# add symlink for .lnd and permissions for macaroons
git clone https://github.com/lightningequipment/circuitbreaker.git
cd circuitbreaker
go install
# create config file
# ~/.circuitbreaker/circuitbreaker.yaml
# include example:
# maxPendingHtlcs: 5

# create systemd service or run:
circuitbreaker --network ${chain}net
```

rootzoll:
Now that lnd 0.11.0 will be a foreced update for v1.6.1 we can include Circuit Breaker already as config-script.
Will be part of v1.6.1 - use to activate: `/home/admin/config.scripts/bonus.circuitbreaker.sh on`

More improvents planned for v1.7 release: see #1654

PR #1649, opened 2020-10-10
Add the CircuitBreaker

openoms:
Running well so far.

Thinking that the setting menu would be a good place for it for now to activate and leave.

Test with running the script in this PR:
```
# download
wget https://raw.githubusercontent.com/openoms/raspiblitz/circuitbreaker/home.admin/config.scripts/bonus.circuitbreaker.sh
# run
bash bonus.circuitbreaker.sh on
# start in the background
sudo systemctl start circuitbreaker
```

Related issue:
https://github.com/rootzoll/raspiblitz/issues/1581

### DIFF
--- a/home.admin/config.scripts/bonus.circuitbreaker.sh
+++ b/home.admin/config.scripts/bonus.circuitbreaker.sh
@@ -1,0 +1,134 @@
+#!/bin/bash
+
+# command info
+if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
+ echo "config script to switch the circuitbreaker on or off"
+ echo "bonus.circuitbreaker.sh [on|off|menu]"
+ exit 1
+fi
+
+source /mnt/hdd/raspiblitz.conf
+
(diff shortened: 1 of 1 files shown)


rootzoll:
I tested PR on my node ... install over the SETTINGS menu failed - tried directly with config script: install worked.

@openoms  Also because there is no menu entry after install in main menu, how to use it is a bit confusing still (start systemd and then what?) and its stil a very early ALPHA release I would recommed: Deactivate the install from SETTINGS menu for now but have it as config-script for v1.6.1 release ... so people can experiment & report best practice. Based on that we can then add it to the SSH-GUI for v1.7.

openoms:
agreed, good to have an as install script, but no need to put it into the menu.
I use it as well with the default setting just to set an forget.
It should be an entry in lnd.conf really.

openoms:
ok removed the menu entries from the PR, only the install script left.
Did not have enough time to test the menu anyway and as you say the how-to-use is very limited now so config script only it is.

rootzoll:
+cp into dev branch

Issue #1579, opened 2020-09-22
Experimental Feature: Vagrant Support for Development


rootzoll:
As PR #1523 offers - this aims to offer the possibility to run RaspiBlitz in a Vagrant/VirtualBox VM on your development laptop. This is still experimental in v1.6.1 and should get improved towards the v.17 release. See also #1578

- [X] Merge PR
- [X] Test PR on on own laptop setup for v1.6.1 release

rootzoll:
Following this README to setup vagrant: https://github.com/rootzoll/raspiblitz/blob/v1.6/alternative.platforms/amd64/README.md

rootzoll:
If no further feedback ... merge #1657 and walk with this for v1.6.1 release.

rootzoll:
OK got a RaspiBlitzVM from vagrant working on my Laptop - important: Run as clearnet Tor seems not to work.
So experimental Vagrant support will be part of v1.6.1 release. Closing this issue - while for future tasks having these open:
- #1670 documentation & pre-synced data disk VDI file
- #1671 make vagrantVM also work with Tor

PR #1657, opened 2020-10-12
Vagrant: Updates for MacOS & Packer >1.6.0

geco91:


rootzoll:
This is in regard to #1579

@nolith & @adevoss to make it build on macos the packer-file needed to be fixed for packer >= 1.6.0 also the ISO debian file needed to be updated (the old was no longer available). Please check if those changes also work for you.

rootzoll:
@nolith & @adevoss also I think I will change two further things on the vagrant file: 

1. I will remove the 500GB data drive that is listed in the vagrant file and let the user manually add a datadrive on the virtualbox UI. Because that way the datadrive file can be stored on an external drive of the laptop - in my case its a 1TB hard drive. That I just keep connected while running the VM. That way all that big blockchain data stuff stays off my laptop and people can use whatever thay have available.

2. I will remove the USB3 driver modification .. because the data drive is connected thru the internal VirtualBox HDD connector. So even I use an external data drive on my laptop to the VM it looks like an internal drive and so no special USB driver is needed. Because on my Mac I would need to install the Oracle Extension Package to the VM to get USB3 working and I get the feeling that complicates things and my even make distibution of the vagrant box more complicated from licenses .. so why make it more complicated if its working without? This way I can also remove the experimental "disks" flag from vagrant.

3. Changing from NAT to "Host-only" networking .. this will make it easier to test additional Apps (runnign on the VirtualBox RaspiBlitz) from the laptops browser. Was shortly thinking about "Bridge-Network" (so that VM would also be reachable from other devices in the same local network) but that might run into problems when developing mobile with the laptop being connected to hotel or other public hotspots. Host-only" networking seems like a good enough compromise here.

I am testing the setup right now ... but I think thats a good start :) Thanks much for getting the ball rolling.

nolith:
>  to make it build on macos the packer-file needed to be fixed for packer >= 1.6.0 also the ISO debian file needed to be updated (the old was no longer available). Please check if those changes also work for you.

I think we should consider publishing the box images online when we tag a new version, so only who wants to develop on dev branch needs to rebuild it from scratch.

> 1. I will remove the 500GB data drive that is listed in the vagrant file and let the user manually add a datadrive on the virtualbox UI. Because that way the datadrive file can be stored on an external drive of the laptop - in my case its a 1TB hard drive. That I just keep connected while running the VM. That way all that big blockchain data stuff stays off my laptop and people can use whatever thay have available.

This makes a lot of sense.

> 2. I will remove the USB3 driver modification .. because the data drive is connected thru the internal VirtualBox HDD connector. So even I use an external data drive on my laptop to the VM it looks like an internal drive and so no special USB driver is needed. Because on my Mac I would need to install the Oracle Extension Package to the VM to get USB3 working and I get the feeling that complicates things and my even make distibution of the vagrant box more complicated from licenses .. so why make it more complicated if its working without? This way I can also remove the experimental "disks" flag from vagrant.

Good point, I already had the extension installed so didn't have to do the extra step.

I added the USB driver not for the extra drive, but for supporting HW wallets with specter, but I had no time to test it out properly

> 3. Changing from NAT to "Host-only" networking .. this will make it easier to test additional Apps (runnign on the VirtualBox RaspiBlitz) from the laptops browser. Was shortly thinking about "Bridge-Network" (so that VM would also be reachable from other devices in the same local network) but that might run into problems when developing mobile with the laptop being connected to hotel or other public hotspots. Host-only" networking seems like a good enough compromise here.
> 

👍 

> I am testing the setup right now ... but I think thats a good start :) Thanks much for getting the ball rolling.

Thank you @rootzoll for making this possible

PR #1668, opened 2020-10-14
use btrfs on vm

adevoss:
When there is a second datadrive btrfs is used


rootzoll:
What is the idea behind this? The drive may be virzualized anyway - does make BTRFS sense in a VM setup?

adevoss:
When I looked how to get btrfs I had to make these changes.

If it makes sense in a vm setup I don't know if you ask this question.
As I understand it btrfs is a fs with redundancy. So if the datadrive file breaks there is the extra disk(file) to restore from.
Of course you have to make backups but that is no different with physical drives.

So please give me a link or explain if I don't understand the whole story of using btrfs.

rootzoll:
> When I looked how to get btrfs I had to make these changes.

Having a second drive connected is just a hidden switch to activate the experimental BTFS option. This will later be replaced (if BTRFS gets out of experimental state) - so there is not a pressing need to make this switch working with a VM setup.

> As I understand it btrfs is a fs with redundancy. So if the datadrive file breaks there is the extra disk(file) to restore from. Of course you have to make backups but that is no different with physical drives.

If both virtual disks are just speperate files on the same physical drive it does not deliver the hardware redundance of a RAID. If you run the VM with physical seperated data drives the redundancy can make sense. But thats already a "PRO VM setup" we can target for once the VM support & BTRFS atre both out of experimental.

I would like to keep those two experimental features sepereate for now to reduce complexity (pushing this PR to a future release) until both are stable and then we can combine them. But if you already see a design decession that is important to make early so that both can work together please share.

adevoss:
> so there is not a pressing need to make this switch working with a VM setup.

The switch works but the btrfs formatting of disks does not without this pr and #1612.
I just tried btrfs and created this pr because of the bugs I got. IMO it's not polite to just report an issue and let you solve it, if the reporter of the issue can solve it.
And it was fun to do. It's up to you if you want to use the pr and when.

PR #1673, opened 2020-10-14
Update LNbits install to new frameworks

rootzoll:
Does not fix all the problems but supports the new Quart framework and hypercorn server needed for newer LNbits releases. Install is set to tag `raspiblitz` in the lnbits repo ... this needs still to be set to a new framework commit and also later on updated to commit that maes voucher work.


rootzoll:
+cp into dev branch

PR #1677, opened 2020-10-16
fix letsencrypt with duckdns

ChuckNorrison:
Fix duckdns + letsencrypt:

If duckdns was choosen from SUBSCRIBE -> LetsEncrypt, the setup gets blocked with 

> not supported yet: duckdns


rootzoll:
Thanks - good catch.
+cp into dev branch

PR #1684, opened 2020-10-19
Update mempool installation script to use v1.0.1

wiz:


rootzoll:
Thanks. Test worked - will include in v1.6.1 release.

PR #1687, opened 2020-10-19
circuitbreaker: update to v0.2.0 + update option

openoms:
featuring virtual hodlFees: https://twitter.com/joostjgr/status/1317546071984427009
+ update option


openoms:
@rootzoll this is still CLI only, but would be nice to have the improved update options in v1.6.1 if there will be any modification on the RC3

PR #1697, opened 2020-10-21
Fix for btrfs migration with raid (usb drive) does not work

holgern:
This prevents that the usbdrive is identified as hdd

### DIFF
--- a/home.admin/_bootstrap.provision.sh
+++ b/home.admin/_bootstrap.provision.sh
@@ -596,7 +596,11 @@ source <(sudo /home/admin/config.scripts/blitz.datadrive.sh status)
 # update /etc/fstab
 echo "datadisk --> ${datadisk}" >> ${logFile}
 echo "datapartition --> ${datapartition}" >> ${logFile}
-sudo /home/admin/config.scripts/blitz.datadrive.sh fstab ${datapartition} >> ${logFile}
+if [ ${isBTRFS} -eq 0 ]; then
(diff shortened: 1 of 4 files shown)


adevoss:
@holgern Why the change in blitz.datadrive.sh? When there are no partitions why the test for the biggest partition?
Did a clean install without errors. I will test again on my VM.

holgern:
My 32 GB USB drive was detected as hdd. The changes on blitz.datadrive.sh prevent that the smaller USB which is /dev/sdb with btrfs filesystem is overwriting hdd, hddDataPartition and sizeDataPartition, due to testpartitioncount  == 0 .

testpartitioncount  was 0 for the 32 GB USB drive

After applying this PR, I could migrate from 1.6 to 1.6.1 with BTRFS and Raid enabled (32GB USB drive).

adevoss:
Interesting. Tested rootzoll/v1.6 with 'usb drive' /dev/vda, dev/vdb and /dev/vdc (edit: maybe disk order is the issue). All were fine.
The 'v' or 's' in the disk name should not make any difference, I think.

adevoss:
I think the root of this error is in blitz.migration.sh not in blitz.datadrive.sh
blitz.datadrive.sh should be called with the correct parameters (different with ext4/btrfs)
Working on a pr. (Edit: See #1706)

adevoss:
@holgern Is 'isBTRFS' (line 322) the old format or the format from the menu? I think it's the old one. I have a different solution in #1706.

rootzoll:
+cp into dev

PR #1707, opened 2020-10-22
migration - get hostname from config file

adevoss:
Why the search and remove? blitzname was empty.


rootzoll:
I am not exactly sure what was the reason for this ... will take a look again for v1.6.2 release.

PR #1709, opened 2020-10-22
migration - get localip

adevoss:
Please test on others setups. My ip address was blank.


rootzoll:
Replaced this PR with the solution to get the local IP from the specialized config-script `internet.sh` and added the `inet` to its scope to look for: https://github.com/rootzoll/raspiblitz/blob/2c9924eac140239302a757f0ea5ceca85d8c79cc/home.admin/config.scripts/blitz.migration.sh#L32

PR #1719, opened 2020-10-27
curl hangs if target is not available

ChuckNorrison:
Add some flags to curl:
```
-f, --fail          Fail silently (no output at all) on HTTP errors
-S, --show-error    Show error even when -s is used
-m, --max-time <seconds> Maximum time allowed for the transfer
```


rootzoll:
good improvement. thx.

rootzoll:
+cp into dev

Issue #1691, opened 2020-10-20
Provisioning of "DYNAMIC DOMAIN" gets stuck


rootzoll:
On last log line "# start to create new generate new TLSCert" recovery got stuck ... check if systematic or edge case.

rootzoll:
manual fix: ssh in (with password 'raspiblitz') and deactivate dynamic DNS. turn off. reflash sd card again and recover.

openoms:
Had another user reporting this.

Deactivate the Dynamic Domain with the script by running:
`config.scripts/internet.dyndomain.sh off`

https://t.me/RaspiBlitzFAQ/145

openoms:
@rootzoll shall we recommend to deactivate the dynDomain before updating to avoid getting stuck?

rootzoll:
@openoms Yepp .. until we have a v1.6.2 - targeting mid november.

geco91:
Will check what is going wrong on the logs when updating with dyndomain on.

Also maybe thats a good point to optimize the `internet.dyndomain.sh` config script with:
- have a `status` parameter
- fail on `on` when no domain is given
- make sure outputs are comment safe
- check if can be set thru settings menu correctly

geco91:
These are the logs where it blocks the update/provisioning:

```
Provisioning DYNAMIC DOMAIN - run config script
number of args(2)
making sure services are not running
# switching the DynamicDNS ON
# dynDomain(home.rotzoll.de)
# dynUpdateUrl()
# lnd.tlscert.sh domain-add
# OK the domain was already added lnd.conf
# DynamicDNS is now ON
# checking if LND is running
# making sure services are not running
# keep old tls data as backup
# start to create new generate new TLSCert
```

Because LND is not running the TLS cannot get created and it blocks. But during recovering there is no need to recreate the TLS - the macaroons and certs from backup will be used anyway after setup. Will check script again to make this work again.

geco91:
Can confirm that #1721 fixed the "getting stuck on update" problem on my test machine.

rootzoll:
Tested with RC2 - working fine with dynamic domain on. Closing issue for v1.6.2 release.

PR #1721, opened 2020-10-28
#1691 refactor internet.dyndomain.sh

geco91:
- should fix the "stuck on update" problem
- add a status that also checks if dydomain is uptodate
- moved the update process into this script (from background) for cleaner capsulating


rootzoll:
+cp into dev

Issue #1718, opened 2020-10-27
Lnd gets announced as 127.0.0.1:9735


ChuckNorrison:
### Bug
After LAN got disconnected and reconnected, 1ml.com shows up my node with a new address: 127.0.0.1:9735. This Address keeps active till reboot and is shown in my menu screen.

### Environment
- Clearnet (no IPSec or TOR)
- OpenVPN Service which connects as Client to VPS with static IP (port forwardings at the VPS ok)
- DynDNS with duckdns.org (blitz is recognizing the public static ip fine)
- `curl -s http://v4.ipv6-test.com/api/myip.php` from internet.sh works if my I-Net is up, hangs if it is down

### Expected Behaviour
Never write private or loopback IPs as a publicIP

ChuckNorrison:
I did a pull request which could be related to this issue (not sure if it will fix it)

https://github.com/rootzoll/raspiblitz/pull/1719

// I debugged this down to 00infoBlitz.sh:
```
 ln_getInfo=$(sudo -u bitcoin /usr/local/bin/lncli --macaroonpath=${lnd_macaroon_dir}/readonly.macaroon --tlscertpath=${lnd_dir}/tls.cert getinfo 2>/dev/null)
 ln_external=$(echo "${ln_getInfo}" | grep "uris" -A 1 | tr -d '\n' | cut -d '"' -f4)
```

The json gets his data (127.0.0.1) from the `lncli` call above

// Fixed with restart lnd service... So lnd service seems to cause this issue itself if started without connectivity?!

rootzoll:
Which IP is announced is set by parameter when LND starts - and is is getting started by sytstemd service:
https://github.com/rootzoll/raspiblitz/blob/eea158949899f816d59f9adca19b1d8972814036/home.admin/assets/lnd.service#L14

And the value of `${publicIP}` is set by the `raspiblitz.conf` with:
https://github.com/rootzoll/raspiblitz/blob/eea158949899f816d59f9adca19b1d8972814036/home.admin/assets/lnd.service#L12

So the question would be: Why is their a `127.0.0.1` as publicIP in `raspiblitz.conf`?

The value get written there by config-script `internet.sh update-publicip`:
https://github.com/rootzoll/raspiblitz/blob/eea158949899f816d59f9adca19b1d8972814036/home.admin/config.scripts/internet.sh#L210

And that is called during bootup (before LND is started by systemd) by `_bootstrap.sh`:
[...]

PR #1729, opened 2020-10-29
#1718 wait for public IP before LND start

geco91:
this should fix that LND is announcing a local IP to the network


ChuckNorrison:
Hey, i tested this code on my blitz successful:
- Reboot
- Cable Disconnect -> Screen with message to user gets displayed
- Cable connect -> common info screen appears, no localIP found so far

Well done 👍

PR #1748, opened 2020-11-07
Update node JS to 14.15.0

openoms:
Fixes a problem with `bos rebalance` in BoS 7.2.4

to update manually before the next release:
```
# download the script
wget https://raw.githubusercontent.com/openoms/raspiblitz/nodeJS-14.15.0/home.admin/config.scripts/bonus.nodejs.sh
# inspect the script
cat bonus.nodejs.sh
# uninstall current
bash bonus.nodejs.sh off
# install the update
bash bonus.nodejs.sh on
# done
```

### DIFF
--- a/home.admin/config.scripts/bonus.nodejs.sh
+++ b/home.admin/config.scripts/bonus.nodejs.sh
@@ -1,8 +1,10 @@
 #!/bin/bash
 
+VERSION="v14.15.0"
+
 # command info
 if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
- echo "small config script to install NodeJs"
+ echo "config script to install NodeJs $VERSION"
  echo "bonus.nodejs.sh [on|off]"
  exit 1
 fi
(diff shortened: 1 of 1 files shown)


grnqrtr:
I was having problems with bos and found this script. I'm not sure how I ended up with a higher version of node, but the script doesn't seem to be actually removing node. I tried off and on to get v14.15.0.

node

openoms:
> I was having problems with bos and found this script. I'm not sure how I ended up with a higher version of node, but the script doesn't seem to be actually removing node. I tried off and on to get v14.15.0.

You probably have nodeJS installed somewhere else.
Find it with:
`which node`
delete and reinstall.

grnqrtr:
Thank you. That was the problem. I don't remember doing that, but I must have at some point!

frennkie:
Well, this didn't really work for me: `bash bonus.nodejs.sh off` - but after removing `nodejs` with `apt-get purge` I was able to upgrade.

rootzoll:
Test ran good. Merging for v1.6.2 release.

PR #1751, opened 2020-11-09
Update BBopenChannel.sh

Hear7beaT:
Allowing the user to choose how fast they want to have their channel opened / how much they wish to pay in fees.

### DIFF
--- a/home.admin/BBopenChannel.sh
+++ b/home.admin/BBopenChannel.sh
@@ -104,14 +104,14 @@ if [ ${#amount} -eq 0 ]; then
   exit 1
 fi
 
-# let user enter a confirmation speed
+# let user enter a confirmation target
 l1=""
 l2="Urgent = 1 / Economy = 20"
 dialog --title "Open channel speed" \
 --inputbox "$l1\n$l2" 10 60 2>$_temp
-speed=$(cat $_temp | xargs | tr -dc '0-9')
+conf_target=$(cat $_temp | xargs | tr -dc '0-9')
 shred -u $_temp
-if [ ${#speed} -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


home.admin/BBopenChannel.sh
```diff
@@ -89,7 +89,7 @@ if [ $(echo "${error}" | grep "channel is too small" -c) -eq 1 ]; then
   minSat=$(echo "${error}" | tr -dc '0-9')
 fi
 
-# let user enter a amount
+# let user enter an amount
```
Hear7beaT, line 92: Correct UK grammer.

home.admin/BBopenChannel.sh
```diff
@@ -104,8 +104,22 @@ if [ ${#amount} -eq 0 ]; then
   exit 1
 fi
 
+# let user enter a confirmation speed
+l1=""
+l2="Urgent = 1 / Economy = 20"
+dialog --title "Open channel speed" \
```
openoms: good idea, but can we call it `confirmation target` rather than `speed`?
It would align to the explanation in the lnd manual:
``` 
# --conf_target value                      (optional) the number of blocks that the transaction *should* confirm in, will be used for fee estimation (default: 0)
```
https://api.lightning.community/#openchannel

Hear7beaT: Happy with this yeah!

Hear7beaT: Like this?

openoms: yes, but most importantly in the dialog!

rootzoll:
Test looks good. Will merge for v1.6.2 release.

Issue #1725, opened 2020-10-29
BTCpay: use postgres database by default 


openoms:
Postgres is better supported database format for for BTCPayServer as described in:  https://github.com/btcpayserver/btcpayserver/issues/1996

Will need to install postgres in the install script and have it running on BTCpay install to be used by default.

openoms:
Will try to make this work asap as restoring my sqllite database did not work no the second time. BTCpay cannot generate invoices any more despite all data was preserved.

logs when attempting to create an invoice: https://termbin.com/6tbc

```
Nov 10 21:04:59 raspberrypi dotnet[5703]: fail: Microsoft.AspNetCore.Server.Kestrel: Connection id "0HM45N8F199LP", Request id "0HM45N8F199LP:00000001": An unhandled exception was thrown by the application.
Nov 10 21:04:59 raspberrypi dotnet[5703]: Microsoft.EntityFrameworkCore.DbUpdateException: An error occurred while updating the entries. See the inner exception for details.
Nov 10 21:04:59 raspberrypi dotnet[5703]:  ---> Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 19: 'UNIQUE constraint failed: AddressInvoices.Address'.
Nov 10 21:04:59 raspberrypi dotnet[5703]:    at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
Nov 10 21:04:59 raspberrypi dotnet[5703]:    at Microsoft.Data.Sqlite.SqliteDataReader.NextResult()
Nov 10 21:04:59 raspberrypi dotnet[5703]:    at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
Nov 10 21:04:59 raspberrypi dotnet[5703]:    at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
Nov 10 21:04:59 raspberrypi dotnet[5703]:    at Microsoft.Data.Sqlite.SqliteCommand.ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
[...]

PR #1760, opened 2020-11-11
btcpay: update to v1.0.5.9, NBXplorer v2.1.46

openoms:
This is just a maintenance update with the new versions and improving the comments.

Would be nice to have postgres as the default database if someone can have look at it:
https://github.com/rootzoll/raspiblitz/issues/1725


rootzoll:
Test look good. Merging for v1.6.2 release.

Issue #1752, opened 2020-11-10
Tor repo doesn't support architecture 'armhf' (32bit ARM) - need to compile from source?


openoms:
This is a possible issue for all newly set up Tor nodes:
```
$ sudo apt update
                                     
Hit:5 https://deb.torproject.org/torproject.org buster InRelease                   

N: Skipping acquire of configured file 'main/binary-armhf/Packages' as repository 'https://deb.torproject.org/torproject.org buster InRelease' doesn't support architecture 'armhf'
```

Reported in RaspiBolt: https://github.com/Stadicus/RaspiBolt/issues/656

If this remains to be an issue there are two obvious solutions:
* switch to a 64bit base image (https://github.com/rootzoll/raspiblitz/issues/1199 - aimed for v1.7)
* compile Tor from source (can help to get around some issues with some ISP censorship since can be downloaded from github)

Methods to compile Tor from source:

1) 
* https://t.me/RaspiBlitzFAQ/168
* https://tor.stackexchange.com/questions/75/how-can-i-install-tor-from-the-source-code-in-the-git-repository
    ```
    sudo apt-get install -y git build-essential automake libevent-dev libssl-dev zlib1g-dev

    # git clone https://git.torproject.org/tor.git
    git clone https://github.com/torproject/tor

    cd tor
    ./autogen.sh
 
    ./configure --disable-asciidoc 

    make
    make install
    ```

2) 
* https://github.com/openoms/joininbox/blob/4042dde809eb9cba5a3e1c2941dc414eb93251b6/build_joininbox.sh#L343
* https://2019.www.torproject.org/docs/debian#source

   ```
    # https://2019.www.torproject.org/docs/debian#source
    echo "# running on armv6l - need to compile Tor from source"
    apt install -y build-essential fakeroot devscripts
    apt build-dep -y tor deb.torproject.org-keyring
    mkdir ~/debian-packages; cd ~/debian-packages
    apt source tor
    cd tor-*
    debuild -rfakeroot -uc -us
    cd ..
    dpkg -i tor_*.deb
    # setup Tor in the backgound
    # TODO - test if remains in the background after the Tor service is started
    tor &
    ```

rootzoll:
I just tested a recover/update with the v1.6.1 image (where Tor gets installed) and it was working - so I pushed this to v1.7 where we anyway consider to test the 64-bit image. @openoms do you see anything pressing for the v1.6.2 release?

frennkie:
[...]

PR #1769, opened 2020-11-13
Tor from source

openoms:
as suggested in: https://github.com/rootzoll/raspiblitz/issues/1752#issuecomment-726746955

During the build process the node is running normally.
Tor is only restarted to change to the updated .deb package.


Dikay900:
worked for me after reinitializing an existing installation with image 1.6.1

PR #1795, opened 2020-11-22
fix typos: replace s/ngnix/nginx/g

DivineDominion:
I grepped the repo, these were the only occurrences.


rootzoll:
Thanks - good catch :)

PR #1809, opened 2020-11-27
sphinxrelay: fix permission when editing config.json

openoms:


rootzoll:
I think we worked on the same thing the same time - this should be fixed already in v1.6 branch - please check.

openoms:
> this should be fixed already in v1.6 branch - please check.

Still there at: https://github.com/rootzoll/raspiblitz/blob/bca71a95a5268dbd2a19383230a30cf3dd4c9212/home.admin/config.scripts/bonus.sphinxrelay.sh#L129
admin can't write to: `/home/sphinxrelay/sphinx-relay/config/app.json.tmp`

It is similar, but not the same as https://github.com/rootzoll/raspiblitz/pull/1789

rootzoll:
this should ne OK because `/home/admin/config.scripts/bonus.sphinxrelay.sh write-environment` will always be excuted as "sphinxrelay" user. So we can close this PR without merging?

rootzoll:
To be more precise ... this PR would not work because `write-environment` is excuted by the systemd as a pre-start to running the sphinx relay as user `sphinxrelay` to it would be allowed to write to the admin users home. The code should be fine as it is now ... will test another round. Planning to close this PR without merge.

openoms:
Yes, good points. Tested another time with the latest version of the script and there were no errors. Closing this.

Issue #1758, opened 2020-11-11
[Feature request] SphinX relay


openoms:
https://sphinx.chat/

Manual deployment notes (might be outdated):
https://github.com/stakwork/sphinx-relay/blob/master/docs/raspiblitz_deployment.md

rootzoll:
Lets give it a shot if we can include this into v1.6.2

rootzoll:
OK this is the `app.json` 
```
admin@raspberrypi:/home/sphinxrelay/sphinx-relay $ sudo cat ./config/app.json
{
  "development": {
    "senza_url": "http://localhost:3000/api/v2",
    "macaroon_location": "/Users/evanfeenstra/code/lnd-dev/alice/data/chain/bitcoin/simnet/admin.macaroon",
    "tls_location": "/Users/evanfeenstra/Library/Application Support/Lnd/tls.cert",
    "node_ip": "127.0.0.1",
    "lnd_port": "10001",
    "node_http_protocol": "http",
    "node_http_port": "3002",
    "hub_api_url": "http://lvh.me/api/v1",
    "hub_url": "http://lvh.me/ping",
    "hub_invite_url": "http://lvh.me/invites",
    "hub_check_invite_url": "http://lvh.me/check_invite",
    "media_host": "localhost:5000",
    "tribes_host": "tribes.sphinx.chat"
  },
  "production": {
    "senza_url": "https://staging.senza.us/api/v2/",
    "macaroon_location": "/mnt/hdd/app-data/lnd/data/chain/bitcoin/mainnet/admin.macaroon",
    "tls_location": "/mnt/hdd/app-data/lnd/tls.cert",
    "lnd_log_location": "/mnt/hdd/lnd/logs/bitcoin/mainnet/lnd.log",
    "node_ip": "localhost",
    "node_http_protocol": "http",
    "node_http_port": "3300",
    "lnd_port": "10009",
    "hub_api_url": "https://hub.sphinx.chat/api/v1",
    "hub_url": "https://hub.sphinx.chat/ping",
    "hub_invite_url": "https://hub.sphinx.chat/invites",
    "hub_check_invite_url": "https://hub.sphinx.chat/check_invite",
    "media_host": "memes.sphinx.chat",
    "tribes_host": "tribes.sphinx.chat",
    "public_url": "http://192.168.178.95:3300"
  }
}
```

[...]

PR #1810, opened 2020-11-27
sphinxrelay: fix update option

openoms:
as in: https://github.com/rootzoll/raspiblitz/issues/1758#issuecomment-734513624

### DIFF
--- a/home.admin/config.scripts/bonus.sphinxrelay.sh
+++ b/home.admin/config.scripts/bonus.sphinxrelay.sh
@@ -4,7 +4,7 @@
 
 # command info
 if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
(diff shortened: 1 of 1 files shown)


rootzoll:
I think we worked on the same thing the same time - this should be fixed already in v1.6 branch - please check.

rootzoll:
@openoms ah OK the config overwrite could be a think still on my impelementation. Please rebase again against my latest merge.

openoms:
> I think we worked on the same thing the same time - this should be fixed already in v1.6 branch - please check.

yes I've seen and I have also started with the same update method. 
Tested this to update from v1.1.0 to v1.1.1 and worked ok. 
It is based on the latest commit.

rootzoll:
Ah thats clever to use the git stash to preserve config .. will merge much thx.

Issue #1842, opened 2020-12-04
Update to v1.6.2 wiped out Specter and electrs fails to start


shawnyeager:
So, my Specter config is gone – devices and wallets.

And electrs isn't starting with the following:

```
Dec 04 22:12:23 raspberrypi systemd[1]: Started Electrs.
Dec 04 22:12:23 raspberrypi systemd[7819]: electrs.service: Failed to execute command: No such file or directory
Dec 04 22:12:23 raspberrypi systemd[7819]: electrs.service: Failed at step EXEC spawning /home/electrs/electrs/target/release/electrs: No such file or directory
Dec 04 22:12:23 raspberrypi systemd[1]: electrs.service: Main process exited, code=exited, status=203/EXEC
Dec 04 22:12:23 raspberrypi systemd[1]: electrs.service: Failed with result 'exit-code'.
Dec 04 22:13:23 raspberrypi systemd[1]: electrs.service: Service RestartSec=1min expired, scheduling restart.
Dec 04 22:13:23 raspberrypi systemd[1]: electrs.service: Scheduled restart job, restart counter is at 1.
Dec 04 22:13:23 raspberrypi systemd[1]: Stopped Electrs.
Dec 04 22:13:23 raspberrypi systemd[1]: Started Electrs.
Dec 04 22:13:23 raspberrypi systemd[1]: electrs.service: Main process exited, code=exited, status=203/EXEC
Dec 04 22:13:23 raspberrypi systemd[1]: electrs.service: Failed with result 'exit-code'.
Dec 04 22:14:23 raspberrypi systemd[1]: electrs.service: Service RestartSec=1min expired, scheduling restart.
Dec 04 22:14:23 raspberrypi systemd[1]: electrs.service: Scheduled restart job, restart counter is at 2.
Dec 04 22:14:23 raspberrypi systemd[1]: Stopped Electrs.
Dec 04 22:14:23 raspberrypi systemd[1]: Started Electrs.
Dec 04 22:14:23 raspberrypi systemd[13427]: electrs.service: Failed to execute command: No such file or directory
Dec 04 22:14:23 raspberrypi systemd[13427]: electrs.service: Failed at step EXEC spawning /home/electrs/electrs/target/release/electrs: No such file or directory
Dec 04 22:14:23 raspberrypi systemd[1]: electrs.service: Main process exited, code=exited, status=203/EXEC
[...]

PR #1847, opened 2020-12-06
specter: move to dedicated user and store data on the disk

openoms:
Fixing: https://github.com/rootzoll/raspiblitz/issues/1842

* create the `specter` user
* store data on the disk
* update to v0.10.2

### DIFF
--- a/home.admin/config.scripts/bonus.cryptoadvance-specter.sh
+++ b/home.admin/config.scripts/bonus.cryptoadvance-specter.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # https://github.com/cryptoadvance/specter-desktop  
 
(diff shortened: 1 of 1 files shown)


rootzoll:
@openoms do you see this as something that can wait a bit until v1.7 (maybe spring 2021) is ready or should we target a v1.6.3 release for end of this year?

openoms:
The only big difference here that with this the Specter data can be carried over during an update or restore. 

It depends how one uses Specter, but should utilise the built-in backup option of Specter if storing valuable data.  Now even hot wallets are possible to be done with Specter so can be an issue.

rootzoll:
So if this is the only slightly critical fix I think we can delay it to v1.7 ... but if we make a 1.6.3 this should be part of it.

PR #1848, opened 2020-12-06
Validate TCP port number

roshii:
Making sure the port is within TCP port range


rootzoll:
Good input check - thx :) Will include in v1.7 release.

PR #1879, opened 2020-12-17
Update CHANGES.md

we-do-it-lu:
Added reference to BTCPay Server 1.0.5.9


rootzoll:
Thanks, merged - good addition :)

Issue #1906, opened 2020-12-25
Test 1.6.3 Release Candidate


rootzoll:
Download Release Candidate on your test/development node:

https://raspiblitz.org/raspiblitz-v1.6.3rc2-2020-12-24.img.gz
SHA: c11816fd8ffda991991bf687ce1fd70ceeb6c61fb960e633b1bd5a885501d0e0

Please give feedback. Thx.

Cranky4u:
Having a crack to see if resolves i2c hanging issue on boot for RPi 3b+

Cranky4u:
> 
> 
> Having a crack to see if resolves i2c hanging issue on boot for RPi 3b+

1st attempt with 1.6.3 unsuccessful - RPi will not boot
* with or without HDMI
* imaged with balenaEtcher
* system hangs on red boot light

Now making image with RUFUS and testing boot up.

Cranky4u:
> 
> 
> > Having a crack to see if resolves i2c hanging issue on boot for RPi 3b+
> 
> 1st attempt with 1.6.3 unsuccessful - RPi will not boot
> 
>     * with or without HDMI
> 
>     * imaged with balenaEtcher
> 
>     * system hangs on red boot light
> 
> 
> Now making image with RUFUS and testing boot up.
RUFUS built SD card fails too :(

My RPi 3b+ still not running with 1.6.3 or 1.6.2 or 1.6 or 1.5

Not sure what is wrong....Pi does work though as use it for retropi gaming

Dikay900:
working without a problem, things tested after update from v1.6.2 on rpi 4

- headless (no monitor)
- bitcoin + lnd all over tor
- joinmarket (needed to reconfigure maker configuration / fees)
- rtl
- lndmanage
- pool

Cranky4u:
Looks like 1.6.3 is not working on my RPi 3 & Seagate 1TB drive. Hangs on HDD detection and has been left to run overnight for ~12 hours. With HDD removed, system hangs on i2c driver stage...
Maybe the issue is with compatibility with new Seagate drive?
@rootzoll Should the Seagate 1TB drive be compatible out of the box or should it be formatted first (use Win7 PC to set it up or does raspiblitz handle reformatting / set up)?

Cranky4u:
> 
> 
> working without a problem, things tested after update from v1.6.2 on rpi 4
> 
>     * headless (no monitor)
> 
>     * bitcoin + lnd all over tor
> 
[...]

PR #1919, opened 2021-01-06
tor: fix keys and repos

openoms:
Discussed in:
https://github.com/rootzoll/raspiblitz/issues/1906#issuecomment-755299759


ghost:
As I mentioned in https://github.com/rootzoll/raspiblitz/issues/1906#issuecomment-757078249 I'm 99% sure you shouldn't do it (and maybe apt won't even let you)

PR #1926, opened 2021-01-10
Update 00mainMenu.sh

NilByte:
Also listing bitcoind command line help is appropriate.


rootzoll:
Its a good idea to mention also the bitcoin command line options. I changed to `bitcoin-cli` because its the more practical option to communicate with the running bitcoind server.

NilByte:
Shouldn't it be `bitcoin-cli -h` then?

rootzoll:
`bitcoin-cli help` gives the list of all commands - I think thats the best place to point people to to get started

Issue #1901, opened 2020-12-24
Bitcoind: Can't open PID from file


jodobear:
Fresh 1.6.2 install with LND recovered from backup. Bitcoind is `active` but it isn't syncing, get `Failed to parse PID from file` and after service restart get `Can't open PID from file` as shown in the below `status` output:
```sh
● bitcoind.service - Bitcoin daemon
   Loaded: loaded (/etc/systemd/system/bitcoind.service; enabled; vendor preset: enabled)
   Active: active (running) since Wed 2020-12-23 12:20:38 GMT; 1min 32s ago
  Process: 10999 ExecStartPre=/home/admin/config.scripts/blitz.systemd.sh log blockchain STARTED (code=exited, status=0/
  Process: 11001 ExecStart=/usr/local/bin/bitcoind -daemon -conf=/home/bitcoin/.bitcoin/bitcoin.conf -pid=/home/bitcoin/
 Main PID: 11002 (bitcoind)
    Tasks: 18 (limit: 4915)
   Memory: 199.2M
   CGroup: /system.slice/bitcoind.service
           └─11002 /usr/local/bin/bitcoind -daemon -conf=/home/bitcoin/.bitcoin/bitcoin.conf -pid=/home/bitcoin/.bitcoin

Dec 23 12:20:38 raspberrypi systemd[1]: Starting Bitcoin daemon...
Dec 23 12:20:38 raspberrypi systemd[1]: bitcoind.service: Can't open PID file /home/bitcoin/.bitcoin/bitcoind.pid (yet?)
Dec 23 12:20:38 raspberrypi systemd[1]: Started Bitcoin daemon.
```

```sh
$ ls -l /home/bitcoin/.bitcoin/
total 324
-rw------- 1 bitcoin bitcoin     37 Dec 22 20:22 banlist.dat
-r--r--r-- 1 root    root       854 Dec 22 23:56 bitcoin.conf
-rw------- 1 bitcoin bitcoin      5 Dec 24 08:43 bitcoind.pid
drwx------ 1 bitcoin bitcoin 114010 Dec 21 16:33 blocks
drwx------ 1 bitcoin bitcoin  48572 Dec 24 08:43 chainstate
-rw------- 1 bitcoin bitcoin   6877 Dec 24 08:49 debug.log
-rw------- 1 bitcoin bitcoin 247985 Dec 23 15:30 fee_estimates.dat
[...]

PR #1930, opened 2021-01-10
#1901 remove PID file from service

rootzoll:
getting rid of the PID file bitcoind is creating to prevent problems like #1901

### DIFF
--- a/home.admin/assets/bitcoind.service
+++ b/home.admin/assets/bitcoind.service
@@ -14,7 +14,7 @@ Group=bitcoin
 Type=forking
 PIDFile=/home/bitcoin/.bitcoin/bitcoind.pid
 ExecStartPre=-/home/admin/config.scripts/blitz.systemd.sh log blockchain STARTED
-ExecStart=/usr/local/bin/bitcoind -daemon -conf=/home/bitcoin/.bitcoin/bitcoin.conf -pid=/home/bitcoin/.bitcoin/bitcoind.pid
+ExecStart=/usr/local/bin/bitcoind -daemon -conf=/home/bitcoin/.bitcoin/bitcoin.conf
 KillMode=process
 Restart=always
 TimeoutSec=120


openoms:
It seems the PID file is needed to be able to spawn more instances of the process: https://stackoverflow.com/questions/47826123/what-is-systemd-pid-file

Looking at htop I have multiple bitcoimd processes running unsurpisingly since many services are relying on it.

It might be a bad idea to remove that PID file, but will need to try to confirm.

rootzoll:
OK pushing this to next release until we can confirm.

openoms:
I think we would need to remove the `-daemon` entry from the `bitcoin.service` to get rid of the PID file. Its shoul not be needed if running with systemd.

From `bitcoind -help`:
```
  -daemon
       Run in the background as a daemon and accept commands

  -pid=<file>
       Specify pid file. Relative paths will be prefixed by a net-specific
       datadir location. (default: bitcoind.pid)
```

Extract from a chat on #cl-lightning cdecker:
```
Never tried setting `--daemon` in a config, setting it from the command line should work though
I see I try to put it to the service, but if all it does is suppress the output it might be not needed
it does not seem to like the daemon setting with multiple processes, complaining about left-overs and unable to lock the pidfile
is there any problem with running with running without daemon=true?
Nope, `--daemon` just detaches from the terminal and runs in the background, so if you're using systemd, init, supervisor, or any other service/process manager it won't change anything (well except a bit more work forking the process in `--daemon` mode)
thank you for the explanation. I was only curious because @darosior mentioned that it would deafult to create a logfile in daemon mode, but I rather just use log-file= option in the config.
```

openoms:
Tested and the `bictoind.service` does not start without the `-daemon` option.

The PID file (/home/bitcoin/.bitcoin/bitcoind.pid) is recreated automaically on restart if the `-dameon` option is present so the PID file path does not need to be defined.
 The line 
 `PIDFile=/home/bitcoin/.bitcoin/bitcoind.pid`

can be removed also.


If the process type is set to  
`Type=simple`
instead of `Type=forking`
the `-daemon` option can be removed.

Currently testing ok with the settings:

```
[Unit]
Description=Bitcoin daemon on signet

[Service]
User=joinmarket
Group=joinmarket
Type=simple
ExecStart=/home/joinmarket/bitcoin/bitcoind -signet
KillMode=process
Restart=always
TimeoutSec=120
RestartSec=30
StandardOutput=null
StandardError=journal

[Install]
WantedBy=multi-user.target
```

Will apply the changes to a mainnet node as well.

openoms:
On mainnet the change is not going well:
```
# RaspiBlitz: systemd unit for bitcoind

[Unit]
Description=Bitcoin daemon
Wants=bootstrap.service
After=bootstrap.service

# for use with sendmail alert (coming soon)
#OnFailure=systemd-sendmail@%n

[Service]
User=bitcoin
Group=bitcoin
Type=simple
#PIDFile=/home/bitcoin/.bitcoin/bitcoind.pid
ExecStartPre=-/home/admin/config.scripts/blitz.systemd.sh log blockchain STARTED
ExecStart=/usr/local/bin/bitcoind -conf=/home/bitcoin/.bitcoin/bitcoin.conf
# -daemon -pid=/home/bitcoin/.bitcoin/bitcoind.pid
KillMode=process
Restart=always
TimeoutSec=120
RestartSec=30
StandardOutput=null
StandardError=journal

[Install]
WantedBy=multi-user.target
```
```
Mar 09 11:29:12 raspberrypi systemd[1]: bitcoind.service: Found left-over process 2808 (bitcoind) in control group while starting unit. Ignoring.
Mar 09 11:29:12 raspberrypi systemd[1]: This usually indicates unclean termination of a previous run, or service implementation deficiencies.
Mar 09 11:29:12 raspberrypi systemd[1]: Starting Bitcoin daemon...
Mar 09 11:29:12 raspberrypi systemd[1]: bitcoind.service: Found left-over process 2808 (bitcoind) in control group while starting unit. Ignoring.
Mar 09 11:29:12 raspberrypi systemd[1]: This usually indicates unclean termination of a previous run, or service implementation deficiencies.
Mar 09 11:29:12 raspberrypi systemd[1]: Started Bitcoin daemon.
Mar 09 11:29:12 raspberrypi bitcoind[3368]: Error: Cannot obtain a lock on data directory /mnt/hdd/bitcoin. Bitcoin Core is probably already running.
Mar 09 11:29:12 raspberrypi systemd[1]: bitcoind.service: Main process exited, code=exited, status=1/FAILURE
Mar 09 11:29:12 raspberrypi systemd[1]: bitcoind.service: Failed with result 'exit-code'.
```

openoms:
When applying only the change in this PR it works:
```
# RaspiBlitz: systemd unit for bitcoind

[Unit]
Description=Bitcoin daemon
Wants=bootstrap.service
After=bootstrap.service

# for use with sendmail alert (coming soon)
#OnFailure=systemd-sendmail@%n

[Service]
User=bitcoin
Group=bitcoin
Type=forking
PIDFile=/home/bitcoin/.bitcoin/bitcoind.pid
ExecStartPre=-/home/admin/config.scripts/blitz.systemd.sh log blockchain STARTED
ExecStart=/usr/local/bin/bitcoind -conf=/home/bitcoin/.bitcoin/bitcoin.conf -daemon
#-pid=/home/bitcoin/.bitcoin/bitcoind.pid
KillMode=process
Restart=always
TimeoutSec=120
RestartSec=30
StandardOutput=null
StandardError=journal

[Install]
WantedBy=multi-user.target
```

The line `PIDFile=/home/bitcoin/.bitcoin/bitcoind.pid` still defines the PID file which was likely overridden before.

It is ok to merge this, but not expecting any change in the behaviour.

rootzoll:
Yeah you seem to be right ... this might not change anything in behaviour - so why changing and risking side effects. Closing without merging. Thanks for investigating.

PR #1943, opened 2021-01-13
lnd.conf: add tls settings, improve startup time

openoms:
Adding tls settings as discussed in: https://github.com/rootzoll/raspiblitz/issues/1117

+ settings to speed up LND startup and avoid unnecessary gossip from
https://github.com/alexbosworth/run-lnd#install-lnd

### DIFF
--- a/home.admin/assets/lnd.bitcoin.conf
+++ b/home.admin/assets/lnd.bitcoin.conf
@@ -7,6 +7,18 @@ alias=raspiblitz
 color=#68F442
 nat=false
 
+# Avoid historical graph data sync
+ignore-historical-gossip-filters=1
+# Avoid slow startup time
+sync-freelist=1
+# Avoid high startup overhead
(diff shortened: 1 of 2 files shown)


rootzoll:
@openoms  can you rebase this PR on the `dev` branch?

openoms:
ok, rebased to dev

rootzoll:
OK changes were merged into `dev` for v1.7RC1 release. @openoms the new settings would only effect new fresh setup nodes right?

Do we also want fo the final v1.7 to edit these values on update into existing lnd.conf ... or would that renew existing tls certs and break existing connections to wallets etc? If so maybe at least force those parameters in on TLS renew option?

openoms:
Good point, it is difficult to predict when it would renew on update although the TLS cert is only valid for a year by default, so older nodes can run into the problem anyway.

Should be included with the tls renew, but agree that probably better to leave the old configs alone.

PR #1831, opened 2020-12-01
txindex: only restart bitcoind when running

openoms:
fixes: https://github.com/rootzoll/raspiblitz/issues/1830

An edge case only if the recovery process starts with a new blockchain database without txindex, but BTC-RPC-Explorer enabled.

PR #1945, opened 2021-01-13
txindex: restart bitcoind only if already active

openoms:
activated BTCRPCExplorer only in the raspiblitz.conf without switching txindex on and it got stuck on recovery similar to: https://github.com/rootzoll/raspiblitz/pull/1722

This PR only fixing this edge case.

txindex should be already on if BTC-RPC-Explorer was installed previously.

Related: #1906 

If someone would be stuck during recovery after `txindex` is activated run:
`sudo systemctl stop bitcoind`
and the recovery process will continue.


openoms:
included in https://github.com/rootzoll/raspiblitz/pull/1831/files so closing.

PR #1947, opened 2021-01-14
Organizing ./XXdebugLogs.sh debug logs - Change Wifi to Network and more

nyxnor:
Network info because it show what type of network (eth or wlan).
Mounted drives to human readable, MB, GB, KB, maintaining type of format(ext4,ntfs).

As for the other changes, they are just to maintain the same order as BTC and LND logs (Status, Error, Info).

### DIFF
--- a/home.admin/XXdebugLogs.sh
+++ b/home.admin/XXdebugLogs.sh
@@ -176,7 +176,6 @@ echo "*** MOUNTED DRIVES ***"
 df -T -h
 echo ""
 
-echo ""
 echo "*** NETWORK Info ***"
 sudo /home/admin/config.scripts/internet.sh status | grep 'network_device\|localip\|dhcp'
 echo ""
--- a/home.admin/XXdebugLogs.sh
+++ b/home.admin/XXdebugLogs.sh
@@ -78,37 +78,39 @@ echo "--> CHECK CONFIG: sudo nginx -t"
 sudo nginx -t
 echo ""
 
-if [ "${touchscreen}" = "1" ]; then
+if [ "${touchscreen}" = "0" ]; then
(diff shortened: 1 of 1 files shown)


nyxnor:
Correcting, enhancing.
Please see last info only, my mention, on forked repo.
Will need to see all the document, unfortunately its just showing deleting extra space, but there is more once expands de code.

rootzoll:
@Ben10Brasileiro I think changing the if clauses to test for 'off' instead for 'on' has the problem that in the case  a user has never activated a service there is no "off" value in the config for that variable - the variable is empty instead ... so even the service was never activated the else-clause will get activated and as if the service was set to on.

nyxnor:
@rootzoll  Yes, unfortunately I noticed that after, but corrected. Sorry for displaying this pull request messy, but here it is.
This was my first pull, so I didnt knew that deleting the branch in my directory would delete here also.
Changes(correction to previous changes): https://github.com/Ben10Brasileiro/raspiblitz/commit/2d0614caab1bfab6e92f6880583f201ca34fd6d0
Total file: https://github.com/Ben10Brasileiro/raspiblitz/blob/v1.6/home.admin/XXdebugLogs.sh

For a better picture, here it is the debug file: 
[debugtest.txt](https://github.com/rootzoll/raspiblitz/files/5864158/debugtest.txt)

Termbin didnt generate the entire file, got stuck at half of electRs status
https://termbin.com/xf33

rootzoll:
np - its hard to get PR right for a running project. Maybe just close this one (because it still contains the changes if-clauses) and make a fresh one?

nyxnor:
Ive referenced the code in the issue https://github.com/rootzoll/raspiblitz/issues/1986
Closed pull.

PR #1948, opened 2021-01-14
Adding number of bitcoin peers to the debug

nyxnor:
Just to add number of bitcoin peers to the debug log ./XXdebugLogs.sh via ./blitzstatusscan.sh.

### DIFF
--- a/home.admin/config.scripts/blitz.statusscan.sh
+++ b/home.admin/config.scripts/blitz.statusscan.sh
@@ -69,6 +69,22 @@ if [ ${bitcoinRunning} -eq 1 ]; then
     initialSync=$(echo ${blockchaininfo} | jq -r '.initialblockdownload' | grep -c 'true')
     echo "initialSync=${initialSync}"
 
+    # get number of bitcoin connections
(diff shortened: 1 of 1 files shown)


rootzoll:
v1.6.3 was adding a script to monitor peer connections:
https://github.com/rootzoll/raspiblitz/blob/v1.6/home.admin/config.scripts/network.monitor.sh

Do you think its important to add this info again to the `blitzstatusscan.sh`?

nyxnor:
Well, I made this before 1.6.3, but posted close to release date. 
Why: because there were a lot of people that didnt understand why it wasnt syncing and the debug wasnt presenting this information. But now that I havent seen any complaints, maybe just for information. If there is the btc peers count, why not the lnd? 
But to be fair, the status screen now has the lnd peers, only of couse if its past the IBD.

rootzoll:
The v1.6.3 is showing the lnd peers now also during IBD - see:
https://github.com/rootzoll/raspiblitz/blob/82e0d6c3714ce1b2878780c4bdef72a6417f71c7/home.admin/80scanLND.sh#L257

nyxnor:
Yes, now it shows, when I made it, it wasnt implemented yet on the blitz screen of IBD. Very nice for 1.6.3 having this. Closing now.

Issue #1993, opened 2021-01-30
[New] Lightning Terminal


openoms:
https://github.com/lightninglabs/lightning-terminal will replace the separate Loop, Pool and Faraday installations.

Find out more about usage: https://github.com/lightninglabs/lightning-terminal/blob/master/doc/WALKTHROUGH.md

To test:
```
# remove the previous version of the script:
rm -f bonus.lit.sh
# download:
wget https://raw.githubusercontent.com/openoms/raspiblitz/new-lit/home.admin/config.scripts/bonus.lit.sh
# check:
cat bonus.lit.sh
# install:
bash bonus.lit.sh on
# start service:
sudo systemctl start litd
# info:
bash bonus.lit.sh menu
```

~~If you have used Pool, Loop and Faraday before now there will be duplicate instances under separate users. The Lightning Terminal won't see your old Loop and Pool transactions and accounts.~~
UPDATE with https://github.com/rootzoll/raspiblitz/pull/1994/commits/910dcf5466b119514c99de3e0f58fc7da9f2220b the data directories of Loop and Pool are symlinked to LiT and previous data will appear.
Reinstall with the updated script to take effect if missing Pool data especially.

Omicron59:
Hello openoms  I tested the new terminal last night, really great working. I rebooted my raspiblitz and since it is blocked on lightning progress and I have an error, will you have an idea, or it is normal that it is so long. thanks to you

![image](https://user-images.githubusercontent.com/46853496/106382065-460a9700-63bd-11eb-9d09-58ad2e671511.png)

openoms:
@mamar59 your issue i not Lightning Terminal related. Seems the same as: https://github.com/rootzoll/raspiblitz/issues/1982, please discuss there. 
Looking at the output of the logs with `debug` can help a lot with details.

nyxnor:
UI password is presented on the logs. just a warn to anyone who post the output to this `sudo systemctl status litd`.
Openoms, I needed to restart (useful to be in the instructions), because stopping tor stopped all other services, so starting one by one was not optimal.
@mamar59 Have you tried the UI? Try the lightning pool screen to see if you get the same thing.

Ligthing pool segment, it´s saying the token was incorrect and I needed to `listauth` to check.
![image](https://user-images.githubusercontent.com/69700936/106388367-ef9d5880-63bc-11eb-9a80-d1812e51639a.png)

using `sudo systemctl status litd`
[...]

PR #1994, opened 2021-01-30
add the Lightning Terminal 0.4.1-alpha

openoms:
As discussed in: https://github.com/rootzoll/raspiblitz/issues/1993

### DIFF
--- a/home.admin/_commands.sh
+++ b/home.admin/_commands.sh
@@ -216,6 +216,21 @@ function pool() {
   fi
 }
 
+# aliases for lit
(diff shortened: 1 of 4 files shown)


openoms:
Updated RTL as in: https://github.com/rootzoll/raspiblitz/issues/2019

refactored the RTL config, now tested and working with LiT, should work with the CLI Loop install too, but I suggest we can leave that to be CLI only

rootzoll:
OK test looks good - merging :)

PR #2016, opened 2021-02-21
Update blitz.notify.sh

frennkie:
Make sure to run apt-get update unless it has been already been run recently (within last 7 days).

I got the one-liner from here: https://askubuntu.com/questions/410247/how-to-know-last-time-apt-get-update-was-executed/904259#904259

The problem if `apt-get update` is not run "recently" before apt-get install is that the system could try to use an outdated apt/repo mirror (which happend to me just now).


rootzoll:
@frennkie can you rebase the PR on the `dev` branch?

rootzoll:
Merged into `dev` for v1.7RC1 by cherrypicking.

PR #2034, opened 2021-03-02
Remove typo in echo

keblek:
partion -> partition


rootzoll:
Good catch - thanks :)

PR #2040, opened 2021-03-07
Fixed a typo

Nerfzooka:
Changed waist of time to waste of time in line 528


rootzoll:
Thanks for finding that one :)

Issue #2066, opened 2021-03-16
Test & make sure updated apps run correctly


rootzoll:
merged #2043 .. but now make sure that all those apps run correctly.

BTCpayServer has still error:
`Mar 16 23:14:36 raspberrypi dotnet[12088]: Unhandled exception. System.InvalidOperationException: No database option was configured.` see https://github.com/rootzoll/raspiblitz/pull/2043#issuecomment-800674779

openoms:
BTCpay is fixed in: https://github.com/rootzoll/raspiblitz/pull/2067

rootzoll:
Worked in Test:
- [X] BTCPayServer
- [X] Thunderhub
- [X] BOS
- [X] Specter
- [X] joininbox & joinmarket

rootzoll:
OK my first tests look good - keeping this issue open for final testing before release.

rootzoll:
OK .. apps looking good. Closing issue for v1.7 release.

PR #2043, opened 2021-03-09
Tested app updates

openoms:
79e9b717 bos: update to v0.8.0
0fc3f828 btcpay: update to v2.1.49, nbxplorer to v1.0.6.8

PR #2072, opened 2021-03-17
Small corrections for upadted Apps #2066

rootzoll:
fixing smaller things while testing the updated apps from merge #2043

### DIFF
--- a/home.admin/_commands.sh
+++ b/home.admin/_commands.sh
@@ -142,6 +142,8 @@ function fwdreport() {
 function bos() {
   if [ $(grep -c "bos=on" < /mnt/hdd/raspiblitz.conf) -eq 1 ]; then
     echo "# switching to the bos user with the command: 'sudo su - bos'"
+    echo "# use command 'exit' and then 'raspiblitz' to return to menu"
+    echo "# use command 'bos --help' to list all possible options"
     sudo su - bos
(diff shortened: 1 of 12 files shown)


home.admin/_commands.sh
```diff
@@ -154,7 +157,10 @@ function bos() {
 function pyblock() {
   if [ $(grep -c "pyblock=on" < /mnt/hdd/raspiblitz.conf) -eq 1 ]; then
     echo "# switching to the pyblock user with the command: 'sudo su - pyblock'"
+    echo "# use command 'exit' and then 'raspiblitz' to return to menu"
+    echo "# use command 'pyblock' again to start"
     sudo su - pyblock
+    echo "# use command 'raspiblitz' to return to menu"
```
openoms, line 163: just wondering is it not `exit` the correct command here, since `raspiblitz` won't change back to the admin user?

rootzoll: this last echo is just displayed if the user typed already `exit` to leave the pyblock user mode (as adviced to use in the echos above) then the user is back at the admin user (see example image below) and that echo line is a reminder how to get back to the MENU.
Bildschirmfoto 2021-03-21 um 21 19 42

openoms: makes sense, thanks!

Issue #2090, opened 2021-03-22
CONNECT menu to collect how to connect wallets and services


openoms:
Would help with answering the questions of how I connect XX app or wallet to the node.

By default should show connection to bitcoind and LND:

- [x] Bitcoind RPC for Specter Desktop and a remote Joininbox / JoinMarket
https://d11n.net/connect-specter-desktor-with-raspiblitz.html
https://github.com/openoms/joininbox/blob/master/prepare_remote_node.md
- [x] Fully Noded (from the MOBILE menu)
- [x] Bitcoind P2P - Bisq connects to v2 Tor service
https://twitter.com/openoms/status/1355885287881240578 
- [x] LND rest and GRPC with the text with
  - [x] TLS cert (from LNDCREDS)
  - [x] macaroons (from LNDCREDS)
  - [ ] Tor Hidden Services
  - [x] LNDconnect as now in the MOBILE menu for Zeus / Zap
  - [ ] show the own watchtower service address
  - [ ] connect third party watchtowers
  
Then could change dynamically depending on what is installed:
- [x] LND connection string for BTCpayserver (from LNDCREDS)
- [x] Electrum Server from the ELECTRS menu
- [ ] Can consider showing the various ports to open different services
- [ ] Hidden services for the various services

rootzoll:
Just quick note: There is already a "CONNECT" menu entry for connectig to another Lightning peer. We could change that to "PEERING" ... or will that get a sub entry in the new "CONNECT" menu?

openoms:
I wonder if we could declutter the Main menu significantly.
Currently with only LN related services installed I can only see 33% of the options:
![image](https://user-images.githubusercontent.com/43343391/112274943-f470e100-8c76-11eb-964f-574fc28b7454.png)

I am thinking to breakdown to a couple of submenus:
```
INFO - for the status screen
LIGHTNING - submenu for LN wallet related actions
    FUNDING
    CONNECT - can be called PEERING
    CHANNEL
    SEND
    RECEIVE
    NAME
    CLOSEALL
    CASHOUT
    lnbalance
    lnchannels
    lnfwdreport
SERVICES - all the installed app info-s including:
    TOR - show nyx + dilaog editbox for toorc
    LND - logs + dialog editbox for lnd.conf
    BITCOIN - logs + dialog editbox for bitcoin.conf
ADD/REMOVE - instead of the current SERVICES menu to install apps 
SETTINGS - kept as it is with the toggle options
[...]

PR #2116, opened 2021-03-26
Remove all Tor v2 services from the defaults

openoms:
would be probably the safest to not have any v2 Tor services configured by default in case it brakes some Tor versions, like it happened here: https://gitlab.torproject.org/tpo/core/tor/-/issues/40348
and expected network wide from July 2021. 

The entry for Bisq can be added to be set up as part of a BISQ option in the CONNECT menu: https://github.com/rootzoll/raspiblitz/issues/2090

The other fallback services seem to be redundant now.


openoms:
merged in #2148 closing

PR #2133, opened 2021-04-03
Update stacking-sats-kraken version

dennisreimann:
Two important additions:

- Trading agreement option, fix for german customers for which this might be required
- Market buy option (limit orders might not work in case of bullrun)


rootzoll:
Looks good :) will be part of v1.7RC2

PR #2135, opened 2021-04-03
remove rpcpassword related code for lnd/btc config syncing

nygellyndley:
70initLND.sh pulls the user's rpcpassword for the relevant network then
uses a regex sed replace to add this into the lnd config.
This requires a `<network>.rpcpass=` entry in order to swap in the password.
This commit adds that entry back in.

### DIFF
--- a/home.admin/50copyHDD.sh
+++ b/home.admin/50copyHDD.sh
@@ -213,9 +213,6 @@ fi
 # if started after intial setup - quit here
 if [ "${setupStep}" = "100" ]; then
   sudo cp /home/admin/assets/bitcoin.conf /mnt/hdd/bitcoin/bitcoin.conf
-  rpcpass=$(sudo cat /mnt/hdd/lnd/lnd.conf | grep 'bitcoind.rpcpass' | cut -d "=" -f2)
-  sudo chown bitcoin:bitcoin /mnt/hdd/bitcoin/bitcoin.conf
-  sudo sed -i "s/^rpcpassword=.*/rpcpassword=${rpcpass}/g" /mnt/hdd/bitcoin/bitcoin.conf 2>/dev/null
   sudo systemctl enable bitcoind
   echo "DONE - rebooting: sudo shutdown -r now"
   sudo shutdown -r now
--- a/home.admin/70initLND.sh
+++ b/home.admin/70initLND.sh
@@ -82,8 +82,6 @@ if [ ${configExists} -eq 0 ]; then
   sudo mkdir /mnt/hdd/lnd 2> /dev/null
   sudo chown -R bitcoin:bitcoin /mnt/hdd/lnd
   sudo cp /home/admin/assets/lnd.${network}.conf /mnt/hdd/lnd/lnd.conf
-  source <(sudo cat /mnt/hdd/${network}/${network}.conf 2>/dev/null | grep "rpcpassword" | sed 's/^[a-z]*\./lnd/g')
-  sudo sed -i "s/^${network}d.rpcpass=.*/${network}d.rpcpass=${rpcpassword}/g" /mnt/hdd/lnd/lnd.conf
   sudo chown bitcoin:bitcoin /mnt/hdd/lnd/lnd.conf
   if [ -d /home/bitcoin/.lnd ]; then
     echo "OK - LND config written"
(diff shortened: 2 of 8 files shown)


rootzoll:
@nygellyndley thx for finding this in the RC1

@openoms you removed the RPC from config because LND (when running as the same user as bitcoind) can now self/auto configure .. so wee need to clean up 70initLND still, correct?

openoms:
Yes, we rather need to remove the sed line from 70initLND script as there is no need to have passwordb in lnd.conf when lnd has access to the bitcoin.conf anyway.

Good finding @nygellyndley!

openoms:
Will need to remove the checks for rpcuser and rpcpassword from:
https://github.com/rootzoll/raspiblitz/blob/9da5eb879b668ffb6566caeebec20cbb3c144300/home.admin/config.scripts/lnd.check.sh#L145
as well.
@nygellyndley are you ok modifying this PR?

nygellyndley:
PR updated.

It's not often I Github fork, I'm more Bitbucket, it's now two commits, I can recreate the PR with one change if need be.

`70initLND.sh` has been updated to removed the uneeded password insertion lines.

There does seem to still be an issue though but I'll leave it to you to resolve.

Specifically the issue that prompted me to submit this fix was this post build :

image

But as opennoms says adding the rpcpassword into `lnd.conf` should not be required due to the lnd process using the btc user.

After a bit of digging around this error seems to occur due to the 10 minute wait conditional here : https://github.com/rootzoll/raspiblitz/blob/v1.6/home.admin/80scanLND.sh#L124

Essentially on first build it's doing a lot of verifying, 10 mins gets exceeded on the first run through, and if you restart on seeing the error (as suggested) you see it again as the blockchain and lnd wallet verification still isn't complete.

So really, despite the error complaining about passwords, it seems to be a bit of a red-herring.  And my initial fix adding an rpcpassword stub into the base file likely only worked because I took longer than 10 minutes to figure it out!

I don't know enough about what you'd expect on a build to submit a patch for this, but I think you can expect a "fail" when only waiting 10 minutes.

and probably nuke this whole section : https://github.com/rootzoll/raspiblitz/blob/v1.6/home.admin/config.scripts/lnd.check.sh#L161

FWIW my build scenario was a reuse of an existing btc blockchain on the /mnt/hdd plus an import of a pre-existing lnd wallet via seed phrase, but with no lnd data on the drive.

openoms:
@nygellyndley all good.
You can add an other commit to remove the check mentioned here: https://github.com/rootzoll/raspiblitz/pull/2135#issuecomment-812917848

nygellyndley:
Sorry missed your #2135 comment earlier.  Done.

But I'm now looking at :

1) https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/config.scripts/lnd.credentials.sh#L181
I feel this should be removed.  There may be an argument for leaving it in, in the event that a user has explicitly provided the rpc creds in `lnd.conf `,  but I'm not sure this is now a valid scenario so would lean towards removing this.

2) I think I'd be tempted to remove these too, but as above it's not clear to me that there isn't some valid use-case :
- https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/config.scripts/blitz.setpassword.sh#L222
- https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/config.scripts/blitz.setpassword.sh#L227
- https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/config.scripts/blitz.setpassword.sh#L228

3) There's also this which seem like it might copy the pw from lnd back into btc in some scenario https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/50copyHDD.sh#L216  I'm thinking dump `216` and `218` since they won't function but I'd worry this bit of code will fail if a user is importing lnd but is missing btc.  (perhaps this is now a non-working scenario with 1.7?)

4) This which seems to be a similar recovery mechanism to (3) above, I'm thinking remove these lines? https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/_bootstrap.provision.sh#L607

I can add another commit for these fixes if you can clarify.

openoms:
All lines concerning the rpc password and rpc username in the lnd.conf should be removed together with the sed lines using the variables generated by extracting the password from the lnd.conf.

This simplifies the code which is beneficial especially concerning passwords are handled less.

rootzoll:
Perfect PR update @nygellyndley  .. I just ran into the same issues on testing. All removals look good - will merge into dev.

Issue #1604, opened 2020-09-30
onion socket 9735 is closed when restarting Tor


holgern:
I had the problem that the tor socket 9735 was not reachable. I checked it by installing
```
sudo apt-get install proxychains4 nmap
```
on a server and checked the port with:
```
proxychains4 nmap -Pn -p 9735 -sT -v my_onion_addr.onion
```
The command returned a 
```
PORT     STATE SERVICE
9735/tcp closed unknown
```

Only a 
```
./config.scripts/internet.tor.sh off
```
reboot
```
./config.scripts/internet.tor.sh on
```
reboot
did repair the port.
Now I see:
```
PORT     STATE SERVICE
9735/tcp open  unknown
```

It is also possible to do the proxychains4 command on the raspi itself. 

Restarting the tor service and removing the ethernet cable did not helped.

holgern:
After testing, I found out that the port 9735 is closed every time tor is restarted due to installing something on the raspiblitz.

holgern:
I found the fix:
lnd must be stopped before tor is restarted

```
sudo systemctl stop lnd
sudo systemctl restart tor@default
sudo systemctl start lnd
```

openoms:
you can get the LND HS address with:
`echo "$(lncli getinfo)" | grep "uris" -A 1 | tr -d '\n' | cut -d '"' -f4 | cut -d @ -f2 | cut -d : -f1`
so testing with:
`proxychains4 nmap -Pn -p 9735 -sT -v $(echo "$(lncli getinfo)" | grep "uris" -A 1 | tr -d '\n' | cut -d '"' -f4 | cut -d @ -f2 | cut -d : -f1)`

It always gives me:
```
PORT     STATE  SERVICE
9735/tcp closed unknown
```

I often find that if some of my peers are disconnecting restarting tor (and often lnd too) helps.
It is sufficient to do:
```
sudo systemctl restart tor
sudo systemctl restart lnd
```

holgern:
I'm getting a denied when I only restart tor.
```
:9735 <--denied
```
This is the problem, when installing something on the raspiblitz. 

I don't know if this happens only to my installation, or if this is a  general problem.
I'm on the current patched github for branch 1.6 (release 1.6.1), so I have included the PR #1571 (I did all the necessary changes on my system)
A 
```
sudo systemctl restart tor@default
sudo systemctl restart lnd
```
fixes the closed port.

openoms:
[...]

PR #2137, opened 2021-04-04
LCD: make peers and URI red if unreachable on Tor

openoms:
as discussed in: #1604

(duplicate of https://github.com/rootzoll/raspiblitz/pull/2136 which was reverted)


openoms:
The commit is part of: https://github.com/rootzoll/raspiblitz/pull/2148 so closing this.

PR #2147, opened 2021-04-06
Add mempool to update menu

deverickapollo:
- [x] Tested on RaspiBlitz v1.6.3 
- [x] Mempool update pulls latest changes from git repo and rebuilds mempool service. 
- [x] Mempool service restarts after update

Current implementation pulls update from working repo.  Should we install from official releases, instead?


rootzoll:
Your checking for the latest release and your checking that version out ... looks good. I will squash merge the PR into `v1.6-interim` and then cherry-pick it over into the `dev` to be part of v1.7 release. Much thx.

PR #2148, opened 2021-04-06
Run LND with a separate Tor instance to avoid restarts

openoms:
As discussed in:
https://github.com/rootzoll/raspiblitz/issues/1604#issuecomment-813867108

This PR adds  a dedicated tor instance for LND.

Includes https://github.com/rootzoll/raspiblitz/pull/2137.

The Tor self test is down in 00infoBlitz.sh and peers and URI changes to red if LND is unreachable over Tor.


rootzoll:
Do you consider PR ready to be merged for v1.7RC2?

openoms:
Yes, running with these settings.

openoms:
Just got an issue this am with the to@lnd instance failing due to permissions.
It might be logrotate related which is changing files back to `debian-tor`:
```
$ sudo ls -la /mnt/hdd/tor-lnd/
total 154760
drwx------  3 _tor-lnd   _tor-lnd        4096 Apr  8 00:00 .
drwxr-xr-x 13 root       root            4096 Apr  6 16:24 ..
-rw-r-----  1 debian-tor debian-tor         0 Apr  8 00:00 info.log
-rwx------  1 _tor-lnd   _tor-lnd   158069673 Apr  8 00:00 info.log.1
-rw-r-----  1 debian-tor debian-tor         0 Apr  8 00:00 notice.log
-rwx------  1 _tor-lnd   _tor-lnd      378916 Apr  8 00:00 notice.log.1
drwx------  3 _tor-lnd   _tor-lnd        4096 Apr  8 00:00 sys
```
Need to sort this out before merging.

openoms:
ok, fixed the log ownership issue with: https://github.com/rootzoll/raspiblitz/pull/2148/commits/189be5054dcd6d24ee8cdbc681a537c0e9cadb94

@rootzoll now this is ready to be merged

openoms:
removed the unnecessarily heavy (and frequent) tor selftest in: https://github.com/rootzoll/raspiblitz/pull/2148/commits/c5841b9367186e75668ab537ab1af4d4df4b3ddf

The problem against which the test was needed is solved with the separate `tor@lnd` instance and only restarting `tor@default` in scripts.

The self check can also be reintroduced by applying the changes in the last commit.

nyxnor:
@openoms As a new instance of Tor is created (tor@lnd), independent from tor@default in both circuit and configuration, maybe this traffic should use bridges for who needs it. What do you think, tor@lnd needs bridges too when Tor connections are blocked?

openoms:
it only makes sense to configure the bridges for all instances.
Would be good to use a script (or function) which is reusable for possible multiple other instances.
See: 
https://github.com/rootzoll/raspiblitz/blob/d432f86ff79b608e34c412afdfa58dbaa3ee2c15/home.admin/config.scripts/internet.tor.sh#L94

nyxnor:
>Some things to do (to no forget this) Use a bridge option for this script (could use my questions script patch-5, or modify it to be inside a msg box) and always make a backup of the bridges info (it is easier to save a backup of the bridges when entering it with Edita script compared to greping from the file, which could miss some information, empty lines would split). This would facilitate if more instances are implemented in the future.

>When the bridges are entered in the `torrc` file placed on the home folder when building the sd card, or when using the interntet.tor.sh bridge option (to be implemented), save a backup of the bridges and always copy them to `/etc/tor/torrc` and `/etc/tor/instance/lnd/torrc` (if someone wants different bridges, would need to be manually written or a question on the script asking if they want to configure different bridges, but for now is not necessary, even though it would increase anonymity.).

>The problem with greping bridges info, if they are spaced like this (would fail because of the empty line):
```
Use Bridges 1

Bridge ip:port
```
>So include dir `/etc/tor/torrc.d` and `bridges` inside it would make more sense, just like the hidden services, to clear one file or cat all of it.
>I will start a draft bridge option for internet.tor.sh. I also need to update build_sdcard.sh to include bridges for the lnd instance....
Edit: If you know how to appear multiple lines on inputbox, Im duckduckgoing but no good results so far...

Edit:
Ok, pull request done to fulfill this needs /pull/2172

Issue #2012, opened 2021-02-18
Termbin is not printing the all the debug logs


nyxnor:
Please consider saving the debug logs to a file, so the termbin prints the entire contents. 
This is because there is a time out, termbin just run debug again with a time out, doesnt get the previous outputs.

If I run `debug`, it prints the file and transfer to var
if I run `debug -l`, do all this above plus print the termbin link below. 
This solves the issue of not printing the entire debug logs.
Also, the end path of termbin on the XXdebugLogs.sh will need to be updated with the path and file name you guys choose.

I edited the debug alias on _commands.sh:

```
# command: debug
function debug() {
  cd /home/admin
if  [[ $1 = "-l" ]]; then
        bash XXdebugLogs.sh > /var/cache/raspiblitz/debug.log && cat /var/cache/raspiblitz/debug.log | nc termbin.com 9999
else
        bash XXdebugLogs.sh > /var/cache/raspiblitz/debug.log && cat /var/cache/raspiblitz/debug.log
fi
}
```

But, if you dont want to add another script file, just add the code to _commands.sh

Just found there is a onion pasting sites like this https://p.pastly.xyz/ (Will redirect when using tor browser).
If someone knows how to http post there.... cause else termbin.com will still know who is posting.

rootzoll:
Thanks :) Will take a look for the next release.

nyxnor:
Well, first decent pull request.
Closing this now, continue on the pull request above.

PR #2154, opened 2021-04-08
print debug logs and blitz help on terminal

nyxnor:
https://github.com/rootzoll/raspiblitz/issues/2012
print debug logs fully and link option. Raspiblitz help menu

Hope this pull comes without another 999 pulls...


openoms:
good idea, but couple of notes:

lndmanage (`manage`) will be not included in the GUI in 1.7
the users `loop`, `pool` and `faraday` will be deprecated and the install scripts to be left CLI only as well because all their interfaces will be the part of the `lit` user.

rootzoll:
I also like having a help menu for the commands ... I mark this for v1.7.1

nyxnor:
@openoms `manage` wont be on the gui. So should I exclude from this help menu as one just will get to it through command line?

openoms:
yes, since lndmanage is not supported currently (hopefully temporarily) I suggest to no put it into the help text either (same with the separate loop, pool and faraday)

Issue #2054, opened 2021-03-14
Tor domain, torproject.org blocked in authocratic surveillance regime jurisdiction


pretyflaco:
During initial setup up of Raspiblitz v1.6.3, the blitz tries to fetch new packages from https://deb.torproject.org and fails because the domain is blocked by ISP (national regulation). I think raspiblitz can still fall back to the package version from the img-file, so not critical at the moment. If tor upgrade becomes necessary in the future, it could become a problem though if there's no mirror to fallback to. Maybe it would be possible to add alternate sources to fetch?

![Screenshot (19)_LI](https://user-images.githubusercontent.com/32958103/111066679-58561580-84d1-11eb-8fe8-61dd8e9eabd9.jpg)

rootzoll:
The v1.7 will have "Tor by default" and already have it installed on sd build-script - so it will be part of the binary.

But I keep this issue open as if anybody has an idea how to tackle this problem in general .. because it should stay possible also for people to build the sd card image themselves in such regions.

NilByte:
I would try setting static DNS not provided by your ISP in your router.

Try these ones from Cloudflare and Google.

![image](https://user-images.githubusercontent.com/36365571/111085031-bf95b900-8515-11eb-95f0-5b3c7768199f.png)

nyxnor:
# Flaco said NO to CENSORSHIP, Flaco did it boys.
##  This is the most updated code, the one Flaco mentioned below, we did not had built obfs4proxy from source, so the bridges were not working at that time.

**Considerations**
* Compile Tor from source seems to be the only way to get the security fixes, as updating from Tor repo is not working on armhf. The script bellow is to make possible for users to compile Tor from source if the understand the changes to be made, and make possible for users in the same situation to protect themselves with bridges.
[...]

PR #2155, opened 2021-04-08
Build sd card even when tor domain or tor connections are blocked.

nyxnor:
Related to Tor domain blocked. This changes also solve if Tor is blocked, as bridges will be added before Tor starting.
https://github.com/rootzoll/raspiblitz/issues/2054

This is my big boy.
I can explain why every line needs to be on the script, but if you find things to be improved, Im open to changing what you may find necessary.

`build_sdcard.sh`
1. Separate sources for ease to use. deb-src can be uncommented only when necessary.
2. Mask Tor
2. Add distribution sources
2. Install Tor from distribution repository
2. Install bridges package from source (only way to make it work)
3. Add bridges
4. Unmask and start Tor
5. Test Tor connection
6. Add Tor keys with torsocks
7. Add Tor from Tor Project repo
8. Install Tor to the latest version from TP repo

`config.scripts/internet.tor.sh`
1. Just not override bridges from building sd card (tor-hs-patch-config would be better)
2. Uncomment tor sources and comment again.

Edit: Brief note:
Acquiring Tor is made reaching the linux distribution, https with debian and http with ubuntu. A more secure way that would also work if they block the distribution (some govs block wikipedia....), is downloading Tor sources with the Tor Browser and SCP from the desktop to the Blitz, building Tor from source. (Also you would need to modify the script to fit the needs, but this is an overkill for now, will modify if someone that needs it open an issue).


rootzoll:
Wow thanks - the option for Tor bridges would be quite a good addition .. @openoms can you take a look?

This would be a quite big change to the build script quite close to the v1.7 release - review and test might delay us a bit. The next change-window for the build script would be then with the v1.8 release (about 2-3 month) ... I think its PR is valuable and fits into the focus of v1.7 with TorByDefault. But if this is something to be refined/tested longer I would also think its OK to push it to v1.8 .. any opinion?

@nyxnor the `bridgesQuestion` is kind of adding an interactive question during the script running. On design goal of the build script is to have all input as parameters on calling the build script already (with a default value if no parameter was given) so that the build script runs thru without a stop ... that way it can be used later in vagrant/docker scripts etc. Can you adapt on that?

openoms:
Thanks for this, building the SDcard in a trustless way in a censored environment is very important and also the easy use of Tor bridges is great.

Re. the `build_sdcard.sh`:
being a rather large change I'd like to leave a more time for review and testing.
The solution is not required for the SDcard release by @rootzoll and it can be made available for the manual build any time after the release. Then after testing there is no problem to make it the part of the script used to build the next (v1.8) release.

Re. `internet.tor.sh`:
This will be complicated slightly by introducing a separate Tor instance for LND in https://github.com/rootzoll/raspiblitz/pull/2148 which greatly affects the reliability of Tor-only LND nodes.
I suggest to take a second look at it after that is merged to `dev`.
@nyxnor do you think changing the `internet.tor.sh` could be made to be a separate PR or is it dependent on the changed `build_sdcard.sh`?
The changes can be the part of this (v1.7) or the next minor release (v1.7.1) if need further work and review.

rootzoll:
Because v1.7.0 release is already close the idea would be to prepare an alternative build script for v1.7.1 and have that available for further feedback. And for v1.8 we can then see if we can merge this into the official build script as a parameter option.

nyxnor:
@rootzoll The only way I can see this working with no interaction is (If you agree with that, I can make these changes):
* moving bridge question to the menu, so it will be user to configure bridges (but that is just after the setup, just to reuse this code later)
1. A torrc file placed on the desktop of the new flashed raspios arm64. But the user will need to know how to configure bridges.  Maybe moving the info I wrote about bridges to a topic on the README near `Build sd card` and then placing some examples of different types of bridges that can be set on torrc.
2. This way, it would be 2 parameter option (1) Include `torrc` file from the desktop set to `true`. (2) Dont ping Tor Project Domain and add bridges.


>@openoms [SOLVED below merging dev branch, but will test with the build_sdcard.sh params above when ready]
It is dependant because the bridges will be overwritten if not done those changes. Your PR will pass earlier and I plan to adapt according to it.
There is only 2 things there (`internet.tor.sh`), 
>* uncomment deb-src that I let commented on build_sdcard.sh and then commenting again after updating Tor. (This can be removed, so I will need to let the sources on the build uncommented)
>* merge (concatenate) bridges from the build_sdcard.sh with the new lines that will be introduced on this file. (This cannot be removed, bridges will be overwritten....)

nyxnor:
But let me test first all the possible outcomes. I will ping again when ready.

`build_sdcard.sh`
* Ok, updated without user interaction.
* If user chose to use bridges, but no torrc file found on home folder to be moved to /etc/tor/torrc, the build will stop. Maybe this could be at the very beggining.....
* an user that need hardened config should use
'* = whatever
build_sdcard.sh true* false hisUserName hisBranch false* tweakBootdrives* false false true

`internet.tor.sh`
* the `bridgesQuestion` will be better suited there. So now the user needs to know how to configure torrc before running the build_sdcar.sh and place it on the home folder of the OS, which can be described on the Readme or Wiki.
* But I will do it on another PR, cause it will be another step and is not dependent.

nyxnor:
Erased older info because it was related to RC2.
Current info is that this commits are ready, they worked with Flaco building his Blitz.
New info is non-interactive sd card build. https://github.com/rootzoll/raspiblitz/issues/2242
Do you guys want me to move this internet.tor.sh @rootzoll? It is gonna change a lot of things there, but someday this will have to be done either way.

Edit: Probably this wont be merged due to non-intereactive setup process for the build_sdcard.sh. So I made the other related pull request independent from this one, where bridges can be added through whiptail, it also acquires tor sources there.

nyxnor:
Closing this PR regarding [non interactive setup process of the build script. ](https://github.com/rootzoll/raspiblitz/issues/2242)
This code will be integrated only to `internet.tor.sh` in my future PR, so all the commits to `build.sdcard.sh` regarding Tor does not makes sense.
I may take some time to finish all the Tor menu, but it will be worth the time. Bye.

PR #2169, opened 2021-04-11
Update README and FAQ to mention about Tor Bridges code and build_sdcard.sh

nyxnor:
related to https://github.com/rootzoll/raspiblitz/pull/2155

PR #2172, opened 2021-04-12
add bridge option for internet.tor.sh

nyxnor:
I think this should be on the menu also... is it the 00settingsMenuBasics.sh?


rootzoll:
@openoms can you take a quick look? Is it OK to add to v1.7 kind of last minute or wanna push to v1.7.1?

home.admin/config.scripts/internet.tor.sh
```diff
@@ -302,7 +308,7 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
   sudo chown -R debian-tor:debian-tor /mnt/hdd/tor
 
   # create tor config .. if not exists or is old
-  isTorConfigOK=$(sudo cat /etc/tor/torrc 2>/dev/null | grep -c "BITCOIN")
+  isTorConfigOK=$(sudo cat $torrc 2>/dev/null | grep -c "BITCOIN")
```
openoms: For this check we should search for `Bitcoin`, the caps are removed form the default config.
```
$ sudo cat /etc/tor/torrc 2>/dev/null | grep -c "BITCOIN"
0
$ sudo cat /etc/tor/torrc 2>/dev/null | grep -c "Bitcoin"
1
```

openoms:
As it is currently `config.scripts/internet.tor.sh bridge` only works when Tor is already installed.
Should be able to use
`config.scripts/internet.tor.sh on bridge`
to install Tor with bridges set up to start with so it does not attempt to connect without them.

Please correct me if the workflow is meant to be different.

openoms:
The documentation in https://github.com/rootzoll/raspiblitz/pull/2169 is excellent, I think it would be great to have a link provided in the whiptail dialog to it when setting up bridges.

home.admin/config.scripts/internet.tor.sh
openoms: ok, since this was a bug I introduced, fixed in https://github.com/rootzoll/raspiblitz/pull/2172/commits/c8302691c865499ff6b6242fe30af88a44ae8841

nyxnor:
@openoms You are right, should be a secondary option when setting Tor on.
About the link being provided in the whiptail dialog, yes, should be done also.

About bridges being a secondary option, what do you think about 2 options in the menu -> `Tor` and `Tor+Bridges`. So a normal user dont have to refuse bridges when setting Tor, but just the person in need should answer these questions by default.
Awaiting response to make the changes.

rootzoll:
This sounds all great - I push ist to v1.7.1 so that we have a little more time to iron out the details.

openoms:
@nyxnor yes, having an extra `Tor with Bridges` option is a good idea to keep it more simple for the casual user.

@rootzoll agree to aim for this to be included in v1.7.1

nyxnor:
Ok, moving the bridges part is now duplicated, 1-when creating the torrc files for default and lnd, 2-in the whiptail.
This wont cause problems as every part remove the bridges from the torrcs before adding them from the backup.
The reason for this is that a user might already have Tor installed, and for not deactiavating tor, the whiptail comes before checking if he has Tor connections. 
cab21db

nyxnor:
Ok, trying this for v1.7, I missed [this](https://github.com/nyxnor/raspiblitz/blob/2e2c92572f0d46965e416e0826ad5e37641eb110/home.admin/config.scripts/internet.tor.sh#L267) that stops tor if the option is not update. I don think this is needed...
Also, as discussed in another issue to minimize restart. I was prompt to restart that node after selecting the tor on bridge option, sometimes I just want to add bridges, but no need to restart the node for that.

Note, for this to work, need to [build obfs4proxy package from my other pull](https://github.com/nyxnor/raspiblitz/blob/d70711aa9aeb9bc63c5f3d7e99d0227816a1fee8/build_sdcard.sh#L292), which works when I build the sd card, but struggling to build after flashing the image provided due to permissions.
>/usr/bin/mandb: can't chmod /var/cache/man/pt_BR/CACHEDIR.TAG: Operation not permitted
/usr/bin/mandb: can't remove /var/cache/man/pt_BR/CACHEDIR.TAG: Permission denied
...
Reading package lists... Done
E: Can not find version '0.0.7-4' of package 'obfs4proxy'
E: Unable to find a source package for obfs4proxy
dpkg-buildpackage: error: cannot open file debian/changelog: No such file or directory

What fixed this was `sudo chown -R man:man /var/cache/man`
Ok, the rest of the fix (GOT IT TO WORK NOW):
```
echo deb https://deb.debian.org/debian buster main contrib non-free | sudo tee -a /etc/apt/sources.list.d/temp.list
sudo apt update
echo "*** Compile obfs4proxy from source ***" 
mkdir -p obfs4proxy
cd obfs4proxy/
sudo apt install -y obfs4proxy
sudo apt source obfs4proxy
sudo apt build-dep -y obfs4proxy
cd obfs4proxy-*
dpkg-buildpackage -b -uc
cd ..
dpkg -i obfs4proxy_*.deb 
sudo apt update
sudo apt --only-upgrade install obfs4proxy
cd ..
sudo rm -rf obfs4proxy
cd
echo "# Installed $(obfs4proxy --version)"
echo ""
sudo rm -f /etc/apt/sources.list.d/temp.list
```

The other issue about minimalising sd card build, probably Tor and obfs4proxy should be removed there prior to v1.8, adding it inside internet.tor.sh, but for now, building the sd card with my script works. The problem with doing this is to now if the package was already built from source or it was installed via apt. Via apt wont work, so that is the difficulty of adding an option to build bridges when toggling `on`. The option that I see is when setting `on bridge`, will always build `obfs4proxy` from source, as this is a rare case with low frequency, but tell me what you guys think.

Remember, when using the whiptail, you are supposed to just paste the bridge information there, no need to add the string `Bridge` in front, as it will be automatically added. The same way when using the Tor browser, you just  paste what your got from the bridges(dot)torproject(dot)org for simplicity and helping the user.

Questions:
Should it be built from source every time? I think it is not common to set tor to off, so I left the code there.

nyxnor:
Oh wow, TorBox is amazing.  Look how they acquire bridge: [Code](https://github.com/radio24/TorBox/blob/master/bridges_get.py).
Too good, only shell and python, I will need to take a deeper look into their code to add only necessary files and set paths.

There is much more to improve related to bridges with a menu as below, so this can be moved to later release after I manage to include the options below for the Blitz.
tmp


The image above is the menu-bridges from TorBox. The image below is the main Tor menu for the Blitz, the bridges menu will be called inside here and is gonna be the only submenu. The process was to extract the useful functions and scripts by them and combine into a single main menu. It is not ready yet, but here it is:
temp

nyxnor:
TorBox bridges options are much better than mine, so Im rewriting their `menu-bridges`.
For this PR, this need to be closed in favor of a future one, that will give more options for bridges.
The only thing that remains useful for this current one is the FAQ text, so this will be reused.

PR #2207, opened 2021-04-19
Minimize service restarts

PatrickScheich:
When IPv6 mode is enabled, sometimes interface gets briefly set to "::1" 
this would trigger the service restart of `bitcoind` and `lnd` 

This PR tries to minimize such restarts


rootzoll:
Thanks .. because its a change in the `background.sh` I like to give it a bit more test time and mark it for v1.7.1

rootzoll:
Will be part of v1.7.1RC2

Issue #2211, opened 2021-04-20
Update Bitcoin Core UX improvements


rootzoll:
With v1.7 we started to offer the update of bitcoin core #2183 ... some smaller UX improvements to keep oin mind for v1.7.1:

- [ ] enter strings with whiptail/dialog boxes
- [ ] warn on downgrades
- [x] make reboot after bitcoin core update (my lnd was in locked state after update with no UX to unlock)
- [ ] make sure that on update/recover same version gets installed again

PR #2214, opened 2021-04-21
unlock LND after Bitcoin Core install if ready

openoms:
As discussed in:
https://github.com/rootzoll/raspiblitz/issues/2183

Rather than a full restart LND can  be unlocked with the existing script.

### DIFF
--- a/home.admin/config.scripts/bitcoin.update.sh
+++ b/home.admin/config.scripts/bitcoin.update.sh
@@ -247,14 +247,16 @@ if [ "${mode}" = "tested" ]||[ "${mode}" = "reckless" ]||[ "${mode}" = "custom" 
 
   echo "# OK Bitcoin Core ${bitcoinVersion} is installed"
   if [ "${state}" == "ready" ]; then
+    echo "# Starting ..."
     sudo systemctl start bitcoind
+    sleep 10
     sudo systemctl start lnd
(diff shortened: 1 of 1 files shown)


openoms:
Continued from #2183 

>LND restarts when the bitcoin RPC interface is not accessible after a timeout (10s by defaul): https://github.com/lightningnetwork/lnd/blob/1a8d1874a545c456744ad26d0d3a64b9bed58f0e/sample-lnd.conf#L871

Because of this if the LND wallet is unlocked while bitcoind is still syncing LND can restart and lock again.
We could improve by introducing a check in `lnd.unlock.sh` to see if Bitcoin Core is 100% synced (`progress=1.0`) and only unlock LND then.
This might be a new issue or part of https://github.com/rootzoll/raspiblitz/issues/2211

_background.sh  (https://github.com/rootzoll/raspiblitz/blob/8659e641b3763eea4e2f81cd0daec92907ca1a13/home.admin/_background.sh#L363) is checking if LND is unlocked every 10 seconds an would notify on the display, in status or in the menu if called by `00mainMenu.sh` or `raspiblitz`.

The main difference in this PR is that it introduces the `lnd.unlock.sh` 20 seconds after Bitcoin Core restarts.
It might still not be ready then so this is not a complete solution. More to be done for the v1.7.1 as discussed in #2211

PR #2232, opened 2021-04-25
switch to mstmp (fix notifyme)

### LINKED ISSUE #2231: notifyme no longer works
Just noticed that the sstmp package is no longer included in the updated base image.

```
admin@raspberrypi:~ $ sudo apt install ssmtp
Reading package lists... Done
Building dependency tree
Reading state information... Done
Package ssmtp is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'ssmtp' has no installation candidate
```

Need to find an simple alternative.

frennkie:
fix #2231

### DIFF
--- a/home.admin/XXsendNotification.py
+++ b/home.admin/XXsendNotification.py
@@ -11,7 +11,7 @@ try:
 except ImportError:
     raise ImportError("Please install missing package: python3 -m pip install smime")
 
-SSMTP_BIN = "/usr/sbin/ssmtp"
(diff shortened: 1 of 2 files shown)


vv01f:
@frennkie pls edit the title… is there more to be done from your point of view?

rootzoll:
Merged to be part of v1.71RC2 ... please test with RC2 then if you need any further patches to be included in RC3 or final.

PR #2233, opened 2021-04-25
correct bug on debug for v1.7

nyxnor:
This does not impact usability, but just to present clean logs.
What happens is that if a service is `off`, it will still show the logs. The if statement needs to be base if it is `on` (it was, I introduced this bug on a PR a time ago, but just realized it now with v1.7.0 (So sorry, I was an enthusiastic at that time). This goes back to the services part of  v1.6 plus LIT service. Also, added torsocks to termbin temporarily, as onion service for hosting debug logs will soon be available.

To understand: Compare [what is happening](http://kq3nkqhpcb3qwxy7nc45hvimruchynvvhmyce4g2ohvcq5pqzl3ih3id.onion/nyxnor/debug.txt) with [my mods now](http://kq3nkqhpcb3qwxy7nc45hvimruchynvvhmyce4g2ohvcq5pqzl3ih3id.onion/nyxnor/debugmods.txt). It is my onion service for testing, the keys is 5CGIZWMBEPY342OJ5BEL4KKPZELCSR66GV57LVK3HOZFB32FSFJQ. This key will change soon.

### DIFF
--- a/home.admin/XXdebugLogs.sh
+++ b/home.admin/XXdebugLogs.sh
@@ -30,24 +30,24 @@ if [ ${#chain} -eq 0 ]; then
 fi
 
 clear
-echo ""
+echo
 echo "***************************************************************"
 echo "* RASPIBLITZ DEBUG LOGS "
 echo "***************************************************************"
 echo "blitzversion: ${codeVersion}"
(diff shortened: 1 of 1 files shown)


rootzoll:
moved most changes over to `blitz.debug.sh`and merged

PR #2265, opened 2021-05-02
Raspibolt logo when reaching terminal when exiting menu

nyxnor:
Im here once again saying sorry for making this PR to v1.7
IMHereonceagain

This is what it should look like:
asciiBlitz

The ASCII above is height=70
Just got the raspibolt logo https://stadicus.github.io/RaspiBolt/ and used this https://asciigenerator.net/. I would recommend just changing the height between 50-70.
This fits on the screen that I tested that is smaller than normal ones, but maybe you want smaller. The smallest I recommend is 50.
Useful --> No
Cool --> Yes
Should be implemented --> Maybe, tell me your thoughts

Also, I wanted to do this when sshing modifying `/etc/motd`, but maybe a smaller one like this:
```
                     _ _     _ _ _       
 _ __ __ _ ___ _ __ (_) |__ | (_) |_ ____
| '__/ _` / __| '_ \| | '_ \| | | __|_  /
| | | (_| \__ \ |_) | | |_) | | | |_ / / 
|_|  \__,_|___/ .__/|_|_.__/|_|_|\__/___|
              |_| 
```


rootzoll:
Merged and moved artwork to `_00raspiblitz.sh` where since the new setup process the exit to terminal info are.

Issue #2266, opened 2021-05-03
After 1.7 LiT installation, missing Pool account/funds


ghost:
I had pool with an active account and funds on 1.6.X. After upgrading to 1.7 and installing LiT, I no longer seem to have my existing pool account. I have active leases and funds so I'm a bit lost on what I'm supposed to do to recover those. 

It's a non trivial amount so any advise would be great.

rootzoll:
LIT is a new bundle of pool - its not using the old data. But you can reactivate the old pool single instance also which should have your old data with `/home/admin/config.scripts/bonus.pool.sh on` on terminal - funds will only show in the terminal version, not within LIT WebUI. After install see info on `/home/admin/config.scripts/bonus.pool.sh menu`.

ghost:
Doesn't work. 

```
pool@raspberrypi:~ $ pool getinfo
[pool] open /home/pool/.pool/mainnet/tls.cert: no such file or directory
```

ghost:
More problems. 

Reboot seemed to allow me to call `pool getinfo` but now with a different problem

```
pool@raspberrypi:~ $ pool getinfo
[pool] rpc error: code = Unimplemented desc = unknown method GetInfo for service poolrpc.Trader
```

but I can run some other stuff like accounts: 

```
pool accounts list
{
	"accounts": []
}
```

However seems like my pool info was deleted. 

reinstalling lit doesn't work because it doesn't remove the group properly. 

```
admin@raspberrypi:~ $  /home/admin/config.scripts/bonus.lit.sh on
making sure the lit service is not running
# INSTALL LIGHTNING TERMINAL
adduser: The group `lit' already exists.
```

Trying to recover funds with pool or lit-pool results in 

```
pool@raspberrypi:~ $ pool accounts recover
[pool] rpc error: code = Unknown desc = error performing recovery: could not recover account XXX: error during authentication, before sending subscribe: rpc error: code = Internal desc = payment required
```

Despite being well connected enough. 

I guess I'll wipe my device and try again? 

Some suggestions of things to fix so others don't lose funds like I have. 

1. Don't install LiT if there's an existing pool installation. 
2. Warn users that a pool uninstall will result in a loss of funds (your LiT installation seems to be deleting my existing pool data, but if you fix 1, then warn)
3. Uninstall LiT properly so that it may be reinstalled again


I have no idea why a well connected node is having payment problems to pool, I can only imagine something screwed up with the install in general.

[...]

PR #2276, opened 2021-05-05
Prevent Pool data loss

openoms:
Fixes the 
```
adduser: The group `lit' already exists.
```
problem on reinstall mentioned in:
https://github.com/rootzoll/raspiblitz/issues/2266#issuecomment-831411475


openoms:
merged into https://github.com/rootzoll/raspiblitz/pull/2331 so closing

Issue #2188, opened 2021-04-17
Add minimal placeholder for Hardware Test again


rootzoll:
For v1.7 we removed the old hardware test because it was not much needed anymore when most run RP4 
see: https://github.com/rootzoll/raspiblitz/issues/2185#issuecomment-821759248

But it can make sense for little things like checking if HDD is connected to USB3 (not USB2) #909 or check sd card speed #1063 etc and maybe bigger things in the future. So in preperation for the v1.8 refactoring we should have a small placeholder for a hardware check on startup/setup and one in the repair/settings menus. Also to be part of the debug log so that on support people can spot problems with the hardware.

So v1.7.1 should introduce a little placeholder for the hardware test again - it can be further improved then even after v1.8

nyxnor:
- [ ] Mem test https://github.com/rootzoll/raspiblitz/issues/517

rootzoll:
see: https://github.com/rootzoll/raspiblitz/issues/2185#issuecomment-830345792

rootzoll:
use new script `blitz.hardware.sh` for this - see branch `1126-provision`

rootzoll:
SSD Hardwaretest see: #2599

PR #2281, opened 2021-05-05
Remove documentation for non-existent feature

bubelov:


nyxnor:
There is an issue for this: [Add minimal place holder for hardware test](https://github.com/rootzoll/raspiblitz/issues/2188)
They didn't remove from the docs possibly because it was a change only between x.y.0 to x.y.1 versions.

rootzoll:
The placeholder in the doc seems to confuse people ... I will merge this for v1.7 - thanks.

PR #2284, opened 2021-05-07
Update lnd.rescue.sh

originaljeremy:
added instructions to check bytesize with stat since scp rounds file size on download and can be confusing


rootzoll:
Good idea .. I think it makes sense to give that instruction after telling to check for bytesize, but I will merge this for v1.7.1 and then finalize it a bit. @originaljeremy much thanks.

Issue #2306, opened 2021-05-16
SSH Tor serivce is not persisted after update


ghost:
Related to https://github.com/rootzoll/raspiblitz/issues/2273

I added SSH Tor service for allow to connect to the node following this process https://github.com/openoms/joininbox/blob/master/FAQ.md#ssh-through-tor-from-linux

After update, the ssh tor service is no longer accessible.

I like to suggest to persist this, because it may leave someone (like me now) lock out of the node when on the road.

nyxnor:
Do you mean persist the hostname after update?

>No longer accessible.

See Tor logs: `sudo journalctl -eu tor@default` to make sure it is functional.

ghost:
I had to do `./config.scripts/internet.hiddenservice.sh ssh 22 22` again.

Onion address was the same (hostname was persisted).

nyxnor:
If the hostname was the same, it did not create a new address, it used the same priv key. Probably was a Tor issue.
In this case, if Tor fails and you are on the road, setup ZeroTier as fallback option to debug and then ssh with torsocks again.

openoms:
the torrc is on the sdcard so settings get lost on OS updates.
Not a bad thing tbh since during the recovery there is a temporary revert to the default password. Better to not have remote access then.

nyxnor:
@openoms I see your point of the password being reverted to default can be a security issue when open ports to the internet. If using client auth with Tor, there is no problem with this, as the authentication comes before asking the password, so this is pretty secure, but not implemented on the code yet to advise to be done for now. So for the future, just want to keep this in mind.

ghost:
Good point with security after pw reset. However, can there be at least some warning after the first login (post update, post password change)? Something saying: "You had Tor SSH, but it was disabled during update. Please run `<this command>` to enable it again."

Because what happend to me, was that I did update, was happy that everything works so smoothly. Then started doing some stuff and then left home with plan to finish it later remotely. I didn't realize this and had to drive back home from quite far away.

nyxnor:
@peterflock By update away from the node, what do you mean? `Patch` command?

ghost:
@nyxnor No I mean `UPDATE`-`RELEASE` from the menu. I was updating from 1.6 to 1.7.

nyxnor:
[...]

PR #2317, opened 2021-05-20
add client authorization to onion services

nyxnor:
https://community.torproject.org/onion-services/advanced/client-auth/
https://github.com/rootzoll/raspiblitz/issues/2306#issuecomment-843562208

New keys overwrite previous one if they are created with the script.
auth on $service - generate 1 key. (I see no reason to loop creating keys as Blitz is supposed to be manage by only one person)
auth off $service - deletes all the keys

I did not repeat the key for services with Tor daemon in the code example at the end because it was not fitting the screen.


nyxnor:
I will possibly close this PR and all the others in favor of 1 single PR with separate commits for understanding, aiming for Tor everything. Let[s see.

nyxnor:
No more reason for this separate PR due to it being integrated to the menu now.
Credentials will be shown in the main menu where the address is also shown (maybe not, the address and key on the same screen can be dangerous, but will give the best usability)
Command line option will still be available, as the menu option calls for that script, but it will need to be rewritten.

temp

PR #2318, opened 2021-05-21
show electrs sync progress(percentage)

nyxnor:
My electrs is showing tipSynced=0 even tough it is on the same height. What happens is that the last 100 logs do not show  the last header. 1000 is showing, 2000 just to be safe.
Also, adding percentage synced if anybody asks how much has got left to sync electrs.


nyxnor:
Need more error handling to no divide by zero...

rootzoll:
Will take care about checking `$blockchainHeight` for non-zero aftter merge.

nyxnor:
Use netcat for RPC calls to get the blockheight: https://github.com/romanz/electrs/blob/master/doc/usage.md#rpc-examples

Issue #2176, opened 2021-04-13
99% Sync failing -- Problem Bootstraping


ag2121-ag:
My pi seems stuck at 99% and I ran the following and get

sudo grep -i tor /var/log/syslog | nc termbin.com 9999

```
[Apr 13 13:00:03 raspberrypi systemd[1]: /lib/systemd/system/fail2ban.service:12: PIDFile= references path below legacy directory /var/run/, updating /var/run/fail2ban/fail2ban.pid → /run/fail2ban/fail2ban.pid; please update the unit file accordingly.
Apr 13 13:00:04 raspberrypi systemd[1]: Started RaspiBlitz Background Monitoring Service.
Apr 13 13:00:04 raspberrypi _background.sh[5442]: Blockchain Sync Monitoring: peers=
Apr 13 13:00:11 raspberrypi bonus.sphinxrelay.sh[5521]: # ok - copied fresh config.json & app.json into dist directory
Apr 13 13:00:51 raspberrypi tor[681]: Apr 13 13:00:51.000 [notice] New control connection opened from 127.0.0.1.
Apr 13 13:01:11 raspberrypi _background.sh[5442]: Blockchain Sync Monitoring: peers=0
Apr 13 13:01:11 raspberrypi _background.sh[5442]: Blockchain Sync Monitoring: ZERO PEERS DETECTED .. doing out-of-band kickstart
Apr 13 13:01:11 raspberrypi _background.sh[5442]: #network.monitor.sh peer-kickstart
Apr 13 13:01:13 raspberrypi _background.sh[5442]: addressFormat='tor'
Apr 13 13:01:16 raspberrypi npm[5517]: 2021-04-13T12:01:16.691Z btcexp:app Loading historical data for chain=main
Apr 13 13:01:53 raspberrypi tor[681]: Apr 13 13:01:53.000 [notice] New control connection opened from 127.0.0.1.
Apr 13 13:02:55 raspberrypi _background.sh[5442]: Blockchain Sync Monitoring: peers=0
Apr 13 13:02:55 raspberrypi _background.sh[5442]: Blockchain Sync Monitoring: ZERO PEERS DETECTED .. doing out-of-band kickstart
Apr 13 13:02:55 raspberrypi _background.sh[5442]: #network.monitor.sh peer-kickstart
Apr 13 13:02:57 raspberrypi _background.sh[5442]: addressFormat='tor'
[...]

PR #2322, opened 2021-05-22
make sure right owner of /mnt/hdd/tor before last Tor restart

nyxnor:
I had this exact same issue with v1.7.0 https://github.com/rootzoll/raspiblitz/issues/2176
```
Directory /mnt/hdd/tor/web80/ cannot be read: Permission denied
```
Im not sure where is the best place to `chown`, but before the last restart fixes it. The other option I was thinking is [here](https://github.com/rootzoll/raspiblitz/blob/f571dca4857b800740f8c8e5fc980f32a0fcdb92/home.admin/config.scripts/internet.tor.sh#L369), but did that way just to be safe.

It is already [here](https://github.com/rootzoll/raspiblitz/blob/f571dca4857b800740f8c8e5fc980f32a0fcdb92/home.admin/config.scripts/internet.tor.sh#L302), I know, but is not working unfortunately for the blitz webui.

Also, if this is not corrected, the user will need to ssh or use a normal browser with some webui to correct this before using Tor browser.

### DIFF
--- a/home.admin/config.scripts/internet.tor.sh
+++ b/home.admin/config.scripts/internet.tor.sh
@@ -431,6 +431,10 @@ EOF
 }
 EOF
 
+  # make sure its the correct owner before last Tor restart
+  sudo chmod -R 700 /mnt/hdd/tor
+  sudo chown -R debian-tor:debian-tor /mnt/hdd/tor
+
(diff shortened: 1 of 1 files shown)


rootzoll:
Will be part of v1.7.1rc1

PR #2336, opened 2021-05-29
fix spelling of Web Browser

w3irdrobot:
"Web Browser" was misspelled. So I fixed it.


nyxnor:
delete the old lines, it is duplicated now instead of replaced.

PR #2340, opened 2021-06-03
Update 00infoBlitz.sh

FeedMyTummy:
Use headers for total number of blocks to sync.

Currently going through IBD. This is what it looks like with the current logic using getblockcount.

IMAGE 2021-06-03 13:39:47


nyxnor:
ACK 
tested fully synced.

rootzoll:
Thanks :) Will be part of v1.7.1rc1

rootzoll:
Now part of v1.7.1RC2 :)

PR #2360, opened 2021-06-21
Typo fixes

t0mmysm1th:
Tiny fixes here and there.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -14,7 +14,7 @@ echo "* RASPIBLITZ SD CARD IMAGE SETUP v1.7   *"
 echo "*****************************************"
 echo "For details on optional parameters - see build script source code:"
 
-# 1st optional paramater: NO-INTERACTION
+# 1st optional parameter: NO-INTERACTION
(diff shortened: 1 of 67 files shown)


rootzoll:
Thanks for going thru all those scripts. Will delay merge a mit after we merged in the major refactor for v1.7.1

build_sdcard.sh
```diff
@@ -14,7 +14,7 @@ echo "* RASPIBLITZ SD CARD IMAGE SETUP v1.7   *"
 echo "*****************************************"
 echo "For details on optional parameters - see build script source code:"
 
-# 1st optional paramater: NO-INTERACTION
+# 1st optional parameter: NO-INTERACTION
 # ----------------------------------------
 # When 'true' then no questions will be ask on building .. so it can be used in build scripts
```
RequestPrivacy, line 19: Found another typo right here: <When 'true' then no questions will be ask**ED** on building>

rootzoll: thanks. comitted to dev branch

PR #2367, opened 2021-06-22
Update Bitcoin Core version to 0.21.1

atef-zayati:


nyxnor:
nACK
This update is different, you can optionally upgrade with the menu to any version you want.
After November, the default core version can be chosen again.

openoms:
Changing the `build_sdcard.sh` is generally only the part of major releases so this would go for v1.8.0.

Until then the option to update Bitcoin Core from the UPDATE menu will remain.

atef-zayati:
Does copying `datadir` from Bitcoin Core 0.21.1 to Bitcoin Core 0.21.0 have any risks?

openoms:
nothing I am aware of, but you can upgrade Core any time. Just type `menu` and go to `UPDATE`

openoms:
Can rebase this against the `pre-1.7.1` as we decided to update the SDcard script for the next release to ship with updated binaries and c-lightning (#2476)

Would suggest to add a fallback link to download the Bitcoin Core signing key as in: https://github.com/openoms/joininbox/commit/8f49af65c063302c021102a296c8a718b2d76abe
because bitcoin.org had repeated issues lately.

```
if [ ! -f "laanwj-releases.asc" ];then
  wget https://raw.githubusercontent.com/bitcoin-dot-org/Bitcoin.org/master/laanwj-releases.asc
fi
```

openoms:
Closing this as it is a 1 character change included now in https://github.com/rootzoll/raspiblitz/pull/2476/commits/c75741fa40d9100b012cdfb5dbf098fd4dfb02d6 as part of https://github.com/rootzoll/raspiblitz/pull/2476

PR #2370, opened 2021-06-23
Avoid hardcoding options list size

bubelov:
I noticed that settings menu adds unnecessary scrolling if I add one more option to the menu. Looks like it caused by a bunch of magic numbers in this script. `OPTIONS` array is dynamic in size, but its size is statically set in the code. Please note that this patch doesn't solve the issue completely. There is an another variable  (`HEIGHT=19`), which should probably be dynamic too, but I'm not sure which formula to use. The comment states:

```bash
HEIGHT=19 # add 6 to CHOICE_HEIGHT + MENU lines
```

With this patch applied, we can figure out `CHOICE_HEIGHT` dynamically, but I'm not sure where `6` is coming from and how to figure out `MENU lines`

### DIFF
--- a/home.admin/00settingsMenuBasics.sh
+++ b/home.admin/00settingsMenuBasics.sh
@@ -75,7 +75,6 @@ echo "run dialog ..."
 # BASIC MENU INFO
 HEIGHT=19 # add 6 to CHOICE_HEIGHT + MENU lines
 WIDTH=45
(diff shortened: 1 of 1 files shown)


rootzoll:
Nice optimization :)

rootzoll:
Because we had a lot of other changes in the menus I took your way of calculating the height and adding it manually to the scripts. Closing this PR.

PR #2373, opened 2021-06-25
disable password auth btc-rpc-explorer

Tracachang:
By default password it is not activated, so I think it should be disabled since the script already create a specific user to run this service and rpc commands are disabled by default.

### DIFF
--- a/home.admin/config.scripts/bonus.btc-rpc-explorer.sh
+++ b/home.admin/config.scripts/bonus.btc-rpc-explorer.sh
@@ -151,7 +151,7 @@ BTCEXP_BITCOIND_PASS=$PASSWORD_B
 BTCEXP_BITCOIND_RPC_TIMEOUT=10000
 # Password protection for site via basic auth (enter any username, only the password is checked)
 # Default: none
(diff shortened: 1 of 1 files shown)


rootzoll:
Thanks ... will add for testing for v1.7.1 release.

rootzoll:

rootzoll:
Moving this to v1.8 because have to check if the btc-explorer API is really just info a public info service or can reveal any wallet information (also by API) or can execute any actions. Having the austh removed would make things easier and fit into the v1.8 WebUI. When removed also SSH UI infos need to be changed and password/config updates on changing password B.

Tracachang:
I do not think should be any risk as long as BTCEXP_RPC_ALLOWALL is false, which is the default

In settings:
 
#Enable to allow access to all RPC methods
 #Default : false
#BTCEXP_RPC_ALLOWALL=true

rootzoll:
Will make this change now for v1.8.0 - makes especially sense now that BTC-RPC-Explorter will be part of the first apps available thru the new WebUI. Just to keep in mind that after merge I also need to change some more places in the code:

- [x] docs
- [x] auth-info in status
- [x] update password b on change

PR #2374, opened 2021-06-25
privacy mode = true

Tracachang:
Privacy mode it is disabled by default


rootzoll:

rootzoll:
Will be part of v1.7.1rc1

PR #2375, opened 2021-06-25
npm audit fix

Tracachang:
After installing btc-rpc-explorer there are found 17 vulnerabilities (1 low, 4 moderate, 12 high), this can be easily fixed with "npm audit fix".

This is why I have added this command just after installing the explorer. This command should be added in any npm package


rootzoll:

vv01f:
@Tracachang did it similarly with checks in #2459 on a bonus feature. the original author mentiones the tree could be fixed, any ideas how to do that properly also for nested dependecies? would be nice to have less friction even if it is auto-fixed.

PR #2377, opened 2021-06-26
update Dark Science hidden service v2 -> v3

Tracachang:
From joinmarket release-notes-0.8.3:

Darkscience Tor onion address update
The optional Tor onion address in the default config is updated from the old v2 onion to the new v3 onion address. See here. Users should take note and update this in their config files, since v2 onions are going to be deprecated shortly.

1125120 darkscience tor address v3


nyxnor:
Use links to facilitate review 
[JM release note](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/release-notes/release-notes-0.8.3.md#darkscience-tor-onion-address-update) 
[Dark Sciente site](https://darkscience.net/servers/)
onion v3 address is correct.

ACK

openoms:
Thank you. 
On RaspiBlitz v1.7.0 JM v0.8.2 gets installed so users should update JoininBox to v0.5.0 then JoinMarket v0.8.3.

A config reset is recommended after version changes which takes care of the .oinon address.

Going forward planning to use the install script from JoininBox in the RaspiBlitz as well so there is no duplicate work done.

PR #2404, opened 2021-07-06
Feature/extensions

m00ninite:
- Add extensions menu
- Add a recurring keysend helper extension

https://user-images.githubusercontent.com/80651189/124531526-8c787d80-ddc3-11eb-8473-616dc7c02ac4.mp4

A short explainer on what's going on this demo vid:

1. User navigates to extensions menu (this menu is generated automatically based off scripts in `/home/admin/extensions`
2. blitz.recurringpayment.sh is selected, script can be previewed and edited before committing to run it
3. recurring payment script is run, new menus ask user for node pubkey, how many sats they want to send, and at what frequency
4. script spits out a new keysend script and a crontab entry to paste into crontab.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -73,7 +73,8 @@
   - [Let's Encrypt - DNS-01](#lets-encrypt---dns-01)
   - [Let's Encrypt - eMail Address](#lets-encrypt---email-address)
   - [Let's Encrypt - Installation details](#lets-encrypt---installation-details)
-  - [How can I customize my RaspiBlitz or add other software?](#how-can-i-customize-my-raspiblitz-or-add-other-software)
+- [How can I customize my RaspiBlitz or add other software?](#how-can-i-customize-my-raspiblitz-or-add-other-software)
+  - [Can I create my own extensions to distribute?](#can-i-create-my-own-extensions-to-distribute)
 - [Versioning](#versioning)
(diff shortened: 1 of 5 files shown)


openoms:
Thank you, loving this!

One possible improvement I can think of is to store the extensions on the disk so they are preserved during major updates with and SDcard (OS) change.
The usual place is `/mnt/hdd/app-data` where there could be an `extensions` directory owned by admin.

One problem with that is the disk is currently mounted with `noexec` in the fstab so nothing on it can be executable.
For now we need to store the extensions on the SDcard and redownload them after SDcard changes unless this changes #2405

m00ninite:
Could we automatically back up and restore the `extensions` folder when doing updates and manual backups? Unfortunately this wouldn't do anything for catastrophic failures of the SD, since the extensions would be stuck there.. but I envision extensions being fairly replaceable.

Alternatively, we could watch the `extensions` folder for changes and copy them over to the hdd when needed, a-la [channel.backup](https://github.com/rootzoll/raspiblitz/blob/fc1d44104bc22d85ebebd7b0d6ee2240722a9edf/home.admin/_background.sh#L231)

rootzoll:
I also like the idea of extensions ... but might delay it it bit to a v1.8.2, because we are changing quite a lot now on the menus and that will settle down after v1.8.0 release to make further modifications.

Issue #2412, opened 2021-07-07
[RTL] Sample-RTL-Config.json file is now under 'docs' folder


mbombonato:
I was trying to reinstall RLT today and notice the following error:

```
# Make sure rtl is member of lndadmin
# Add the rtl user to the lit group
# Symlink the lit-loop.macaroon
# Make the loop macaroon group readable
# RTL/RTL.conf
cp: cannot stat '/home/rtl/RTL/sample-RTL-Config.json': No such file or directory
chown: cannot access '/home/admin/RTL-Config.json': No such file or directory
chmod: cannot access '/home/admin/RTL-Config.json': No such file or directory
```

I checked and the file /home/rtl/RTL/sample-RTL-Config.json wasn't there. I searched and found the file at:
/home/rtl/RTL/docs/Sample-RTL-Config.json

Not 100% sure if the content is the same, but I think they just changed its path on the last release:
https://github.com/Ride-The-Lightning/RTL/blob/e45d6d598ad95c2789ee547a6dcf58d1799464d2/docs/Sample-RTL-Config.json

After alter the path on the bonus.rtl.sh script, the installation worked as usual.

PR #2414, opened 2021-07-07
correct to path docs/Sample-RTL-Config.json

nyxnor:
Issue https://github.com/rootzoll/raspiblitz/issues/2412


rootzoll:

PR #2428, opened 2021-07-20
Update internet.sh

rek79:
Fixed local IP grep command to find zero or more ")".
Simplified localip_ALL assignment.
Both of these commands now successfully retrieve local IP for Odroid HC2 using Armbian Buster.


rek79:

PR #2429, opened 2021-07-20
Update lnd.export.sh

rek79:
Simplified local IP assignment.  Original does not work on Odroid HC2 using Armbian Buster.


rek79:

PR #2430, opened 2021-07-20
Update 99connectMenu.sh

rek79:
Fixed localIPrange to work with Odroid HC2 Armbian Buster by adding "inet" to grep command.
Simplified localIP assignment.  Now also works with Odroid HC2 Armbian Buster.


rek79:

PR #2431, opened 2021-07-20
Update bonus.lndconnect.sh

rek79:
Simplified host local IP assignment and fixed to work with Odroid HC2 Armbian Buster.


rek79:

PR #2432, opened 2021-07-20
Simplify localIP detection and improve compatibility

rek79:
Simplified localIP assignment.  Now works with Odroid HC2 Armbian Buster.  Where "hostname -I" could not be used I added either "inet" or ")*" to the grep commands.  This now also works with Odroid HC2 Armbian Buster.


openoms:
ACK `cde3359`
@rootzoll tested this on multiple different systems, looks good and simplifies the IP address detection a fair bit.

rootzoll:
Thanks - merged for v1.7.1 release. To be tested witth RC1.

PR #2476, opened 2021-08-13
Add c-lightning and update binaries on the SDcard for v1.7.1

openoms:
as discussed we should ship the next release with the C-lightning binary already compiled on the SDcard

Adding it to the build script with EXPERIMENTAL_FEATURES enabled to be able to switch on features like:
* offers (BOLT12)
* dual funding
* liquidity ads

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -1059,30 +1059,101 @@ echo "- OK install of LND done"
 
 echo "*** C-lightning ***"
 # https://github.com/ElementsProject/lightning/releases
-CLVERSION=v0.10.1
+CLVERSION=0.10.1
+# https://github.com/ElementsProject/lightning/tree/master/contrib/keys
+PGPsigner="rustyrussel"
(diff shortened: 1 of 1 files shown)


openoms:
Updated LND to v0.13.1-beta and changed the PGPauthor to roasbeef's new PGP key.

Tested ok.

rootzoll:
@openoms we would now to make sure that the call of the script`/home/admin/config.scripts/cln.install.sh on mainnet`will detect that there is already a c-lightning install done. Have you tested that the script detects this?

openoms:
@rootzoll yes the
https://github.com/rootzoll/raspiblitz/blob/39afa8246f0ddeafdcbfbd78d773f44e043fcd42/home.admin/config.scripts/cln.install.sh#L46
checks for the presence of the `lightningd` binary and if present skips the compilation and 
and follows with the config steps from https://github.com/rootzoll/raspiblitz/blob/39afa8246f0ddeafdcbfbd78d773f44e043fcd42/home.admin/config.scripts/cln.install.sh#L126

rootzoll:
perfect. thx.

PR #2480, opened 2021-08-16
Add systemd hardening for bitcoind, lnd, electrs, and btcrpcexplorer

xanoni:
Added some systemd hardening that I've been running in production for months
without any issues.

Please note, I kept `ProtectHome` disabled, as RaspiBlitz seems to use several 
files from /home/ and I didn't want to make big changes.

As a first step, though, I moved the PID files to /run/ where they existed.

- bitcoind.service: add systemd hardening
- lnd.service: add systemd hardening
- bonus.electrs.sh: add systemd hardening
- bonus.btc-rpc-explorer.sh: add systemd hardening

### DIFF
--- a/home.admin/_bootstrap.sh
+++ b/home.admin/_bootstrap.sh
@@ -498,7 +498,7 @@ sudo chown bitcoin:bitcoin -R /mnt/hdd/bitcoin 2>/dev/null
 # FIX BLOCKING FILES (just in case)
 # https://github.com/rootzoll/raspiblitz/issues/1901#issue-774279088
 # https://github.com/rootzoll/raspiblitz/issues/1836#issue-755342375
-sudo rm -f /mnt/hdd/bitcoin/bitcoind.pid 2>/dev/null
+sudo rm -f /run/bitcoind/bitcoind.pid 2>/dev/null
 sudo rm -f /mnt/hdd/bitcoin/.lock 2>/dev/null
 
 #################################
--- a/home.admin/assets/bitcoind.service
+++ b/home.admin/assets/bitcoind.service
@@ -11,16 +11,26 @@ After=bootstrap.service
 [Service]
 User=bitcoin
 Group=bitcoin
-Type=forking
-PIDFile=/mnt/hdd/bitcoin/bitcoind.pid
 ExecStartPre=-/home/admin/config.scripts/blitz.systemd.sh log blockchain STARTED
-ExecStart=/usr/local/bin/bitcoind -daemon -conf=/home/bitcoin/.bitcoin/bitcoin.conf -pid=/mnt/hdd/bitcoin/bitcoind.pid
+ExecStart=/usr/local/bin/bitcoind -daemon -conf=/home/bitcoin/.bitcoin/bitcoin.conf -pid=/run/bitcoind/bitcoind.pid
+RuntimeDirectory=bitcoind
+RuntimeDirectoryMode=0710
+PIDFile=/run/bitcoind/bitcoind.pid
+Type=forking
 KillMode=process
 Restart=always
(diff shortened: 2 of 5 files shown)


home.admin/_bootstrap.sh
```diff
@@ -498,7 +498,7 @@ sudo chown bitcoin:bitcoin -R /mnt/hdd/bitcoin 2>/dev/null
 # FIX BLOCKING FILES (just in case)
 # https://github.com/rootzoll/raspiblitz/issues/1901#issue-774279088
 # https://github.com/rootzoll/raspiblitz/issues/1836#issue-755342375
-sudo rm -f /mnt/hdd/bitcoin/bitcoind.pid 2>/dev/null
+sudo rm -f /run/bitcoind/bitcoind.pid 2>/dev/null
```
openoms, line 501: can you please explain what is the benefit of moving the pid file?

home.admin/assets/bitcoind.service
```diff
@@ -11,16 +11,26 @@ After=bootstrap.service
...
 Restart=always
 TimeoutSec=120
 RestartSec=30
 StandardOutput=null
 StandardError=journal
 
+# Hardening
+PrivateTmp=true
+ProtectSystem=full
+ProtectHome=false # TODO: can we move the bitcoind stuff from /home/ elsewhere?
```
openoms, line 30: seems that this line is unnecessary if it is inactive.
Where would be better to have the process running other than the dedicated user home?

openoms:
Thank you for testing, but please see the changes in the current dev branch `pre-1.7.1` (yes should be moved to `dev`).
I have added some of the hardening to all systemd services there.

home.admin/_bootstrap.sh
xanoni: no benefit, think that `/run` / `/var/run` is just a more standard location for PID files

but no hard feelings about this

home.admin/assets/bitcoind.service
xanoni: The line does nothing, I just put it there as a placeholder ("for discussion"). I have all files that `bitcoind` and other daemons need in SSD folders that are not under /home/.

xanoni:
> Thank you for testing, but please see the changes in the current dev branch `pre-1.7.1` (yes should be moved to `dev`).
> I have added some of the hardening to all systemd services there.

That makes sense. I saw systemd hardening discussed but only found one service that had it enabled in `master` (`./home.admin/config.scripts/bonus.btcpayserver.sh`). Didn't realize you had that in a different branch.

xanoni:
> I have added some of the hardening to all systemd services there.

Okay I just looked at it and couldn't see `MemoryDenyWriteExecute=true` ... should that be added? It works well for `bitcoind`, `electrs`, and `lnd`. Only `btcrpcexplorer` is causing trouble with it.

openoms:
Need to test on a raspiblitz with the base image which will be part of the release.
We are planning to make an 1.7.1RC1 soon which will be for syncing up on the many changes and refactoring done since v1.7.

xanoni:
> Need to test on a raspiblitz with the base image which will be part of the release.
> We are planning to make an 1.7.1RC1 soon which will be for syncing up on the many changes and refactoring done since v1.7.

Okay. We can close this PR ... I have no hard feelings about the PID file move ... but I would suggest adding the `MemoryDenyWriteExecute=true`.

I also like the `ProtectHome=true`, because why not. BTW, for `electrs` you have to use `ProtectHome=tmpfs`, see here: https://github.com/romanz/electrs/issues/217#issuecomment-899295834

openoms:
ok, will close this for now as pending more testing.

PR #2481, opened 2021-08-16
Fix raspibolt reference links

RequestPrivacy:
The RaspiBolt Guide moved to GitHub Pages which broke some links in the build_sdcard comments.


rootzoll:
Thanks for updating the links

PR #2483, opened 2021-08-18
Upgrade stacking-sats-kraken version

dennisreimann:


dennisreimann:
Thanks! I always forget which branch to actually target. 🙈😉

Issue #2488, opened 2021-08-23
Force UASP support


dipunm:
I have a hdd enclosure, that has worked for me with UASP on, but I was using raspibolt. I've just "upgraded" to raspiblitz to make installation of stuff a bit easier.

The issue is that I am not confident that my testing is enough. Especially because in my testing, the HDD is about 2 - 3 inches away from the pi. Sounds unimportant, but I found that the wifi was not working properly when it was right under the pi with UASP on.

On top of that, my SSD runs MUCH slower with UASP off. Enough that I've wasted a whole week getting half the blocks to download and now that I have it on, there is a very noticable difference.

Long story short, I think it is a good idea to have a way for people like myself to force UASP to be enabled. I am not sure the best way to go about it right now.

My initial thought was to add a variable to the `raspiblitz.info` file, but I realised that the datadrive.sh file does not source this file so I am open to ideas. I didn't want to add a source of this file to the script because it might break something else, but I do want to ensure that my contribution is "fool proof" or at least robust.

I want a solution that will not require patching after a raspiblitz upgrade in the future.

xanoni:
(Hint: UASP = USB Attached SCSI Protocol)

The problem is that it's highly device-dependent whether UAS is supported at all and if it is, whether it is faster than `usb-storage`. My Samsung T5 uses `usb-storage` and performs well (almost 300 MiB/sec in sequential R/W benchmarks).

What is your kernel using now and are you sure it was using something else before? Try:

```bash
$ lsusb -t
/:  Bus 02.Port 1: Dev 1, Class=root_hub, Driver=xhci_hcd/4p, 5000M
    |__ Port 1: Dev 2, If 0, Class=Mass Storage, Driver=usb-storage, 5000M
[...]

$ dmesg | grep UAS
usb 2-1: UAS is ignored for this device, using usb-storage instead
```

Assuming your drive is using `usb-storage`, have you tried setting `usb-storage.quirks=$VENDOR_ID:$PRODUCT_ID` in your `/boot/cmdline.txt`?

rootzoll:
At the moment the RaspiBlitz is checking the the type of the ssd-driver/enclosure and just for models we know & tested the the UASP support kept on ... for every other its deactivated to prevent the bug reports we had on RaspBlitzes freezing during sync.

[...]

PR #2499, opened 2021-08-25
Uasp force

dipunm:
An implementation to address:
https://github.com/rootzoll/raspiblitz/issues/2488


rootzoll:
Thanks :) Will be part of v1.7.1RC2

PR #2500, opened 2021-08-25
Uasp force for the dev branch

dipunm:
An implementation to address:
#2488

### DIFF
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -386,7 +386,7 @@ if [ "$1" = "status" ]; then
     hddAdapterUSAP=0
     
     # check if user wants to force UASP on
-    if [ -f "/boot/uasp.force"]; then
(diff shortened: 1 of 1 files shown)


home.admin/config.scripts/blitz.datadrive.sh
```diff
@@ -1780,6 +1790,16 @@ if [ "$1" = "uasp-fix" ]; then
   else
```
dipunm, line 1790: Thoughts: 

The code above this basically checks if a new drive has been added that should have UASP turned off, and then if so, it clears all the quirks mode and re-applies the quirks mode for the new drive.

I'm thinking if people agree that if UASP=1, and it was on by a `uasp.force` file or `raspiblitz.conf` file, then we should remove the quirks mode. Otherwise, these options only prevent quirks mode, but they won't remove it for us.

rootzoll:
Was able to merge the other PR into the dev branch - closing this one

PR #2505, opened 2021-08-28
Identify newest BlitzPy version during sd building

RequestPrivacy:
Imho as an absolute newbie to coding the -r (reverse) flag in the ls command leads to an install of the oldest (currently 0.2.0), not the newest version (0.3.0) of BlitzPy.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -609,7 +609,7 @@ sudo -u admin cp -r /home/admin/raspiblitz/home.admin/config.scripts /home/admin
 sudo -u admin chmod +x /home/admin/config.scripts/*.sh
 
 # install newest version of BlitzPy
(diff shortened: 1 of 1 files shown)


RequestPrivacy:
`ls -trR /home/admin/raspiblitz/home.admin/BlitzPy/dist | grep -E "*any.whl"`
leads to a sorting as follows:
BlitzPy-0.3.0-py2.py3-none-any.whl
BlitzPy-0.2.0-py2.py3-none-any.whl

`ls -tR /home/admin/raspiblitz/home.admin/BlitzPy/dist | grep -E "*any.whl"`
to 
BlitzPy-0.2.0-py2.py3-none-any.whl
BlitzPy-0.3.0-py2.py3-none-any.whl

piping into tail -n 1 grabs the last line, therefore example no. 2 should be correct?

rootzoll:
Thanks - good catch. Merging for v1.7.1rc2

PR #2509, opened 2021-08-31
fix the instructions for the manual SD card build

openoms:
Just  a quick fix to extend instructions foregoing further complaints about the manual build.

Currently without parameters it downloads the scripts from the `dev` branch which is not stable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1133,7 +1133,7 @@ A ready-to-use SD card image of RaspiBlitz is provided by us for download, to ge
 
 Now you are ready to start the SD card build script (check the code to see if the installation and config are OK for you). Copy the following command into your terminal and execute:
 
-`wget https://raw.githubusercontent.com/rootzoll/raspiblitz/v1.7/build_sdcard.sh && sudo bash build_sdcard.sh`
+`wget https://raw.githubusercontent.com/rootzoll/raspiblitz/v1.7/build_sdcard.sh && sudo bash build_sdcard.sh "" "" "rootzoll" "v1.7"`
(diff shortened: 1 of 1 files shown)


rootzoll:
I think it makes more sense to make the build script default to the latest default branch. I comitted to dev a new version of the build script that for now defaults to v1.7 ... with the v1.71 release (when dev gets merged into v1.7) the old parameter-less instructions should work again. --> https://github.com/rootzoll/raspiblitz/commit/64930d7d65984353212b1e8593ead8c389333c39

rootzoll:
Tested my update of the build script ... looks good.
```
*****************************************
* RASPIBLITZ SD CARD IMAGE SETUP v1.7.1 *
*****************************************
For details on optional parameters - see build script source code:
1) will use NO-INTERACTION --> 'false'
2) will use FATPACK --> 'false'
3) will use GITHUB-USERNAME --> 'rootzoll'
4) will use GITHUB-BRANCH --> 'v1.7'
5) will use DISPLAY-CLASS --> 'lcd'
6) will use TWEAK-BOOTDRIVE --> 'true'
7) will use WIFI --> 'US'
X) will use CPU-ARCHITECTURE --> 'aarch64'
X) will use OPERATINGSYSTEM ---> 'raspios_arm64'
Do you agree with all parameters above? (yes/no)
```

@openoms I also changed default display to `lcd` from 'hdmi' .. because I think thats what people excpet by default if they build their own sd card to run on the recommended hardware.

Closing this PR.

PR #2512, opened 2021-09-04
Fix typo in systemctl command

periodic1236:
When enabling the btc-rpc-explorer service, it isn't started correctly.
```
sudo: sytemctl: command not found
```


openoms:
Nice find, thanks!

Issue #2502, opened 2021-08-27
Change bitcoind download source or add a dialog menu to choose source


nyxnor:
bitcoincore.org blocks UK clients, so:

1. Github: https://github.com/bitcoin/bitcoin/releases with signed tag:
![image](https://user-images.githubusercontent.com/69700936/131153050-0dee581c-a640-4abf-b83b-a55bdffca645.png)

2. bitcoincore.org onion binary source -> http://6hasakffvppilxgehrswmffqurlcjjjhd76jgvaqmsg6ul25s7t3rzyd.onion/bin

3. Or prefix with `torsocks` this https://github.com/rootzoll/raspiblitz/blob/79ccb8d3433a59ee8268b458436dbb4f3a824aa3/home.admin/config.scripts/bitcoin.update.sh#L122

4. Laanwj onion git -> http://nxshomzlgqmwfwhcnyvbznyrybh3gotlfgis7wkv7iur2yj2rarlhiad.onion/git/bitcoin.git Source -> https://laanwj.github.io/2018/06/08/tor-repository.html
-------------
Probably the project is more inclined to use the trusted signed binaries, so the options 1 and 2 are the most compatible with Blitz, but documenting here the git options 3 and 4.

rootzoll:
Yeah changing to github releases might be the easiest solution for now .. could be combined with the refactor to let the build script call the install scripts to prevent "code duplication".

openoms:
I'd support the fallback to download over Tor if both bitcoincore.org (as in #2527) and the github link fails.

PR #2527, opened 2021-09-11
fix Bitcoin Core download links for UK users

openoms:
Necessary changes to be able to download the Bitcoin Core binary from a UK IP address.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -875,7 +875,7 @@ fi
 gpg --import ./laanwj-releases.asc
 
 # download signed binary sha256 hash sum file and check
-sudo -u admin wget https://bitcoin.org/bin/bitcoin-core-${bitcoinVersion}/SHA256SUMS.asc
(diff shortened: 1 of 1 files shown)


openoms:
Fixes: https://github.com/rootzoll/raspiblitz/issues/2520

nyxnor:
closes https://github.com/rootzoll/raspiblitz/issues/2502 if bitcoincore does not become blocked.

openoms:
@nyxnor this indeed feels like a cat and mouse game. Downloading over tor could be good option when the clearnet download fails.

This PR is just the minimal necessary change, can be extended later.

nyxnor:
TorBox checks if Tor connections are working, if yes, tries over Tor first. 
Example https://github.com/radio24/TorBox/blob/7477e0032847db019c93eb22d338666451840711/update#L542 & https://github.com/radio24/TorBox/blob/master/lib/torbox.lib#L107

PR #2536, opened 2021-09-13
Tor: remove tor@lnd instance

openoms:
The separate tor instance for LND became redundant:
* now Tor is only reloaded, not restarted on config changes so the port remains open
* using LND streamisolation means every LND connection is on a new Tor circuit


rootzoll:
Is this one ready to merge?

openoms:
Yes, merging

Issue #1500, opened 2020-08-30
Raspiblitz installation on a VM


BlockHackTk:
Hello, 
I would like to ask for some help that I didn't get on the Raspiblitz Telegram.

I'm trying to run raspiblitz on a virtual machine. I just bought a Freebox V6 internet box on which you can run a VM in ARM! 
[Here is a link to a tutorial to install a Bitcoin Core Node ](https://otcpm24.com/2020/03/01/how-to-put-in-a-full-bitcoin-node-on-the-most-recent-freebox-delta/)

So I tried on a Debian Buster VM to run the wget command 
`https://raw.githubusercontent.com/rootzoll/raspiblitz/master/build_sdcard.sh && sudo bash build_sdcard.sh`

Everything goes well and I'm at the end of the installation, but I have a problem with the HDD. 
I have a message that it is not connected (the HDD). I have to connect a HDD in Usb for it to be recognized. 

The next problem is that the Network line of the config file is not filled and I have a problem. 
If I enter **Bitcoin and Main** in the file, I manage to move forward, but when I'm asked to download the Blockchain, I get a count from 0 to 200 and then it comes back to the start message to download the Blockchain. 

I have some screenshots that I can share with you. I'm going to start again and make a recording of what happens to share it with you. 

Maybe with your help I could make a tutorial to share with Freebox Delta S users. 

An internet box is always running, H24 7/7 I think it's a good opportunity to run a bitcoin node and get interested in bitcoin. 
Free is a french FAI (in french) or ISP (Internet Service Provider)

I hope I wouldn't be the only owner of this box to want to try the adventure? I have the basic knowledge of Linux that I've been using for more than 20 years.

ChuckNorrison:
This code should be interesting in blitz.datadrive.sh

```
# gathering system info
# is global so that also other parts of this script can use this

# basics
isMounted=$(sudo df | grep -c /mnt/hdd)
isBTRFS=$(sudo btrfs filesystem show 2>/dev/null| grep -c 'BLITZSTORAGE')
isRaid=$(btrfs filesystem df /mnt/hdd 2>/dev/null | grep -c "Data, RAID1")
isSSD=$(sudo cat /sys/block/sda/queue/rotational 2>/dev/null | grep -c 0)
```
I would recommend to try those calls manually and give some feedback on the results

adevoss:
@BlockHackTk Please try #1297
[...]

PR #2538, opened 2021-09-13
blitz.datadrive: detect and protect bootPartition

openoms:
Fixing the problem described in: https://github.com/rootzoll/raspiblitz/issues/1500#issuecomment-917444439

Makes sure that the boot partition does not become a `hddCandidate` and gets formatted during setup.

### DIFF
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -82,7 +82,13 @@ if [ "$1" = "status" ]; then
     # will then be used to offer formatting and permanent mounting
     hdd=""
     sizeDataPartition=0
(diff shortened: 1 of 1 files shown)


rootzoll:
Looks good to me ... I dont have an UEFI systems to test with ... but should make no problems on a basic RaspberryPi. So its OK to me to merge and test it as part of the RC2.

openoms:
Tested on RPi - both partitions correctly identiifed
Debian laptop  (LUKS encrypted and non-EFi) - both partitions correctly identiifed
Ubuntu VM - EFI - both OS and boot partittions were correctly identified
interestingly `df /boot` is on `vda2` and needs `df /boot/efi` to pick up `vda1`

Issue #2542, opened 2021-09-14
remove litecoin


nyxnor:
LOL, it is not being taking care of and it carried on, when it should not be.

rootzoll:
I also miss time to test it for new releases ... maybe like with other abandon features/services we should start to remove it from the SSH setup menus with v1.8.0 ... and if than somebody commits again to maintain it we can reconsider before removing it completly.

openoms:
ACK

(if we want to be bold and busy can consider adding Liquid instead)

nyxnor:
> (if we want to be bold and busy can consider adding Liquid instead)

Pls be a joke

openoms:
> > (if we want to be bold and busy can consider adding Liquid instead)
> 
> Pls be a joke

Not a principle question in mind rather a technical one. Litecoin got there because of the compatibilty with Lightning.
Anyway no usecase would warrant Liquid either.

jasenm69:
i want to build litecoin nodes for the litecoin fam. these are real ppl i know in the real world, good folks and they love ltc. please do not remove this i plan on adding volume here

ghost:
I second @jasenm69. Don't give up on Litecoin support. I sense that more than a few folks want to get on board with setting up LN nodes shortly. A lot of chatter trending on Twitter recently.

Will benefit everyone to continue supporting Litecoin. Thanks for considering.

nyxnor:
Can any of you commit to maintain the litecoin implementation?
> rootzoll: and if than somebody commits again to maintain it we can reconsider before removing it completly.

ghost:
Admittedly, no. I cannot. I don't want to burden anyone with upkeep if only a handful of folks are taking advantage. I was about to install this later in the week, but I'm only one person and I have no idea how to perform maintenance.

Is there a way to archive everything in case someone wanted to make that commitment in the future?

jasenm69:
Is there a way to grab whats build for litecoin . I'm a noob but ill pull together a team and dedicate to building it out.

Sent from ProtonMail mobile

-------- Original Message --------

jasenm69:
I'm in my only problem is I'm not very experienced. I know a solid crew of litecoin OG'S we can pick up the slack.

Sent from ProtonMail mobile

-------- Original Message --------

jasenm69:
Ove been chattering it up on Twitter to the litecoin family. However I can get a ltc lightning node built i want to.

Sent from ProtonMail mobile

[...]

PR #2545, opened 2021-09-14
removing litecoin option from the menu, code still there

nyxnor:
I just commented it as instructed to remove from the ssh menu https://github.com/rootzoll/raspiblitz/issues/2542#issuecomment-919190331


jasenm69:
please do not remove it i am building this out and i have many litecoin friends whom want in. for me its just fun but they are litecoiners and i respect them over the coin.

PR #2569, opened 2021-09-22
Bisq changed to tor onion v3

Emzy:
Bisq v1.7.4 will support v3 onions. Will be out in a week.
v2 onions will still work, but they getting generally more unreliable.


Emzy:
Tested it with Bisq release/v1.7.4 and it works.

rootzoll:
Will be part of v1.7.1rc3

PR #2570, opened 2021-09-23
Added timeout to blitz.subscriptions.py -> NEW1 menu option

blackcoffeexbt:
If the sphinxrelay service is not available (removed port forwarding ot other cause) selecting
    NEW1 from the subscriptions menu hangs the script. This commit adds a 10 second timeout to the
    bonus.sphinxrelay.sh subprocess call to prevent this.


blackcoffeexbt:
Will fork from dev for this request

PR #2590, opened 2021-10-02
Update Odroid_HC1_XU4.md

MichaIng:
- Add Odroid MC1 (another variant)
- Update image download links
- Fix dietpi.com links
- Update "getting started" link to point to a reasonable section in our new documentation
- Remove obsolete hassle with first run setup loop of the v6.14 image

### DIFF
--- a/alternative.platforms/dietpi/Odroid_HC1_XU4.md
+++ b/alternative.platforms/dietpi/Odroid_HC1_XU4.md
@@ -1,13 +1,13 @@
 
 # ⚡️ RaspiBlitz-on-DietPi ⚡️
-# For the Odroid HC1 / HC2 / XU3 / XU4
+# For the Odroid HC1 / HC2 / XU3 / XU4 / MC1
 
 ![](pictures/DroidBlitz.jpg)
 
 ## There are 3 options provided:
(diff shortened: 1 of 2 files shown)


openoms:
Thank for this.
I did not test DietPi for a while, but the changes look straightforward.

ACK

MichaIng:
Another thing is that the `dietpi.txt` should be updated. Actually a nice solution for this is to use a `Automation_Custom_PreScript.sh` script to only change the `dietpi.txt` to contain `AUTO_SETUP_AUTOMATED=1` and `AUTO_SETUP_INSTALL_SOFTWARE_ID=73`, so it doesn't require an updated whenever the `dietpi.txt` content changes in future DietPi versions.

The `Automation_Custom_PreScript.sh` runs at a very early boot stage, before any login ability is ready, so it can be used to adjust `dietpi.txt` itself and trigger automation. `Automation_Custom_Script.sh` runs after all first run installs have finished (as correctly stated), so is great the way it is used within these instructions.

rootzoll:
merged into dev for v1.7.1 release

MichaIng:
Great. I'll open another PR with the `Automation_Custom_PreScript.sh` suggestion.

PR #2605, opened 2021-10-07
Update README.md

sys-shock:
Adding detail about Tor response time with links.


rootzoll:
Thanks @sys-shock for adding this info ... I will merge it into the README but put it further down in the text to the SETTINGS chapter. I laso need to edit it, because during setup RaspiBlitz is TorByDefault since some time now - just after the setup the user can then deactivate Tor and run the node on clearnet.

Issue #2601, opened 2021-10-06
man blitz - manual pages for raspiblitz


nyxnor:
### help per script
normally inside scripts we already add a `-h` flag to show instructions about scripts, and I coded this to catch any `help` flag on POSIX:
```sh
script_usage{
  printf "this script does this thing"
}

{ [ -z "${1}" ] || [ "${1}" = "help" ] || [ "${1}" = "-h" ] || [ "${1}" = "-help" ] || [ "${1}" = "--help" ]; } && { script_usage; exit 1; }
```
For `bash` it is posible to use wildcard
```sh
{ [ -z "${1}" ] || [ "${1}" = "-h" ] || [ "${1}" = *help ] ; } && { script_usage; exit 1; }
```

But for the blitz environment overall, good documentation is important and it is handy to have it on the command line using man pages.

### How to use man:

terminal
```
man man
```
or read online [here](https://bash.cyberciti.biz/guide/Man_command)

### man to html

Man pages can also be converted to html as shown [here](https://linux.die.net/man/1/dialog) so it can be read on the browser also. I did not get to this point yet.

### markdown to man

I did not have luck with this tool yet:
* https://github.com/apjanke/ronn-ng

### build man

[How to build man pages](https://www.howtogeek.com/682871/how-to-create-a-man-page-on-linux/)
This one [man page I built](https://github.com/nyxnor/onionservice/blob/main/docs/onionservice-cli.1)
![image](https://user-images.githubusercontent.com/69700936/136174483-bbe238a2-30b4-485c-abec-e05ceb7ce60f.png)

### plan

* start with [_commands.sh](https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/_commands.sh)
* then progress from there explaining a survival guide in the blitz environment.
* this will not explain specific scripts, they should have a flag `help` inside them, but the man will describe some basics scripts functionalities, like for services normally there is `on`, `off`, `update`.

nyxnor:
Ok, it is possible to use markdown very easily too https://www.howtogeek.com/682871/how-to-create-a-man-page-on-linux/

nyxnor:
[tor.txt](https://github.com/rootzoll/raspiblitz/files/7303334/tor.txt)

see `tor` manual page.

```sh
sudo /usr/share/man/man1/tor.1.gz ~/
sudo gunzip tor.1.gz
nano tor.1
```

rootzoll:
Closing because now `blitzhelp` exists.

PR #2606, opened 2021-10-08
add manual page that can be called with 'man raspiblitz'

nyxnor:
See issue https://github.com/rootzoll/raspiblitz/issues/2601
The manual is written in Markdown format and converted to Man with pandoc.

can call the man as any other man page as it was added to mandb
```
man raspiblitz
```

Mans are for commands, but as blitz is just one of the commands, a single man page is better.

image
image

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -55,7 +55,7 @@ fi
 # 3rd optional parameter: GITHUB-USERNAME
 # ---------------------------------------
 # could be any valid github-user that has a fork of the raspiblitz repo - 'rootzoll' is default
-# The 'raspiblitz' repo of this user is used to provisioning sd card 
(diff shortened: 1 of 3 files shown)


rootzoll:
Because this needs another change of the sd card build ... moving to v1.8.0

nyxnor:
you can also just pre generate the manual page and place it on the `assets` folder. But on every system it needs to be moved to `/usr/local/man/man1/` and update the `mandb`. Can be outside of the build_sdcard.sh but I don't know where exactly.

rootzoll:
I think its good to have this as part of the v1.8.0 build script.

rootzoll:
I would consider this PR as outdated since there is the `blitzhelp` command now? @nyxnor do you agree?

nyxnor:
ok, it was just a manual format instead of just printed to the screen

PR #2638, opened 2021-10-17
Fix adduser error for Ubuntu

matheusbach:
As in Ubuntu there is already a group called admin, when the script tries to add a user called admin it returns ```adduser error: The group 'admin' already exists.```
By specifying the group the bug is fixed.


rootzoll:
@matheusbach can you base the PR on the `dev` branch, on which we prepare the next release?

matheusbach:
> @matheusbach can you base the PR on the `dev` branch, on which we prepare the next release?
Of course, done.

PR #2657, opened 2021-10-24
automatically create cronjob for kindle-display

### LINKED ISSUE #2616: Bonus: kindle-display - why not setup cronjob by default
As far as I see it the install script for the kindle-display only hints that a cronjob could be setup: https://github.com/rootzoll/raspiblitz/blob/v1.7/home.admin/config.scripts/bonus.kindle-display.sh#L167

Why don't we actually setup this by default?
I just upgraded from 1.7.0 to 1.7.1 and had to remember to set up the cronjob - otherwise my kindle wouldn't update.

@dennisreimann thoughts? ;-D

frennkie:
fix #2616

### DIFF
--- a/home.admin/config.scripts/bonus.kindle-display.sh
+++ b/home.admin/config.scripts/bonus.kindle-display.sh
@@ -151,6 +151,14 @@ EOF
       sudo -u $USERNAME $CRON_FILE
     fi
 
+    # set cronjob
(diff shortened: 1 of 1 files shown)


rootzoll:
@frennkie can you rebase the PR on the `dev` branch?

rootzoll:
OK merging into `v1.7merge` branch to then cherrypick over to `dev`.

PR #2670, opened 2021-10-27
improve go script

nyxnor:


openoms:
utACK 7cc3d8a

nyxnor:
what does `ut` in ack means?

openoms:
```
Concept ACK - Agree with the idea and overall direction, but haven't reviewed the code changes or tested them.

utACK (untested ACK) - Reviewed and agree with the code changes but haven't actually tested them.

Tested ACK - Reviewed the code changes and have verified the functionality or bug fix.

ACK - A loose ACK can be confusing. It's best to avoid them unless it's a documentation/comment only change in which case there is nothing to test/verify; therefore the tested/untested distinction is not there.

NACK - Disagree with the code changes/concept. Should be accompanied by an explanation.
```

https://github.com/bitcoin/bitcoin/issues/6100#issuecomment-98471846

nyxnor:
I am pushing posix compliant scripts to be able to work on any unix system, to try to dissipate heat from only using debian and bash.
* shebang to use environment instead of fixed path. See https://github.com/dylanaraps/pure-bash-bible#shebang
* use `shellcheck -s sh bonus.go.sh` to verify the syntax. https://github.com/koalaman/shellcheck
* /etc/profile is used on bourne shells, bash, ksh and ash, so it is okay
* if the installed version is different than the pinned variable, will install the pinned version.
* prefer case statements, more clear to read than a bunch of if-then-else

openoms:
> prefer case statements, more clear to read than a bunch of if-then-else

if statements are used in most scripts. Afaik using `case` does not improve anything and consistency across the codebase is preferred.

The most important question:
was the new script tested on a live RaspiBlitz?

nyxnor:
1. I think it is cleaner but changing a lot of scripts will take too much time to analyze, need to be something done over time. See the difference of readability here on the [current script](https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/config.scripts/bonus.go.sh) and the [modified one](https://github.com/nyxnor/raspiblitz/blob/go-dev/home.admin/config.scripts/bonus.go.sh).
2. tested on debian, will test on live Blitz soon.

openoms:
Fine by me, both works and the case solution is shorter.
One comment re readability:
the help message is useful to be first, because that's where one would look first.

nyxnor:
tested on a blitz now.

nyxnor:
can `GOPATH` be added to the `.bashrc` instead of `/etc/profile`? Would make things easier than sourcing `/etc/profile` everytime it needs to use go. Or at least source the profile from the bashrc.

```sh
[ -f /etc/profile ] && . /etc/profile
```

rootzoll:
Sourcing it is the safer I way I guess .. with .bashrc it would depend of the user running the script.

Issue #2669, opened 2021-10-26
Tor improvements


nyxnor:
### Reviewers
@openoms @rootzoll 

### Description
If someone wants to start testing, do it on the following order:
1. OnionService first because it will changes some scripts and variables name
1. Countermeasure is setting the correct path on all files
1. wait for Build Tor to be complete

### Overall Issues
* https://github.com/rootzoll/raspiblitz/pull/2419 is deprecated now, a lot of the methodology changed and it is now worth it anymore, the core of it, the goals are the same, but the code is not.
* https://github.com/rootzoll/raspiblitz/issues/2209

-------------

### PR:
* https://github.com/rootzoll/raspiblitz/pull/2672
* https://github.com/rootzoll/raspiblitz/pull/2711

Need to configure Pluggable Transports like in https://github.com/radio24/TorBox/blob/master/install/run_install.sh#L534-L667
Snowflake has to be built from source, obfs4 I had problems not building from source on arm64/aarch64, but TorBox is not building it, just installing the package so lets test.

I am developing this now, this is the hard part but stuck in the middle. https://www.youtube.com/watch?v=OMAIsqvTh7g

### Issues
* https://github.com/rootzoll/raspiblitz/issues/2596 This is to be faster and spend less data
* https://github.com/rootzoll/raspiblitz/issues/2054 This is why I started and needs to be solved to help anyone in need
* https://github.com/rootzoll/raspiblitz/issues/2217 I will make the onion service for debug be create by default

### This is to replace
https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/config.scripts/internet.tor.sh

### Code:
* https://github.com/rootzoll/raspiblitz/issues/2677

This is for censorship resistance, use of bridges is the focus.
Vanguards was included on OnionService so discarding it here.
The issue on https://github.com/rootzoll/raspiblitz/pull/2419 related to bridges was that I modified all the files manually to fit blitz paths and it became difficult to keep up with changes upstream to solve conflicts, new commits and update them. Because of this now I bulked this task with `sed` and `find`.
I see this being used in the following manner, use the script every blitz update to keep up with torbox updates, it places the files on blitz path, /home/admin and so on. It aims to be used like this:
```
sh setup.sh
mv -r raspiblitz/ /path/to/raspiblitz/cloned/repo/
git add -a
git commit -a
```

### Issues
* https://github.com/rootzoll/raspiblitz/issues/2054 depends on `build tor`

[...]

PR #2671, opened 2021-10-27
Tor build

nyxnor:
https://github.com/rootzoll/raspiblitz/issues/2669

Preparing tor to build with pluggable transports.
The sd card build still has to be improved, maybe move snowflake build to there? Still I am using the bonus.go.sh to build.


nyxnor:
christ, wrong branch to compare

PR #2672, opened 2021-10-27
[WIP] Tor build - start preparing tor to use bridges with pluggable transport

nyxnor:
* https://github.com/rootzoll/raspiblitz/issues/2669

Preparing tor to build with pluggable transports.
The sd card build still has to be improved, maybe move snowflake build to there? Still I am using the bonus.go.sh to build.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -55,7 +55,7 @@ fi
 # 3rd optional parameter: GITHUB-USERNAME
 # ---------------------------------------
 # could be any valid github-user that has a fork of the raspiblitz repo - 'rootzoll' is default
-# The 'raspiblitz' repo of this user is used to provisioning sd card 
+# The 'raspiblitz' repo of this user is used to provisioning sd card
 # with raspiblitz assets/scripts later on.
 # If this parameter is set also the branch needs to be given (see next parameter).
 githubUser="$3"
@@ -205,7 +205,7 @@ echo "- OK key added"
 
 echo "*** Adding Tor Sources to sources.list ***"
 torSourceListAvailable=$(sudo grep -c 'https://deb.torproject.org/torproject.org' /etc/apt/sources.list)
-echo "torSourceListAvailable=${torSourceListAvailable}"  
+echo "torSourceListAvailable=${torSourceListAvailable}"
 if [ ${torSourceListAvailable} -eq 0 ]; then
   echo "- adding TOR sources ..."
   if [ "${baseimage}" = "raspbian" ] || [ "${baseimage}" = "raspios_arm64" ] || [ "${baseimage}" = "armbian" ] || [ "${baseimage}" = "dietpi" ]; then
@@ -215,7 +215,7 @@ if [ ${torSourceListAvailable} -eq 0 ]; then
   elif [ "${baseimage}" = "ubuntu" ]; then
     echo "- using https://deb.torproject.org/torproject.org focal"
     echo "deb https://deb.torproject.org/torproject.org focal main" | sudo tee -a /etc/apt/sources.list
-    echo "deb-src https://deb.torproject.org/torproject.org focal main" | sudo tee -a /etc/apt/sources.list    
+    echo "deb-src https://deb.torproject.org/torproject.org focal main" | sudo tee -a /etc/apt/sources.list
   else
     echo "!!! FAIL: No Tor sources for os: ${baseimage}"
     exit 1
@@ -253,7 +253,7 @@ if [ "${baseimage}" = "raspbian" ] || [ "${baseimage}" = "dietpi" ] || \
     # remove unnecessary files
     sudo rm -rf /home/pi/MagPi
     # https://www.reddit.com/r/linux/comments/lbu0t1/microsoft_repo_installed_on_all_raspberry_pis/
-    sudo rm -f /etc/apt/sources.list.d/vscode.list 
+    sudo rm -f /etc/apt/sources.list.d/vscode.list
     sudo rm -f /etc/apt/trusted.gpg.d/microsoft.gpg
   fi
   if [ ! -f /etc/apt/sources.list.d/raspi.list ]; then
@@ -308,7 +308,7 @@ echo "*** PREPARE ${baseimage} ***"
 if [ "$(compgen -u | grep -c dietpi)" -gt 0 ];then
   echo "# Renaming dietpi user to pi"
   sudo usermod -l pi dietpi
-elif [ "$(compgen -u | grep -c pi)" -eq 0 ];then  
+elif [ "$(compgen -u | grep -c pi)" -eq 0 ];then
   echo "# Adding the user pi"
   sudo adduser --disabled-password --gecos "" pi
   sudo adduser pi sudo
@@ -320,7 +320,7 @@ if [ "${baseimage}" = "raspbian" ]||[ "${baseimage}" = "raspios_arm64" ]||\
 
   echo ""
   echo "*** PREPARE RASPBIAN ***"
-  sudo apt install -y raspi-config 
+  sudo apt install -y raspi-config
   # do memory split (16MB)
   sudo raspi-config nonint do_memory_split 16
   # set to wait until network is available on boot (0 seems to yes)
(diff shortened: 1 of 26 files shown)


build_sdcard.sh
```diff
@@ -115,136 +101,91 @@ fi
...
+echo
 
 # FIXING LOCALES
 # https://github.com/rootzoll/raspiblitz/issues/138
 # https://daker.me/2014/10/how-to-fix-perl-warning-setting-locale-failed-in-raspbian.html
 # https://stackoverflow.com/questions/38188762/generate-all-locales-in-a-docker-image
-if [ "${baseimage}" = "raspbian" ] || [ "${baseimage}" = "dietpi" ] || \
-   [ "${baseimage}" = "raspios_arm64" ]||[ "${baseimage}" = "debian_rpi64" ]; then
-  echo ""
+if [ "${baseimage}" = "Raspbian" ] || [ "${baseimage}" = "DietPi" ] || \
```
rootzoll, line 178: if baseimage strings changed (now starting with upper case) needs to be checked on all other scripts that might use baseimage later on

nyxnor: base image should respect the commands output of `lsb_release -si` for uniformity. Will check other scripts.

build_sdcard.sh
```diff
@@ -790,27 +707,13 @@ if [ "${fatpack}" == "true" ]; then
...
-  sudo apt-get install -y fbi
-  sudo apt-get install -y ssmtp
-  sudo apt-get install -y unclutter xterm python3-pyqt5
-  sudo apt-get install -y xfonts-terminus
-  sudo apt-get install -y nginx apache2-utils
-  sudo apt-get install -y nginx
-  sudo apt-get install -y python3-jinja2
-  sudo apt-get install -y socat
-  sudo apt-get install -y libatlas-base-dev
-  sudo apt-get install -y mariadb-server mariadb-client
```
rootzoll, line 805: just general - lets remove mariadb packages upfront because its a server running in background not just a tool ... nginx is fine because its a basic server requirement

home.admin/00settingsMenuBasics.sh
```diff
@@ -254,7 +254,7 @@ Please keep in mind that thru your LND node id & your previous IP history with y
 
   # change Tor
   anychange=1
-  sudo /home/admin/config.scripts/internet.tor.sh ${choice}
+  sudo /home/admin/config.scripts/tor.network.sh ${choice}
```
rootzoll, line 257: after merge check if any other scripts use the old tor script naming .. also quickcheck waiting PRs

home.admin/99connectMenu.sh
```diff
@@ -254,7 +254,7 @@ HiddenServicePort 8333 127.0.0.1:8333" | sudo tee -a /etc/tor/torrc
         clear
         echo "# Make sure the bitcoind wallet is on"
         /home/admin/config.scripts/network.wallet.sh on
-        /home/admin/config.scripts/internet.hiddenservice.sh bitcoin${BITCOINRPCPORT} ${BITCOINRPCPORT} ${BITCOINRPCPORT}
```
rootzoll, line 257: same check after merge with this change of script name

home.admin/config.scripts/tor.network.sh
```diff
@@ -154,11 +127,38 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
...
+  sudo chmod a+x /usr/share/tor/geoip*
+  # Copy not moving!
+  (sudo cp /usr/share/tor/geoip* /usr/bin) 2> /dev/null
+  sudo setcap 'cap_net_bind_service=+ep' /usr/bin/obfs4proxy
+  sudo sed -i "s/^NoNewPrivileges=yes/NoNewPrivileges=no/g" /lib/systemd/system/tor@default.service
+  sudo sed -i "s/^NoNewPrivileges=yes/NoNewPrivileges=no/g" /lib/systemd/system/tor@.service
+
+  ## Install Snowflake
+  sudo rm -rf ~/downloads/snowflake
+  git clone https://github.com/keroserene/snowflake.git ~/downloads/snowflake
```
rootzoll, line 148: installs from external git repos should be fixed to a commit or release

home.admin/config.scripts/tor.network.sh
```diff
@@ -238,45 +241,26 @@ EOF
...
-    /home/admin/config.scripts/internet.hiddenservice.sh lnbits 80 5002 443 5003
-  fi
-  if [ "${thunderhub}" = "on" ]; then
-    /home/admin/config.scripts/internet.hiddenservice.sh thunderhub 80 3012 443 3013
-  fi
-  if [ "${specter}" = "on" ]; then
-    # specter makes only sense to be served over https
-    /home/admin/config.scripts/internet.hiddenservice.sh specter 443 25441
-  fi
+  [ "${BTCRPCexplorer}" = "on" ] && /home/admin/config.scripts/tor.onion-service.sh btc-rpc-explorer 80 3022 443 3023
```
rootzoll, line 258: general info: basically those apps should check by themselves before start (with systemd pre-start script) if Tor is now on and they have a hidden service. So when the user turnes Tor on/off a restart of the system is required (when called after setup) and then all apps will reconfigure themselves.

rootzoll: na scratch that ... pre-start scripts can only run as the app-unix-user that is not root and so cannot change Tor configs ... so pre-start is just fine for app-related configs, not system configs. So no change on this place at the moment.

home.admin/config.scripts/tor.network.sh
nyxnor: https://github.com/rootzoll/raspiblitz/pull/2672/commits/6110a76cc99fcea43dc1b84d0b32093dd78184f5

home.admin/config.scripts/tor.network.sh
nyxnor: could you inform where please?

home.admin/config.scripts/tor.network.sh
```diff
@@ -315,22 +299,17 @@ if [ "$1" = "0" ] || [ "$1" = "off" ]; then
...
-  echo ""
+  # deactivate bitcoin over tor (function call)
+  deactivateBitcoinOverTor
+  echo
 
   sudo /home/admin/config.scripts/internet.sh update-publicip
-  
+
   if [ "${lightning}" == "lnd" ] || [ "${lnd}" == "on" ] || [ "${lnd}" == "1" ]; then
     echo "# *** Removing Tor from LND Mainnet ***"
```
rootzoll, line 321: but this can be something that could be done by an pre-start script of lnd (also bitcoin above) .. so then at this place just an restart of the systemd service is needed to let the services update their config themselves

home.admin/config.scripts/tor.network.sh
```diff
@@ -353,16 +332,16 @@ if [ "$1" = "0" ] || [ "$1" = "off" ]; then
...
   sudo systemctl stop tor@default
-  echo ""
+  echo
 
   if [ "$2" == "clear" ]; then
-      echo "# *** Deinstall Tor & Delete Data ***"
+      echo "# *** Uninstall Tor & Delete Data ***"
       sudo rm -r /mnt/hdd/tor 2>/dev/null
-      sudo apt remove tor tor-arm -y
+      sudo apt remove tor nyx -y
```
rootzoll, line 351: so we are not removing tor in general here and keep having it run in the background?

build_sdcard.sh
nyxnor: https://github.com/rootzoll/raspiblitz/pull/2672/commits/90ac78db51e68cbf829dfdfb00609cc12371b305

nyxnor:
> rootzoll said: so we are not removing tor in general here and keep having it run in the background?

I don't know, from the last talk, you said it should be in there, but if you want, I can call `tor.network.sh` after cloning the raspiblitz repo and starting tor by default with a unified script, instead of duplicating things to the `build_sdcard.sh`

home.admin/config.scripts/tor.network.sh
rootzoll: will show later in-person

rootzoll:
deleted last comment about "removing tor" - overlooked a part in code - all good on that

rootzoll:
OK @nyxnor we merged the PR to the 2669tor branch. Naxt steps would be:

- test building an sd card with that branch: `wget --no-cache https://raw.githubusercontent.com/rootzoll/raspiblitz/dev/build_sdcard.sh && sudo bash build_sdcard.sh` .. se if that runs thru and doesned break setup of a new node or recovery
- fork the torbox repo and remove all not needed scripts
- add a `internet.torbox.sh` script that is able to install torbox
- add an option to sd card build script to add torbox (8th parameter)

nyxnor:
`tor.network.sh` is called on `_provision.sh` which is called by `bootsrap.sh` on fresh sd card. This means I will unify tor config to be only on `tor.network.sh` to avoid code duplication.
Is there a reason for tor package to still be in the `build_sdcard.sh`? As explained earlier, pretyflaco flashing the image containing tor was not making it work anyway and duplicating code can make one obsolete or prone to errors.

About the points above, understood.

rootzoll:
> config to be only on tor.network.sh to avoid code duplication. Is there a reason for tor package to still be in the build_sdcard.sh?

As you did in your changes its ok to have the install in a seperate script now.

PR #2687, opened 2021-10-30
Remove TestFlight label from Zeus

kaloudis:


rootzoll:
cherry-picked also into dev

PR #2690, opened 2021-11-01
Fix RESET-LND

### LINKED ISSUE #2685: RESET-LND doesn't create a new wallet
Selecting RESET-LND asked me to setup a new alias, but that was it, it didn't walk me through setting up a new wallet or anything. It seems like all it did was wipe the old lnd and set a new alias.

It looks like the RESET-LND option is supposed to run `/home/admin/70initLND.sh` but that script doesn't even seem to exist.

openoms:
fixes #2685

### DIFF
--- a/home.admin/98repairMenu.sh
+++ b/home.admin/98repairMenu.sh
@@ -124,7 +124,10 @@ case $CHOICE in
     echo "stopping lnd ..."
     sudo systemctl stop lnd
     sudo rm -r /mnt/hdd/lnd
+    # create wallet
     /home/admin/config.scripts/lnd.install.sh on mainnet initwallet
+    # display seed for mainnet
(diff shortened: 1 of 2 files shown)


openoms:
@rootzoll currently the newly created LND seedwords are stored on the disk.

Shall we create a menu option to display and offer to delete the seed in the LND menu or should it be just safely deleted straight after displayed the first time?

openoms:
> Shall we create a menu option to display and offer to delete the seed in the LND menu or should it be just safely deleted straight after displayed the first time?

The outcome here is that the seed will be safely deleted after having been displayed, the same as during the initial setup: fee0345

rootzoll:
also cherry-picked over into dev branch

Issue #2662, opened 2021-10-25
C-lightning update to v0.10.2


openoms:
the update to v0.10.2rc1 finishes successfully, but there is a pip error in the beginning:
```
Cloning into 'lightning'...
remote: Enumerating objects: 81588, done.
remote: Counting objects: 100% (3701/3701), done.
remote: Compressing objects: 100% (1618/1618), done.
remote: Total 81588 (delta 2322), reused 3023 (delta 2026), pack-reused 77887
Receiving objects: 100% (81588/81588), 44.92 MiB | 1.90 MiB/s, done.
Resolving deltas: 100% (61689/61689), done.
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-9se81v11/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-9se81v11/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-req-build-9se81v11/pip-egg-info
         cwd: /tmp/pip-req-build-9se81v11/
    Complete output (27 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-9se81v11/setup.py", line 12, in <module>
        setup(name='pyln-client',
      File "/usr/lib/python3/dist-packages/setuptools/__init__.py", line 144, in setup
        return distutils.core.setup(**attrs)
      File "/usr/lib/python3.8/distutils/core.py", line 108, in setup
        _setup_distribution = dist = klass(attrs)
      File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 447, in __init__
        _Distribution.__init__(self, {
      File "/usr/lib/python3.8/distutils/dist.py", line 292, in __init__
        self.finalize_options()
      File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 740, in finalize_options
        ep.load()(self)
      File "/usr/lib/python3/dist-packages/setuptools/dist.py", line 747, in _finalize_setup_keywords
        ep.load()(self, ep.name, value)
[...]

PR #2707, opened 2021-11-08
Do not offer updates to release candidates for LND and CL

### LINKED ISSUE #2706: do not offer updates to release canditates for LND and CL
Currently if the `reckless` update method was used previously, during recovery there is always the latest untested version is getting installed.

Should not go to a higher `reckless` version only to `stable` updates if that is higher.

openoms:
fix: #2706


openoms:
also updates CL #2662

Issue #2620, opened 2021-10-13
Opt out of telemetry for Thunderhub


dipunm:
During installation, there is a new message from next.js which I think we are not addressing at the moment.

We can remove telemetary as we have done for RTL by running the command shown here:
https://nextjs.org/telemetry#how-do-i-opt-out
```
npx next telemetry disable
```

rootzoll:
This merged with the PR from openoms:
https://github.com/rootzoll/raspiblitz/blob/217c1766d2f46bc38c4cfbcf48dde84a928a5eec/home.admin/config.scripts/bonus.thunderhub.sh#L98

@dipunm is your PR #2628 now a duplicate or an needed addition?

dipunm:
I suppose it is now a duplicate, although I prefer the env variable method because it is easy to see and understand and is permanent whereas the npx command could be undone later with no trace.

I'll do the rebase now, I've given my thoughts, I'll let you make the final decision. It's no harm to also add the environment variable.

rootzoll:
This looks good .. closing for v1.7.2 release

Issue #2686, opened 2021-10-30
[Security] git verify-commit everywhere


openoms:
Introducing a script https://github.com/rootzoll/raspiblitz/pull/2683/commits/d51440216436acbf82f16e4728f29b3ea0ec1fca  to verify the checked out git commits with the BTCpay update: https://github.com/rootzoll/raspiblitz/pull/2683

All scripts installing services built from the source code can have a header like:
```
PGPsigner="nicolasdorier"
PGPpubkeyLink="https://keybase.io/nicolasdorier/pgp_keys.asc"
PGPpubkeyFingerprint="AB4CFA9895ACA0DBE27F6B346618763EF09186FE"
```

and can run:
```
# pin version
sudo -u $installingUser git reset --hard $version
# PGP verify
sudo -u $installingUser /home/admin/config.scripts/blitz.git-verify.sh "${PGPsigner}" "${PGPpubkeyLink}" "${PGPpubkeyFingerprint}" || exit 1
```
       
to reduce trust in the download source.

Failing to verify the PGP signature on the checked commit breaks the installation.

Can implement this on all services as we go on applying the pinned updates.

nyxnor:
YES !!!
If possible, add to SECURITY.md

Kixunil:
Note that you can run `gpg --recv-keys FINGERPRINT` and so don't need to specify URL.

Also be mindful of downgrade attacks (an attacker could send a valid old signed version with known vulnerability so that he can then exploit it).

openoms:
> Note that you can run `gpg --recv-keys FINGERPRINT` and so don't need to specify URL.

yes, but some keys are not shared on the keyservers unfortunately. I find it best when people share it on their own website and then can be double checked in person, on a keyserver and maybe having some verified signatures too from other known entities ("circle of trust").

Kixunil:
So far all keys I use are on keyserver (Ubuntu), if there are any missing it may be nicer to ask people to share them.
One only needs to check fingerprint, no need to check whole key.

rootzoll:
PR merged for final testing

rootzoll:
Works so far good with RCs  .. closing foir release.

PR #2683, opened 2021-10-29
btcpay update to v1.3.3, NBXplorer to v2.2.16 + update option

openoms:
fix #2682

PR #2708, opened 2021-11-09
Git verify everywhere + updates

openoms:
Builds on https://github.com/rootzoll/raspiblitz/pull/2683

Discussed in: https://github.com/rootzoll/raspiblitz/issues/2686

Multiple updates included in CHANGES.md


openoms:
The current state of this PR is stable and tested, but there a few more services I am planning to use the `blitz.git-verify.sh` with.

Also will add to SECURITY.md

openoms:
Ready to be merged. 
The missing verification on the spinx-relay and CL (when installed from source) are noted in the relevant issues

rootzoll:
looks good to me :) But PR is against `v1.7`branch - please change base to `dev` to be merged

openoms:
@rootzoll now this syncs the documentation changes in v1.7 to the dev branch too.

openoms:
fixes #2620

openoms:
@rootzoll can this be merged so testing from dev can go on?

home.admin/config.scripts/blitz.git-verify.sh
```diff
@@ -0,0 +1,84 @@
...
+gpg --import --import-options show-only ./pgp_keys.asc
+fingerprint=$(gpg pgp_keys.asc 2>/dev/null | grep "${PGPpubkeyFingerprint}" -c)
+if [ "${fingerprint}" -lt 1 ]; then
+  echo
+  echo "# !!! WARNING --> the PGP fingerprint is not as expected for ${PGPsigner}"
+  echo "# Should contain PGP: ${PGPpubkeyFingerprint}"
+  source /home/admin/raspiblitz.info
+  if [ "${state}" == "ready" ]; then
+    echo "# Type: 'Accept risk' and press ENTER to TAKE THE RISK if you think all is OK"
+    read -r confirmation
```
rootzoll: Always keep in mind that more and more installation is done in the background (either because of WebUI or during recover process) and the scripts cannot interact with the user. Those `read` commands would just stop the background processes.

rootzoll: A test for "state=ready" is good ... but getting closer to v1.8 install scripts will also be called from WebUI which runs those in background.

openoms: ok, will rather give it an exit code to fail. Do we have any logic about what number it should exit with?

openoms: ok, gave it exit 7 and some output to STDERR in f547854

rootzoll: There is no system wide logic for exit codes. Just pick one.

rootzoll (approved):
OK looks good now.

rootzoll:
OK merged for final testing

home.admin/config.scripts/bonus.electrs.sh
```diff
@@ -14,6 +14,10 @@ if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
  exit 1
 fi
 
+PGPsigner="romanz"
+PGPpubkeyLink="https://github.com/${PGPsigner}.gpg"
+PGPpubkeyFingerprint="87CAE5FA46917CBB"
```
Kixunil, line 19: This looks like the short version - it'd be much safer to use the long version. Same issue is present in several other cases below.

home.admin/config.scripts/blitz.git-verify.sh
```diff
@@ -0,0 +1,77 @@
...
+# sudo -u btcrpcexplorer /home/admin/config.scripts/blitz.git-verify.sh \
+#  "${PGPsigner}" "${PGPpubkeyLink}" "${PGPpubkeyFingerprint}" || exit 1
+
+PGPsigner="$1"
+PGPpubkeyLink="$2"
+PGPpubkeyFingerprint="$3"
+
+wget -O pgp_keys.asc "${PGPpubkeyLink}"
+gpg --import --import-options show-only ./pgp_keys.asc
+fingerprint=$(gpg pgp_keys.asc 2>/dev/null | grep "${PGPpubkeyFingerprint}" -c)
```
Kixunil, line 35: Sadly, this is not actually secure. The attacker can:

1. Create a key with name equal to the PGP fingerprint of the developer
2. Compromise the server
3. Put concatenation of his key and the developer key on the server
4. Sign compromised source with his key

And this script will accept his source:

1. This check here will pass because the key is present - both keys are imported
2. The `git verify-*` will succeed
3. `grep ${PGPpubkeyFingerprint}` will succeed

I really, really recommend studying [the vulnerabilities in LND verification script](https://github.com/lightningnetwork/lnd/pull/5048) to understand how tricky GPG verification is. If you don't have time for it, the key takeaways are:

* Don't assume asc file contains single key
* Don't assume `grep` works with any command - don't use textual matching
* Export expected key into a temporary GPG keyring and use that one for verification

Kixunil (commented):
While better than nothing, I think this should be improved a bit.

home.admin/config.scripts/blitz.git-verify.sh
rootzoll: @openoms will you open issue/PR for further improvements?

PR #2711, opened 2021-11-09
[WIP] 2669tor - improve tor configuration

nyxnor:
* tor.network.sh > relation of tor with bitcoin and lightning implementations
* tor.install.sh and tor.update.sh > install and update tor
* build_sdcard.sh > remove tor from there and lil improvements

It includes `torrc.d`, which faciliates managing sections of torrc separately. It has to be improved to work with `tor.onion-service.sh`.

**Don't merge, it is not ready, just to start seeing the changes**


rootzoll:
@nyxnor So you are removing the tor install section here from the sd build script ... you will still add a change that will call now the `tor.install.sh` script, correct?

@openoms this changed sd card build script uses the new bullseyes raspiOS .. do you see any problems with going with bulllseye for the v1.7.2 release?

nyxnor:
>@nyxnor So you are removing the tor install section here from the sd build script ... you will still add a change that will call now the tor.install.sh script, correct?

It is called by [_provision.sh](https://github.com/nyxnor/raspiblitz/blob/3d14e81767553e09eaa004d2ef67071e1e7171c0/home.admin/_provision_.sh#L392) already. What was happening is that it was only configuring the the bitcoin.conf and lnd.conf, as tor was already installed from the build.

nyxnor:
also, how many days til 1.7.2? I will return to code only next tuesday.

rootzoll:
1.7.2 is roughly targeted for end of November .. no hurry.

_provision.sh is called quite late in the install process ... when the basic install of Tor will just be in _provision.sh, that would mean that a user would have not an sd card with running tor from the beginning (regarding the internet calls during the setup).. wasnt that the idea of adding the torbox like scripts with having an option parameter on the sd card build?

openoms:
> @openoms this changed sd card build script uses the new bullseyes raspiOS .. do you see any problems with going with bulllseye for the v1.7.2 release?

Not aware of any problems yet since I did not try. Would suggest to build a test image early with the new base image to catch any issues.

@nyxnor rplease avoid changing (especially relaxing) the permissions on many script files.

revert and use:
`git config core.fileMode false`
to avoid your local permission settings.

nyxnor:
> _provision.sh is called quite late in the install process ... when the basic install of Tor will just be in _provision.sh, that would mean that a user would have not an sd card with running tor from the beginning (regarding the internet calls during the setup).. wasnt that the idea of adding the torbox like scripts with having an option parameter on the sd card build?

Can call the `/home/admin/config.scripts/tor.network.sh` from the sd card to avoid code duplication. About including torbox, that is the plan, but is is also a lot of new scripts just for bridges, which I prefer to do after this PR.

> @nyxnor rplease avoid changing (especially relaxing) the permissions on many script files.

I thought that every config script and admin script should be executable. Will revert https://github.com/rootzoll/raspiblitz/pull/2711/commits/3d14e81767553e09eaa004d2ef67071e1e7171c0

rootzoll:
> Can call the /home/admin/config.scripts/tor.network.sh from the sd card to avoid code duplication. About including torbox, that is the plan, but is is also a lot of new scripts just for bridges, which I prefer to do after this PR.

Ah OK got it ... so this is the step to clean up and then a second step will add the future option. np.

rootzoll:
@nyxnor can you signal, when this PR is no longer in WIP (Work in Progress) and can be merged? Because other PRs also want to add changes to the sd card build script and I like to merge this one first.

nyxnor:
Yes, I can signal when probably end of next week.

nyxnor:
Ok, I am back. @openoms What is the criteria for the blitz files permissions? I don't understand why this difference.
```
$ stat -c %a bonus.lit.sh 
644
$ stat -c %a bonus.lnbits.sh 
755
```
Also, with executable files, it is not necessary to specify the shell before the script name, this occurs inside some scripts but not others.


See https://github.com/rootzoll/raspiblitz/blob/a0fd288e553678d0816d682e968a2eea0d165a9b/build_sdcard.sh#L615
Some `.sh` files on /home/admin are not executable when cloned from github, but they are after the build, so I guess it does not differentiate.

image

https://github.com/rootzoll/raspiblitz/blob/a0fd288e553678d0816d682e968a2eea0d165a9b/build_sdcard.sh#L618

image

https://github.com/rootzoll/raspiblitz/blob/a0fd288e553678d0816d682e968a2eea0d165a9b/build_sdcard.sh#L619

image

All `.sh` files that are 644 can be found with: `ls -lR | grep "\-rw\-r\-\-r\-\-.*.sh$"`

image

nyxnor:
I am having problems on the build when trying to include multiple torrcs with `%include /etc/tor/torrc.d`.
On raspberry pi, it is not even starting tor, on my debian, it starts but fails to reload. See [issue](https://gitlab.torproject.org/tpo/core/tor/-/issues/40514#note_2761856).

Forget it. it is working on the pi, but not on my debian.

Edit:
```sh
ssh pi@ipaddress ## password=raspberry
curl --tlsv1.3 --proto =https --location --remote-name https://github.com/nyxnor/raspiblitz/raw/2669tor/build_sdcard.sh
sudo bash build_sdcard.sh false false nyxnor 2669tor headless true
```

nyxnor:
fail_count=3

## python version
because bullseye does not have python 3.8 or older, only 3.9.
```
*** Python DEFAULT libs & dependencies ***
!!! FAIL !!!
There is no tested version of python present
```
```
pi@raspberrypi:~ $ ls /usr/bin/python3.9
/usr/bin/python3.9
```

## keyserver failure (2 times)
just to add: `Unable to locate package sqlite`, there is only `sqlite3` on bullseye.
```
*** PREPARING BITCOIN ***
gpg: directory '/root/.gnupg' created
gpg: keybox '/root/.gnupg/pubring.kbx' created
gpg: keyserver receive failed: Server indicated a failure
!!! FAIL !!! Couldn't download Wladimir J. van der Laan's PGP pubkey
```


## Build succeeded. Now, proceeding to the tor scripts

nyxnor:
```
[warn] Error creating directory /mnd/hdd/tor/sys/keys: Read-only file system
```
`/mnt/hdd/tor` was not created as it was mounted read-only when the script was executed... stuck here. Already rebooted, still read-only. **Help wanted**
Just ran `tor.install.sh` manually:

I also used `~/` for home, which ended up being `root`, not `admin` for snowflak download, need to fix.
```
Failed to set capabilities on file `/usr/bin/obfs4proxy' (No such file or directory)
The value of the capability argument is not permitted for a file. Or the file is not a regular (non-symlink) file
Cloning into '/root/downloads/snowflake'...
```

nyxnor:
```
Configuration file '/etc/tor/torrc'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** torrc (Y/I/N/O/D/Z) [default=N] ? dpkg: error processing package tor (--configure):
 end of file on stdin at conffile prompt
Setting up nyx (2.1.0-2.1) ...
Setting up torsocks (2.3.0-3) ...
dpkg: dependency problems prevent configuration of tor-geoipdb:
 tor-geoipdb depends on tor (>= 0.4.5.10-1~deb11u1); however:
  Package tor is not configured yet.

dpkg: error processing package tor-geoipdb (--configure):
 dependency problems - leaving unconfigured
Processing triggers for man-db (2.9.4-2) ...
Errors were encountered while processing:
 tor
 tor-geoipdb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

fix with [this](https://askubuntu.com/a/104912):
```sh
sudo apt -o Dpkg::Options::="--force-confold" install -y tor
```

nyxnor:
the read-only file system error is on the build. just need to find where.

nyxnor:
Build sdcard is okay, failed to mount drive on Virtualbox

comment moved to https://github.com/rootzoll/raspiblitz/issues/2756#issue-1068175303

openoms:
is sr0 the second drive? never seen that. It is just because `blitz.datadrive.sh` does not grep for those letters. We can add it, can you test?

can restart the setup by deleting the:
`raspiblitz.info
`provision.flag
and reboot.

Passwords will remain as set (and will be asked again).

nyxnor:
comment moved to https://github.com/rootzoll/raspiblitz/issues/2756#issuecomment-983459366

openoms:
@nyxnor mounting drives is not relevant to the Tor config, can you move to a new issue?
Mount a second disk in the VM rather than multiple partitions.

nyxnor:
Ok, did the tor install but not the build on another machine, tor is fine, just need to test with build later.

also change `sudo systemctl enable tor@default` to `sudo systemctl enable tor@ tor@.service` if that makes sense
https://github.com/nyxnor/raspiblitz/blob/e56c03bf4e41fe65c7cf34b04b57936c166ab460/home.admin/config.scripts/tor.install.sh#L201

christ. typo. `s/mnd/mnt`
```
*** Create directories and set permissions ***
mkdir: created directory '/etc/tor/torrc.d'
mkdir: created directory '/mnd'
mkdir: created directory '/mnd/hdd'
mkdir: created directory '/mnd/hdd/tor'
mkdir: created directory '/mnd/hdd/tor/sys'
mkdir: created directory '/mnd/hdd/tor/sys/keys'
mkdir: created directory '/mnd/hdd/tor/services'
mkdir: created directory '/mnd/hdd/tor/onion_auth'
mode of '/mnd/hdd/tor' changed from 0755 (rwxr-xr-x) to 0700 (rwx------)
mode of '/etc/tor' retained as 0755 (rwxr-xr-x)
mode of '/etc/tor/torrc.d' retained as 0755 (rwxr-xr-x)
mode of '/etc/tor/torrc' retained as 0644 (rw-r--r--)
chmod: cannot access '/etc/tor/torrc.d/*': No such file or directory
failed to change mode of '/etc/tor/torrc.d/*' from 0644 (rw-r--r--) to 0644 (rw-r--r--)
changed ownership of '/mnd/hdd/tor/services' from root:root to debian-tor:debian-tor
changed ownership of '/mnd/hdd/tor/onion_auth' from root:root to debian-tor:debian-tor
changed ownership of '/mnd/hdd/tor/sys/keys' from root:root to debian-tor:debian-tor
changed ownership of '/mnd/hdd/tor/sys' from root:root to debian-tor:debian-tor
changed ownership of '/mnd/hdd/tor' from root:root to debian-tor:debian-tor
ownership of '/etc/tor/torrc' retained as root:root
ownership of '/etc/tor/torsocks.conf' retained as root:root
ownership of '/etc/tor/torrc.d' retained as root:root
ownership of '/etc/tor' retained as root:root
```

nyxnor:
ok, todo:
/etc/tor/torrc.d/ just will be overwritten if empty, blitz will always overwrite the /etc/tor/torrc if it does not contain an anchor such as `raspiblitz torrc`
maye include the torrc.d ffrom inside the /mnt/hdd to save the config

build_sdcard.sh
```diff
@@ -38,147 +33,113 @@ echo "1) will use NO-INTERACTION --> '${noInteraction}'"
...
 
 # 4th optional parameter: GITHUB-BRANCH
 # -------------------------------------
-# could be any valid branch of the given GITHUB-USERNAME forked raspiblitz repo - take ${defaultBranchVersion} is default
-githubBranch="$4"
-[ -z "${githubBranch}" ] && githubBranch="${defaultBranchVersion}"
-echo "4) will use GITHUB-BRANCH --> '${githubBranch}'"
+# could be any valid branch of the given GITHUB-USERNAME forked raspiblitz repo
+# https://github.com/rootzoll/raspiblitz/tags
+githubBranch="${4:-v1.7.1}"
```
rootzoll: should be kept as var on top and also there is no v1.7.1 branch - it should stay a branch not a tag

build_sdcard.sh
```diff
@@ -38,147 +33,113 @@ echo "1) will use NO-INTERACTION --> '${noInteraction}'"
...
+[ "${modeWifi}" = "true" ] && modeWifi="US"
+echo "7) WIFI --> '${modeWifi}'"
 
 # AUTO-DETECTION: CPU-ARCHITECTURE
 # ---------------------------------------
 # keep in mind that DietPi for Raspberry is also a stripped down Raspbian
 cpu="$(uname -m)"
 architecture="$(dpkg --print-architecture)"
-if [ "${cpu}" != "arm" ] && [ "${cpu}" != "aarch64" ] && [ "${cpu}" != "x86_64" ]; then
+if [ "${cpu}" != "arm*" ] && [ "${cpu}" != "aarch64" ] && [ "${cpu}" != "x86_64" ]; then
```
rootzoll: quick to check if a star notation is working here

home.admin/config.scripts/tor.network.sh
```diff
@@ -59,218 +63,83 @@ activateBitcoinOverTor()
...
-  [ "${thunderhub}" = "on" ] && /home/admin/config.scripts/tor.onion-service.sh thunderhub 80 3012 443 3013
-  [ "${specter}" = "on" ] && /home/admin/config.scripts/tor.onion-service.sh specter 443 25441
-  if [ "${sphinxrelay}" = "on" ]; then
-    /home/admin/config.scripts/tor.onion-service.sh sphinxrelay 80 3302 443 3303
-    toraddress=$(sudo cat /mnt/hdd/tor/sphinxrelay/hostname 2>/dev/null)
-    sudo -u sphinxrelay bash -c "echo '${toraddress}' > /home/sphinxrelay/sphinx-relay/dist/toraddress.txt"
-  fi
+    # ACTIVATE APPS OVER TOR
+    . /mnt/hdd/raspiblitz.conf 2>/dev/null
+    /home/admin/config.scripts/tor.onion-service.sh web80 80 80 443 9999
```
rootzoll: 443 not needed

home.admin/config.scripts/tor.network.sh
```diff
@@ -59,218 +63,83 @@ activateBitcoinOverTor()
...
+    /home/admin/config.scripts/tor.onion-service.sh debuglogs 80 6969
+    [ "${BTCRPCexplorer}" = "on" ] && /home/admin/config.scripts/tor.onion-service.sh btc-rpc-explorer 80 3022 443 3023
+    [ "${rtlWebinterface}" = "on" ] && /home/admin/config.scripts/tor.onion-service.sh RTL 80 3002 443 3003
+    [ "${BTCPayServer}" = "on" ] && /home/admin/config.scripts/tor.onion-service.sh btcpay 80 23002 443 23003
```
rootzoll, line 129: We leave thsi for now ... but will get removed for v1.8 with with issue #2545 2780

home.admin/config.scripts/tor.update.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
rootzoll: Would make sense to have this as part of the tor.install.sh script

build_sdcard.sh
nyxnor: done, just to leave the comment, it can server a branch or a tag.

home.admin/config.scripts/tor.update.sh
nyxnor: done

build_sdcard.sh
nyxnor: it does not work, changed that to case

rootzoll:
OK merging into`2669tor` to then rebase with`dev` branch - clear up conficts and to make new PR against dev.

PR #2731, opened 2021-11-26
Add squeaknode app

yzernik:
Add the squeaknode app: https://github.com/yzernik/squeaknode

I have tested this on my raspiBlitz, and it seems to be working.

One issue that i'm not sure about is the `zmqpubhashblock` config. It is required for squeaknode app, so I added it to `bitcoin.install.sh`, but I'm not sure if the config is being applied.

PR #2732, opened 2021-11-27
Add squeaknode app

yzernik:
This PR replaces: https://github.com/rootzoll/raspiblitz/pull/2731

Add the squeaknode app: https://github.com/yzernik/squeaknode

I have tested this on my raspiBlitz, and it seems to be working.

One issue that i'm not sure about is the zmqpubhashblock config. It is required for squeaknode app, so I added it to bitcoin.install.sh, but I'm not sure if the config is being applied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,10 +35,10 @@ There are further Services that can be switched on:
 * **Telegraf metrics** [details](https://github.com/rootzoll/raspiblitz/issues/1369)
 * **Chantools** (Fund Rescue) [details](https://github.com/guggero/chantools/blob/master/README.md)
 * **Suez** (Channel Visualization for LND & CL) [details](https://github.com/prusnak/suez#suez)
+* **Squeaknode** [details](https://github.com/yzernik/squeaknode)
 * **CL Spark Wallet** (WalletUI with BOLT12 offers) [details](https://github.com/shesek/spark-wallet#progressive-web-app)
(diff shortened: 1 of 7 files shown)


rootzoll:
Great work - this is looks already like a great integration. Marking to be included for the v1.7.2 release. Needs final testing. @yzernik I thin kthe last thing missing would be to add a call to your install script in `_provision_.sh` if the ` `squeaknode=on` parameter is set in the `raspiblitz.conf` so that it gets reinstalled on updates & recovers.

@openoms can you take a quick look at the zmqpubhashblock config added in the `bitcoin.install.sh` script?

yzernik:
Thanks @rootzoll !

I just added a squeaknode section to the provision script.

Also if you want to try out the P2P features, you can connect to my squeaknode instance here: tkdisnw23b6ek2x5lsyuy5ubhbxyvp4nfjbigzt6atxggpmdvesrapqd.onion:8555

openoms:
looks good, but would probably be useful to check for `zmqpubhashblock` entry in the bitcoin.conf in the install script and restart bitcoind if the system is ready. The bitcoin.install.sh currently only runs when setting up the parallel services.

Waiting to be refactored to be called from the build_sdcard.sh.

rootzoll:
@yzernik like openoms mentioned ... dont expect that the `zmqpubhashblock` setting was done during setup. Best would be that you check on your install if that config is part of the bitcoin config and if not add it and restart bitcoind.

rootzoll:
I am not able to merge or resolve confivts on this branch ... pushing to v1.8 for now.

PR #2749, opened 2021-11-30
Update electrs to 0.9.3

AaronDewes:
[Release notes here](https://github.com/romanz/electrs/blob/v0.9.3/RELEASE-NOTES.md).
This update dropped the `verbose` option, so I had to remove that. I haven't looked into your update/migration logic yet, not sure if there is anything required to update an existing config file, so @rootzoll please tell me if this is needed.


openoms:
ACK 45b3755

rootzoll:
> @rootzoll please tell me if this is needed

Looks good to me with the changed config. Thanks

PR #2755, opened 2021-12-01
Change conf edits

### LINKED ISSUE #2750: Potential config file corruption
This is a potential issue, I don't think it actually happened yet.

In many places, RaspiBlitz currently uses `echo "something=something" >> /mnt/hdd/raspiblitz.conf`.

If the user manually modifies the config file, and for some reason removes the newline at the end (it's unlikely, I just wanted to report this), this command will end up with the config file looking something like this:

Before:

```
oldConfig=oldValue
```

Expected:

```
oldConfig=oldValue
something=something
```

Actual config file:

```
oldConfig=oldValuesomething=something
```

In this scenario, doing `echo "something=something" | tee -a /mnt/hdd/raspiblitz.conf` would be better.

Even though this is unlikely you may find it's worth fixing it, because it can be fixed by a simple search and replace.

openoms:
fix #2750

### DIFF
--- a/home.admin/_bootstrap.sh
+++ b/home.admin/_bootstrap.sh
@@ -748,11 +748,11 @@ if [ ${valueExists} -eq 0 ]; then
   if [ ${#lndPort} -eq 0 ]; then
     lndPort="9735"
   fi
-  echo "lndPort='${lndPort}'" >> ${configFile}
+  echo "lndPort='${lndPort}'" | tee -a ${configFile}
 fi
 valueExists=$(cat ${configFile} | grep -c 'lndAddress=')
 if [ ${valueExists} -eq 0 ]; then
-  echo "lndAddress=''" >> ${configFile}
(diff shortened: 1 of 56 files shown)


AaronDewes (approved):

rootzoll:
For the changing of the config I already had a solution with a special script waiting in the branch:
https://github.com/rootzoll/raspiblitz/tree/2388redis

See script in detail:
https://github.com/rootzoll/raspiblitz/blob/2388redis/home.admin/config.scripts/blitz.conf.sh

It made sense to have it as part of the redis refactor because it will take care to keep the redis cache uptodate.

I just waited to merge that branch because it will change so many scripts that it might will conflict with all the outstanding PRs - so I wanted to wait for all community PRs to be merged into dev first and then clean the conflicts on my side.

@openoms so I am in favor to reject this PR and #2750 to be addressed when branch 2388redis gets merged.

openoms:
Sure, will close as there is a better solution.

PR #2757, opened 2021-12-01
Python3 pip deps failure in build_sd.sh

m0ntyG:
Section '# libs (for global python scripts)' was not able to run because of missing python dependencies
-> moving install python-deps in front of '# libs (for global python scripts' section


openoms:
Please rebase this to the dev branch.

Issue #2622, opened 2021-10-14
Password C is asked two times after startup


cstenglein:
At startup, Password C is asked to unlock the lightning wallet. (Working as intended).

But after that, it takes some minutes to re-sync everything and Password C is asked a second time to unlock.

rootzoll:
I saw this also now on one of testing machines. I think its not critical for the v1.7.1 release, but for the v1.8.0 when we have a WebUI just one unlock should be needed. So I added it to the v1.8.0 release issues to find out in detail why it happens and how we can prevent it for the future.

openoms:
It is difficult to reproduce so still investigating (possibly only happens on first setup? or when bitcoind is not fully synced yet?) but might be this:
https://github.com/rootzoll/raspiblitz/blob/2993ad76e8e856acf2ec7f2c925b19f48b7e8aa0/home.admin/config.scripts/blitz.statusscan.sh#L284

Should not unlock while not running since the output won't change and will be asked to unlock again.

Dehumanizer77:
I have installed my raspiblitz yesterday and I am also experiencing this. However, after initial setup and a few reboots after that, it was only asked once. Only after I have enabled some more apps, it started happening, so it might be connected with some apps beeing enabled...

Tracachang:
I have also experimented and it is because bitcoind is still loading (Loading block index...).

To avoid this instead of entering the password when is prompted, I wait until all blocks have been loaded and running, then entering the password just once is enough.

I guess that just changing TimeoutSec=120 to let's say 240 or 300 should be enough  in the lnd.install.sh/cl.install.sh files.

rootzoll:
@Tracachang thx thats worth a try

rootzoll:
merged the two PRs .. putting on final testing for v1.7.2 release candidates

rootzoll:
OK havent seen this problem anymore since the release candidates. Considered fixed. Closing issue.

PR #2795, opened 2021-12-13
increase timeoutsec to avoid re-entering the password lnd

Tracachang:
Related to the issue #2622 

Adding more time will avoid service to restart and having to re-enter password


rootzoll:
Will merge into temp branch to then cherry-pick into dev branch for v1.7.2 release candidates.

PR #2803, opened 2021-12-16
rename variables when datadrive is mounted

adevoss:
Other scripts source `blitz.datadrive.sh status` and expect the renamed variables.


adevoss:
Variable 'datadisk' is also used in other scripts. So `blitz.datadrive.sh status` has to return both.

PR #2807, opened 2021-12-16
remove files from /etc/ssh

adevoss:
Recursive option needed


rootzoll:
I am not sure anymore - but I think the recursive delete created problems on sshd restart - thats why I defaulted to the *-Notation to just make sure that all key files get deleted and recreated fresh on sshd restart.

adevoss:
Without the '-r' it displays an error message. If the command is more specific then there will be no error message.

PR #2809, opened 2021-12-16
blitz.datadrive.sh: extra names for variables used in other scripts

adevoss:
Both variable names are used in other scripts


rootzoll:
Basically the logic was that as HDD/SSD is an `hddCandidate` as long its not mounted - once it gets permanent mounted then it becomes the `datadisk`. Can you point me to a place in a script where the missing `hddCandidate` becomes a problem and makes the double output necessary?

rootzoll:
merging to ease VM mode

PR #2814, opened 2021-12-17
Copy ssh files to/from hdd

adevoss:
The ssh files on the hdd are overwritten and the subdirectories on the hdd has changed (mine had none).
And to preserve timestamps.
Please review.


rootzoll:
Yeah something gets mixed up with the directories in the old scripts - so it makes sense to put seperate it more clearly into `sshd` & `root_backup`.

BUT this will make everybody on the next update/recover to go with new ssh-keys and have this "SSH keys have changed ..." security warning. Maybe still the need to provide backwards compatibility - have to give it a bit more thought.

adevoss:
After I created the pr I thought also about backwards compatibility.
The conversion to the new directory structure can be done.
Also have an idea how, so I will create a new pr.

adevoss:
Replaced by #2818

PR #2818, opened 2021-12-18
Copy ssh files to/from hdd with backward compatibility

adevoss:
The ssh files on the hdd are overwritten and the subdirectories on the hdd has changed (mine had none).
And to preserve timestamps.
The ssh files on the hdd can be in old or new directory structure and will be converted to the new directory structure.
Replaces #2814.

Please review.

### DIFF
--- a/home.admin/_bootstrap.sh
+++ b/home.admin/_bootstrap.sh
@@ -268,7 +268,7 @@ fi
 if [ -d "/var/cache/raspiblitz/hdd-inspect/ssh" ]; then
   # INIT OLD SSH HOST KEYS on Update/Recovery to prevent "Unknown Host" on ssh client
   echo "SSH SERVER CERTS RESTORE activating old SSH host keys" >> $logFile
-  /home/admin/config.scripts/blitz.ssh.sh restore /var/cache/raspiblitz/hdd-inspect >> $logFile
+  /home/admin/config.scripts/blitz.ssh.sh restore /var/cache/raspiblitz/hdd-inspect/ssh >> $logFile
 else
   echo "No SSH SERVER CERTS RESTORE because no /var/cache/raspiblitz/hdd-inspect/ssh" >> $logFile
 fi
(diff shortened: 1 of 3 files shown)


rootzoll:
Good to have now a "converting" section ... let me see how this would work:

First of all its good to make a complete backup before any modifying/conversion - but I would put the backup under `/mnt/hdd/app-storage` because that directory will not be part of any further data migrations in the future.

If we move the ssh data anyway ... it would make sense to move it under the `/mnt/hdd/app-storage` directory so that its part of the migration data - what would make sense for the root ssh data. So lets move that directory anyway.

Then take a look how the old `/mnt/hdd/ssh` directory on the HDD/SSH looks like:
```
admin@192.168.178.50:~ ₿ sudo ls -la /mnt/hdd/ssh
total 632
drwxr-xr-x  5 root root   4096 Dec 17 23:44 .
drwxr-xr-x 10 root root   4096 Dec 18 00:28 ..
-rw-r--r--  1 root root 577771 Dec 17 23:44 moduli
drwxr-xr-x  5 root root   4096 Dec 17 23:44 ssh
-rw-r--r--  1 root root   1650 Dec 17 23:44 ssh_config
drwxr-xr-x  2 root root   4096 Dec 17 23:44 ssh_config.d
-rw-r--r--  1 root root   3289 Dec 17 23:44 sshd_config
drwxr-xr-x  2 root root   4096 Dec 17 23:44 sshd_config.d
-rw-------  1 root root   1381 Dec 17 23:44 ssh_host_dsa_key
-rw-r--r--  1 root root    606 Dec 17 23:44 ssh_host_dsa_key.pub
-rw-------  1 root root    505 Dec 17 23:44 ssh_host_ecdsa_key
-rw-r--r--  1 root root    178 Dec 17 23:44 ssh_host_ecdsa_key.pub
-rw-------  1 root root    411 Dec 17 23:44 ssh_host_ed25519_key
-rw-r--r--  1 root root     98 Dec 17 23:44 ssh_host_ed25519_key.pub
-rw-------  1 root root   2602 Dec 17 23:44 ssh_host_rsa_key
-rw-r--r--  1 root root    570 Dec 17 23:44 ssh_host_rsa_key.pub
-rw-r--r--  1 root root    342 Dec 17 23:44 ssh_import_id
```

That `drwxr-xr-x  5 root root   4096 Dec 17 23:44 ssh` directory shouldnt be there anyway - but happend somehow. On conversion this should be deleted. 

Then  all the rest should be moved to `/mnt/hdd/app-storage/sshd` and if `mnt/hdd/ssh/root_backup` exists it should be moved to `/mnt/hdd/app-storage/ssh-root`.

I will create a `2818ssh` branch - merge this PR there and adpat the directories & conversions.

PR #2830, opened 2021-12-20
Update broken links / use GitHub links.

FeedMyTummy:
Need help with one more link. Do not know which one is the new link for
https://raspibolt.github.io/raspibolt/raspibolt_20_pi.html#raspi-config

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -178,7 +178,7 @@ Of course, people should modify the system, add own scripts, etc ... but if you 
 ### How can I update LND or bitcoind even before the next RaspiBlitz update?
 
 Try updating before a official RaspiBlitz at your own risk - you can find some info about that here:
-https://stadicus.github.io/RaspiBolt/raspibolt_faq.html#how-to-upgrade-bitcoin-core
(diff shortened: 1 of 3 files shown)


rootzoll:
About the links in general ... I think the official links are now to `raspibolt.org` - like this one:
https://raspibolt.org/operating-system.html#configure-boot-options

FeedMyTummy:
@rootzoll Yes. I changed to the GitHub repo as https://github.com/raspibolt/raspibolt/ is the backing source for https://raspibolt.org. Do you prefer links to https://raspibolt.org?

rootzoll:
Yeah I dont think we should have links with long blobIDs `a21788c0518618d17093e3f447f68a53e4efa6e7` and use the raspibolt.org ones.

FeedMyTummy:
@rootzoll Moved to the raspibolt.org links. The reason for the prepare-wifi long blob is that it has been removed from the repo (Thus, from raspibolt.org) and using Raspberry Pi Imager is what is now recommend by the author. I left the manual  instructions (the one found in the long blob) as that may be of use to someone who is using a flashing software other than Raspberry Pi Imager.

rootzoll:
perfect. much tanks. merging :)

PR #2840, opened 2021-12-22
Update NodeJS to v16.10.0

knorrium:
Update NodeJS to v16.10.0 which is the [last v16 release](https://nodejs.org/en/download/releases/) bundled with npm 7.

### DIFF
--- a/home.admin/config.scripts/bonus.nodejs.sh
+++ b/home.admin/config.scripts/bonus.nodejs.sh
@@ -3,11 +3,11 @@
 # consider installing with apt when updated next
 # https://github.com/nodesource/distributions/blob/master/README.md#installation-instructions
 
-VERSION="v14.17.6"
+VERSION="v16.10.0"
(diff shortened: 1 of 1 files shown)


rootzoll:
Good idea to update NodeJS for next release ... but isnt v16.13.1 the lastest LTS version? Any particular reason why v16.10.0?

knorrium:
@rootzoll we recently upgraded mempool to angular 13 and there were some dependency issues with node 16.13+npm 8 so we pinned to the latest v16 which came with npm 7. 

I think ideally we should be using a node version manager like nvm so different apps don’t have to rely on the system wide node installation

rootzoll:
Yeah nvm could be an idea for the future, but that would be some deeper changes. For the v1.7.2 I will then give `v16.10.0` & `npm7` a shot and see how the other apps work on it. Mergeing into dev for final testing.

Issue #2813, opened 2021-12-17
[dev] lnd not running after install


adevoss:
After the final reboot lnd is not running:
`[WRN] LTND: Error validating config: ValidateConfig: error normalizing RPC listen addrs: parse address 0.0.0.0:10009 failed: dial tcp 127.0.0.1:9050: connect: connection refused`
(See https://github.com/rootzoll/raspiblitz/issues/2772#issuecomment-996876250)

rootzoll:
I just tested with a fresh build of the sd card in `dev` brach and lnd is running on fresh install. 
Is your report from a scenario running a VM?

adevoss:
Yes. lnd still not started.
Where do I look for logs? `/mnt/hdd/lnd/logs/bitcoin/mainnet` only contain old *.gz files.

openoms:
you can use the alias `lndlogs` and the `SYSTEM` menu also
the lnd logs and path is in `debug` too

adevoss:
> use the alias `lndlogs`

Does not work.
`debug -l ` gives no link to share

`debug`:
`tail: cannot open '/mnt/hdd/lnd/logs/bitcoin/mainnet/lnd.log' for reading: No such file or directory`
...
```
*** LAST BLOCKCHAIN (MAINNET) 20 INFO LOGS ***
sudo tail -n 20 /mnt/hdd/bitcoin/debug.log
2021-12-20T15:44:24Z connect() to *.*.*.*:9050 failed after wait: Connection refused (111)
2021-12-20T15:44:25Z connect() to *.*.*.*:9050 failed after wait: Connection refused (111)
```
...
```
*** LAST LND (MAINNET) ERROR LOGS ***
sudo journalctl -u lnd -b --no-pager -n12
-- Journal begins at Mon 2021-12-06 22:32:55 CET, ends at Mon 2021-12-20 16:44:34 CET. --
Dec 20 16:43:03 RASPIBLITZ-LIVE systemd[23479]: lnd.service: Failed to locate executable /usr/local/bin/lnd: No such file or directory
Dec 20 16:43:03 RASPIBLITZ-LIVE systemd[23479]: lnd.service: Failed at step EXEC spawning /usr/local/bin/lnd: No such file or directory
Dec 20 16:43:03 RASPIBLITZ-LIVE systemd[1]: lnd.service: Main process exited, code=exited, status=203/EXEC
```

openoms:
need find out why Tor does not work eg.:

[...]

PR #2846, opened 2021-12-23
typos in lnd.install.sh

adevoss:


openoms:
Is there a functional difference with these changes or are these more of a style suggestion?

adevoss:
There is no functional difference but without these changes the script returns errors and exits.

openoms:
The double `==` are not needed in bash scripts
the added spaces outside the brackets are not needed
comparing to single words with or without " " are also the same.
Are you sure the errors you are getting are fixed by these changes?

adevoss:
No, not sure. At the same time I created this pr there was this (#2847) which you solved at the same time yourself.
I don't know if that caused the errors or the if statement syntax.
Now I will try an install of `dev` branch.
If that gives no errors then I will close this.

adevoss:
Setup has no errors. But there is
```
/usr/local/bin/lnd --configfile=/home/bitcoin/.lnd/lnd.conf
-bash: /usr/local/bin/lnd: No such file or directory
```
(#2813)

PR #2847, opened 2021-12-23
aliases file does not exist

adevoss:
If the file doesn't exist the script shows errors.
Shouldn't it be `/home/admin/.bash_aliases` (sourced by `/home/admin/.bashrc`)


adevoss:
Solved by @openoms

Issue #2852, opened 2021-12-26
Improve build option parsing for scripts


nyxnor:
If this is implemented the benefits are:
* not necessary to specify the options in order
* if the option requires and argument, warn and exit
* if the option was not specified, use default values
* if option value is not withing the range, warn and exit

I did this one example for the build_sdcard.sh.
```sh
$ sh build_sdcard.sh 
build_sdcard.sh [--option <argument>]

Options:
  -i, --interaction [0|1]                  interaction before proceeding with exection (default: 0)
  -f, --fatpack [0|1]                      fatpack mode (default: 0)
  -u, --github-user [rootzoll|other]       github user to be checked from the repo (default: rootzoll)
  -b, --branch [v1.7|v1.8]                 branch to be built on (default: v1.7)
  -d, --display [lcd|hdmi|headless]        display class (default: lcd)
  -t, --tweak-boot-drive [0|1]             tweak boot drives (default: 1)
  -w, --wifi-region [US|GB|other]          wifi iso code (default: US)

Notes:
  all options, long and short accept --opt=value mode also
  [0|1] can also be referenced as [false|true]

$ sh build_sdcard.sh -i 1 -f 1 -u nyxnor -b v1.6 -t 0 -w GB
interaction=1
fatpack=1
github_user=nyxnor
branch=v1.6
display=lcd
tweak_boot_drive=0
wifi_region=GB

$ sh build_sdcard.sh -i
build_sdcard.sh: Option '-i' requires an argument.

$ sh build_sdcard.sh -i 2
build_sdcard.sh: Option '--interaction' cannot be '2'! It can only be: 0 1 false true.

$ sh build_sdcard.sh -i 1 -f 1
interaction=1
fatpack=1
github_user=rootzoll
branch=v1.7
display=lcd
tweak_boot_drive=true
wifi_region=US

$ sh build_sdcard.sh --interaction 1 --fatpack=1
interaction=1
fatpack=1
github_user=rootzoll
branch=v1.7
display=lcd
tweak_boot_drive=true
wifi_region=US
```

```sh
#!/usr/bin/env sh

#set -x

me="${0##/*}"

nocolor="\033[0m"
red="\033[31m"
## default user message
error_msg(){ printf %s"${red}${me}: ${1}${nocolor}\n"; exit 1; }

## usage as a function be be called whenever there is a huge mistake on the options
usage(){
  printf %s"${me} [--option <argument>]

Options:
[...]

PR #2912, opened 2022-01-25
make boolean options have the same value

nyxnor:
https://github.com/rootzoll/raspiblitz/issues/2852

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -55,6 +55,10 @@ assign_value(){
     -*) value="${2#-}";;
     *) value="${2}"
   esac
+  case "${value}" in
+    0) value="false";;
+    1) value="true";;
+  esac
   ## Escaping quotes is needed because else if will fail if the argument is quoted
   # shellcheck disable=SC2140
   eval "${1}"="\"${value}\""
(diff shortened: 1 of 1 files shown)


build_sdcard.sh
```diff
@@ -116,9 +120,11 @@ range_argument(){
   fi
 }
 
-echo -e "\n*** SOFTWARE UPDATE ***"
 general_utils="curl"
-sudo apt install -y ${general_utils}
+if ! command -v curl >/dev/null; then
```
openoms: thanks, this would have been my other suggestion, to check for curl

openoms:
LGTM

build_sdcard.sh
nyxnor: improved now if there are multiple programs on general utils https://github.com/rootzoll/raspiblitz/pull/2912/commits/a0f7674c426818766174a64feb918bc9c71e6815

openoms:
since curl is the only thing needed why not just keep it simple and get rid of the `general_utils` variable?

nyxnor:
Just in case for the future if any utility is needed to be downloaded before confirmation.

openoms:
consider readibility being a helpful attribute to help new contributors and who are using these scripts as a guide

nyxnor:
I can add comments...

If command is not found on path, than add it to the list of commands to be installed
```sh
! command -v ${prog} >/dev/null && general_utils_install="${general_utils_install} ${prog##*/}"
```

If any program is not installed, install only the required ones
```sh
if [ -n "${general_utils_install}" ]; then
```
this avoids trying to install program if update fails
```sh
sudo apt update -y && sudo apt install -y ${general_utils_install}
```

rootzoll:
I would like to make a new release candidate soon. Please signal me, when this is ready to merge.

nyxnor:
on my part it is done

PR #2916, opened 2022-01-28
Update bonus.thunderhub.sh

apotdevin:
In the latest release there is a breaking change. To run on a different port you now have to specify an env var


rootzoll:
Thanks for the fast PR :)

apotdevin:
np! Thanks for the fast merge

PR #2415, opened 2021-07-09
Bitcoinminds.org to be made available as a service in Raspiblitz

raulcano:
With this PR I would like to enable a new Raspiblitz service to allow any Raspi to host and run fully the BitcoinMinds.org site.
This pull request adds 3 things:
- A bonus script to download the code from the bitcoinminds.org repo https://github.com/raulcano/bitcoinminds
- A menu item to enable and disable this service.
- A command 'bm' to start the server (npm run serve) and make the site available to be visited from the LAN

I leave this PR in draft until I have properly tested it in my Raspiblitz.

image

image

image

image

image

image

PR #2933, opened 2022-02-07
Bitcoinminds.org to be made available as a service in Raspiblitz

raulcano:
With this PR I would like to enable a new Raspiblitz service to allow any Raspi to host and run fully the BitcoinMinds.org site.
This pull request adds 3 things:
- A bonus script to download the code from the bitcoinminds.org repo https://github.com/raulcano/bitcoinminds
- A menu item to enable and disable this service.
- A command 'bm' to start the server (npm run serve) and make the site available to be visited from the LAN

I leave this PR in draft until I have properly tested it in my Raspiblitz.

NOTE: this PR replaces a deleted one #2415 

image

image

image

image

image

image

### DIFF
--- a/home.admin/00settingsMenuServices.sh
+++ b/home.admin/00settingsMenuServices.sh
@@ -43,7 +43,7 @@ if [ "${network}" == "bitcoin" ]; then
   OPTIONS+=(a 'BTC Mempool Space' ${mempoolExplorer})
   OPTIONS+=(j 'BTC JoinMarket+JoininBox menu' ${joinmarket})
   OPTIONS+=(w 'BTC Download Bitcoin Whitepaper' ${whitepaper})
-  OPTIONS+=(v 'BTC BitcoinMinds.org in Local' ${bitcoinminds})
+  OPTIONS+=(v 'BTC Install BitcoinMinds.org' ${bitcoinminds})
 fi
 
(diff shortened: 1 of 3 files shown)


raulcano:
Once the service is started with "npm run serve", there is a server in the local raspi running at "localhost:8081".

If I want to access that web UI from my laptop (connected to the network) I cannot just type 

192.168.X.Y:8081

At the moment, I need to create an ssh tunnel to access the UI, which is less than ideal.

I suspect there must be some firewall restriction.
For other webUIs in raspiblitz (e.g. RTL), this works well. Can you guys help me on how to configure the firewall properly? thanks!

rootzoll:
> I suspect there must be some firewall restriction.

RaspiBlitz uses `ufw` - so you would need to open that port on install script - like with:
```
sudo ufw allow 8081 comment 'bitcoinminds'
```

> Once the service is started with "npm run serve", there is a server in the local raspi running at "localhost:8081".

port 8081 might be a collusion with loop swap server - would be best to chose another default port.

raulcano:
Excellent, that works as expected :)
With this , I set the PR as ready for review.
Cheers!

rootzoll:
This tests OK - will merge for 1.7.2RC2c

PR #2936, opened 2022-02-07
CL install from source from github instead of zip

openoms:
installing from zip causes an error with the pyln-client:
https://github.com/rootzoll/raspiblitz/issues/2662#issue-1034847654
so switched to installation from source from github

left in the well tested code to install from zip, but commented out in case we would use it later


rootzoll:
OK will merge and test with new image - opened this issue for final testing #2937

PR #2951, opened 2022-02-12
feat: removes preinstalled cups

GoetzGoerisch:


rootzoll:
Thanks - marking for v1.8 release where we will build the next new sd card base image.

purpleKarrot:
Instead of removing unnecessary packages, wouldn't it be more appropriate to use the non-desktop base image and then only install needed packages?

According to README.md, the base image is taken from: https://downloads.raspberrypi.org/raspios_arm64/images
Why not use this instead: https://downloads.raspberrypi.org/raspios_lite_arm64/images

openoms:
ACK - we don't need cups
but we need the desktop environment for the display @purpleKarrot

PR #2955, opened 2022-02-12
add gzip to all nginx configs; add http2 support for all https nginx configs

cstenglein:
This enables http2 support to the HTTPS version of the index page.

Would be a nice addition when accessing the webui / the default index page over HTTPS.

I used the certificate & key from the `app-data/nginx` directory which I don't know if I should use.

No automatic redirect to the HTTPS version was currently set => needs to be discussed if we enable it bc it's a self-signed certificate and could lead to confusion if non-technical users connect to it over local IP.

As always proof image below. Tested on 1.7.2rc2:

http2

EDIT:

I think the other services would also benefit from HTTP2 support. Should I do separate PRs for them or put them all in this one?

EDIT2: 

I also added `gzip` to compress the file sizes. It's not a big win for the index page, but the apps / webui would benefit very much from it.

The red line is the Request with http 1.1 & gzip off, blue http2 with gzip:

image

### DIFF
--- a/home.admin/assets/nginx/sites-available/btcpay_ssl.conf
+++ b/home.admin/assets/nginx/sites-available/btcpay_ssl.conf
@@ -1,12 +1,14 @@
 ## btcpay_ssl.conf
 
 server {
-    listen 23001 ssl;
-    listen [::]:23001 ssl;
+    listen 23001 ssl http2;
+    listen [::]:23001 ssl http2;
     server_name _;
 
     include /etc/nginx/snippets/ssl-params.conf;
     include /etc/nginx/snippets/ssl-certificate-app-data.conf;
+
+    include /etc/nginx/snippets/gzip-params.conf;
 
     access_log /var/log/nginx/access_btcpay.log;
     error_log /var/log/nginx/error_btcpay.log;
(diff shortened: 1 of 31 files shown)


cstenglein:
Converting to draft - will optimize further and enable for all apps in this PR.

cstenglein:
Added gzip params to all apps which provide nginx configs now.

Also for the https versions, added http2 support.

## Some stats:

I only checked with TOR + HTTPS which should show the biggest increase in performance due to the lower speed in the TOR network. The initial load sizes should benefit http + https connections though.

### Thunderhub

Thunderhub didn't see a big speed boost cause they load a 3,4MB GIF. This will be resolved in https://github.com/apotdevin/thunderhub/pull/406

Otherwise, you can see some small improvements in transferred size but a nicer one in transfer time:

#### Before:
thunderhub_before

#### After:
thunderhub_after


### BTC Pay

Here the changes showed a really nice improvement in speed, around 1/3 of the previous transfer size. Forgot to include the total size in this screens.

#### Before:

btc-pay_before

#### After:

btc-pay_after

#### Mempool

This also shows a really nice decrease in transfer size, 842kb vs 2,41MB (~1/3).

#### Before:
mempool_before

#### After:

mempool_after

rootzoll:
Cool work :) I added this for the v1.8 release ... the v1.7.2 is a bit too close to call for a change touching so many places.

cstenglein:
Yes, would also see this for 1.8 :)

Issue #2924, opened 2022-02-01
Fulcrum: the most performant Electrum Server implementation


openoms:
See the udpatedd article from Craig Raw: https://www.sparrowwallet.com/docs/server-performance.html

https://twitter.com/craigraw/status/1488470446211321856?s=19

Fulcrum: https://github.com/cculianu/Fulcrum

arm64 binaries are published for the RPi.

Proceeding to test.

openoms:
First sync took ~48h on an RPi4: https://twitter.com/openoms/status/1487099572694437892
Seems to be a viable option which could be offered next to / instead of Electrs.

A quick manual guide to install: https://github.com/openoms/bitcoin-tutorials/blob/master/fulcrum.md

openoms:
Test this install script https://github.com/openoms/raspiblitz/blob/fulcrum/home.admin/config.scripts/bonus.fulcrum.sh:
```
# download
wget -O bonus.fulcrum.sh https://raw.githubusercontent.com/openoms/raspiblitz/fulcrum/home.admin/config.scripts/bonus.fulcrum.sh

# check
cat bonus.fulcrum.sh

# run with debug
bash -x bonus.fulcrum.sh
```

Looking to test:
- [x]  sync while txindex is being built
- [x]  sync without ZRAM
- [ ] 2GB RAM

TODO:
- [ ] menu
- [ ] status
- [ ] how to connect
- [ ] integrate with mempool
- [ ] integrate with btc-rpc-explorer
- [ ] ask to uninstall database

openoms:
More experience: https://github.com/rootzoll/raspiblitz/pull/2966#issuecomment-1059190101

- [x] try to use the same SSL certificate as for electrs

Sparrowhawk1337:
**Hardware**

- RPi4B
- 8GB RAM 
- 1Tb SSD (powered from an expansion board not from RPi)
- SanDisk Extreme 32GB microSD Card (A2, C10, U3, V30)

**Software**

- Fresh install of RaspiBlitz v1.7.2
- Fulcrum from script https://openoms.github.io/bitcoin-tutorials/fulcrum.html

**Run 1**

- no ZRAM set
- fulcrum.conf using settings for 8Gb RAM
- fulcrum sync failed at 90.9%
- log file [fulcrum_log_extract_1.txt](https://github.com/rootzoll/raspiblitz/files/8239241/fulcrum_log_extract_1.txt)

**Run 2**

- no ZRAM set
[...]

PR #2966, opened 2022-02-15
[New] Fulcrum install script

openoms:
Moved the description to: https://github.com/rootzoll/raspiblitz/issues/2924

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@
 - New: Migration from Citadel to RaspiBlitz [details](https://github.com/rootzoll/raspiblitz/issues/2642)
 - New: Bitcoinminds.org local on RaspiBlitz [details](https://github.com/raulcano/bitcoinminds)
 - New: JoinMarket Web UI v0.0.3 (CLI install of the first public alpha release) [details](https://github.com/joinmarket-webui/joinmarket-webui/releases/tag/v0.0.3)
+- New: Fulcrum Electrum server v1.6.0 (CLI install script) [issue](https://github.com/rootzoll/raspiblitz/issues/2924)
 - Update: LND v0.14.2-beta [details](https://github.com/lightningnetwork/lnd/releases/tag/v0.14.2-beta)
 - Update: C-lightning v0.10.2 [details](https://github.com/ElementsProject/lightning/releases/tag/v0.10.2)
 - Update: LNbits 0.6.0 [details](https://github.com/lnbits/lnbits-legend/releases/tag/0.6.0)
--- a/home.admin/config.scripts/bonus.fulcrum.sh
+++ b/home.admin/config.scripts/bonus.fulcrum.sh
@@ -106,16 +106,16 @@ rpcpassword = ${PASSWORD_B}
 # RPi optimizations
 # avoid 'bitcoind request timed out'
 bitcoind_timeout = 300
-# reduce load
+# reduce load (4 cores only)
 bitcoind_clients = 1
-worker_threads = 2
+worker_threads = 1
 db_mem=1024
 
-# settings tested with 4GB RAM + ZRAM
+# for 4GB RAM
 db_max_open_files=200
 fast-sync = 1024
 
-# settings testetd with 8GB RAM + ZRAM
+# for 8GB RAM
 #db_max_open_files=500
 #fast-sync = 2048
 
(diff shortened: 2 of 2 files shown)


openoms:
@rootzoll I (and few others) have synced Fulcrum on 4 and 8GB blitzes with and without ZRAM set.
I think this initial install script could be the part of v1.7.2 as a CLI-only option as it is . 
Will continue to iterate on it, but good enough to get more testing.

cricktor:
Oh, this looks promissing, thank you! Looking forward to give it a try and hopefully I can dismiss electrs which kinda hates me.

cricktor:
I ditched electrs from my RaspiBlitz 1.7.2 (8GB RAM) and Fulcrum took around 4 days to build his index (I have no ZRAM active and used the options for 8GB RAM in provided fulcrum.conf file; I didn't tweak other options afair). My node has TOR enabled.

I'm not sure why my initial sync took roundabout double the time as @openoms reported. I have seen some use of the swap file but didn't track it. So I don't know if there was some heavy swapping involved which would've delayed the process. LND is running with only a few channels open.

I had some issues to connect my Electrum wallet to my Fulcrum: it seemed to work only by using the node's hostname:50022 as the electrum server address for the Electrum wallet. No chance within the local network to use the server address as <local IP address>:50022, frankly no idea why that failed. Electrum wallet didn't switch to green light for the server connection with anything other than hostname:50022 as electrum server address.

Having only Fulcrum in the very basic setup provided by the script bonus.fulcrum.sh needs some more tweaks to be properly recognized by other services which I have active on my node. Of course this is mainly caused by missing electrs and the very rudimentary integration of Fulcrum in the node's setup. Some minor tweaks were necessary to make Mempool and BTC-RPC-Explorer happy with Fulcrum.

So far I am very happy with Fulcrum, for me it works much better than electrs and likely I won't go back to electrs unless some very major issues come up with Fulcrum. Looking forward for better integration into RaspiBlitz!

What I would love to see if it's feasible to preserve the Fulcrum index especially in the process of a RaspiBlitz update. To my knowledge for electrs the index had always to be rebuilt. This was only a minor nuissance as this took only about a day or less. Four days, at least for me, would be a bit more painful to wait for.

openoms:
thank you for comment, it is very valuable.

I limited the CPU usage to 1 core to make sure other services can run undisturbed. Otherwise I have seen load numbers of 5-7 while syncing which means multiple processes waiting for the cpu. This can result in slower sync.
Also it could be a result of syncing the txindex the same time. Were you running the BTC-RPC-Explorer before (that would switch the txindex on already)? 

About the problem with the LAN IP vs hostname: the router just resolves the hostname to the IP so they are practically the same.
The problem is likely that Fulcrum uses a different SSL certificate so you need delete the old one saved atvthe client (on the desktop/phone) when you connected Electrs.
We can look into a way to use the same.

The database for both Electrs and Fulcrum is stored on the disk, so no resync needed on SDcard change. It will remain so.

cricktor:
Oh, didn't know that Electrum remembers the former SSL certificate or do I get it wrong?

I had mostly your template for `fulcrum.conf` active during the sync, like this:
```
datadir = /home/fulcrum/.fulcrum/db
bitcoind = 127.0.0.1:8332
rpcuser = <redacted>
rpcpassword = <redacted>

# RPi optimizations
# avoid 'bitcoind request timed out'
bitcoind_timeout = 600
# reduce load (4 cores only)
bitcoind_clients = 1
worker_threads = 1
db_mem=1024

# for 4GB RAM
#db_max_open_files=200
#fast-sync = 1024

# for 8GB RAM
db_max_open_files=500
fast-sync = 2048

# server connections
# disable peer discovery and public server options
peering = false
announce = false
tcp = 0.0.0.0:50021
# ssl via nginx

# tweaks for large address histories (need those for some blockchain research)
max_history=250000
```
Well, I thought this would already apply some limits to Fulcrum's threads, but I've seen load numbers 5-7 during sync and quite some processes active with top or htop. So, apparently my fulcrum.conf didn't stop Fulcrum to spawn a whole lot of workers.

I had txindex=1 already synced for bitcoind and I have used services Mempool and BTC-RPC-Explorer before I abandoned Electrs and switched to Fulcrum.

I know that the indexes are stored on the SSD and hopefully shouldn't be affected during an update of RaspiBlitz, unless something weird happens.

The speed of Fulcrum is a joy when I tested it with some watch-only wallets and addresses where I had big issues with electrs before. Thanks for the addition to RaspiBlitz and special thanks to @cculianu!

Issue #2772, opened 2021-12-07
Test: Install on VM (Debian QEMU host and Debian guest)


adevoss:
Steps
- Install Debian (no gui) on two partitions (boot/root)
- Install sudo

- Download build script
  add debian to line 211
  modify tor repo to bullseye in line 212-214
  comment out raspberry pi repo
  
- Execute build sdcard script from my own home directory
  `sudo ./build_sdcard.sh false false rootzoll dev headless true false`

- Reboot
- Login as the user I created with the initial install (not 'pi' or 'admin')
  When login as admin I get a blue screen 'Wait...' and then the main menu.
  I don't know if the next script is executed without errors

- Edit _bootstrap.sh
  ./_bootstrap.sh: line 301: HDD/SSD connected: ${$hddCandidate}: bad substitution
  removed the second '$'

- Edit /mnt/hdd/raspiblitz.conf
  Added the following line:
  `uaspForced=0`

- Execute `sudo ./_bootstrap.sh` from /home/admin

adevoss:
These two errors are displayed when executing the build sdcard script:
```
Adding user admin to group sudo
Done.
%sudo ALL=(ALL) NOPASSWD:ALL
mv: cannot stat './raspiblitz.info': No such file or directory

```
and
```
Cloning into 'raspiblitz'...
remote: Enumerating objects: 34915, done.
remote: Counting objects: 100% (2772/2772), done.
remote: Compressing objects: 100% (940/940), done.
remote: Total 34915 (delta 2035), reused 2526 (delta 1832), pack-reused 32143
Receiving objects: 100% (34915/34915), 99.86 MiB | 2.84 MiB/s, done.
Resolving deltas: 100% (26091/26091), done.
cp: -r not specified; omitting directory '/home/admin/raspiblitz/home.admin/config.scripts'
cp: -r not specified; omitting directory '/home/admin/raspiblitz/home.admin/setup.scripts'
```

adevoss:
When I execute the bootstrap script I get:
```
Resetting the InfoFile: /home/admin/raspiblitz.info
ls: cannot access '/boot/stop': No such file or directory
ls: cannot access '/home/admin/provision.flag': No such file or directory
# OK - file exists: /home/admin/raspiblitz/pictures/startlogo6.png
*** Checking Log Size ***
OK - logs are at 66 MB - within safety limit
```
and at the end
```
Failed to enable unit: Unit file bitcoind.service does not exist.
[...]

PR #2984, opened 2022-02-21
Update Helipad version

SpencerPearson:
v0.1.9 just released, update config script to pull latest version


rootzoll:
Thanks. Marked for next version release.

adevoss:
#2772 has a 1.8 milestone tag
Don't know if that means it is officially supported. That up to the Raspiblitz devs.
On a Debian VM this install script will fail because of the missing package 'pkg-config'

PR #2987, opened 2022-02-22
Fix for JoinMarket web UI

dennisreimann:
The SSL file permissions didn't work, hence we generate them on our own now. Also removed a leftover nginx file for the superfluous http version.


rootzoll:
@openoms can you give it a quick test & then we can merge into v1.7 for people to use "PATCH" to update for that feature fix?

dennisreimann:
I think it needs one more commity though, because the files for existing installs would be present (the symlinks to the nginx cert files have been made). We need to replace them and the if prevents that right now.

dennisreimann:
Done, though I think it needs to be reinstalled with `off` and then `on` anyways to fix the problem. Nevertheless, at least fresh installs should work now. Existing installs will need to first run `config.scripts/bonus.joinmarket-webui.sh off` and then reinstall.

openoms:
@dennisreimann is this to be tested together with #2988 ?

dennisreimann:
They are separate issues, so they can be tackled after another – this one first, because it solves a general installation problem. 

If it makes testing on your end easier, we can @dnlggr and I can certainly integrate/merge the two PRs.

openoms:
no need, I can carry on testing separately, the changes are straightforward and already halfway.

Once merged will also change to downloading the source code with git rather than using the tar.gz as the tar.gz is not signed with PGP.

openoms:
Tested ok, merging.
Users on v1.7.2 will be able to run the command `patch` to update to the latest commits in the v1.7 branch. (More details in the UPDATE -> PATCH menu)

dennisreimann:
Thank you, @openoms!

PR #2988, opened 2022-02-23
Fix auth header for JoinMarket Web UI

ghost:
This is a fix necessary for the JoinMarket Web UI to be able to talk to the `jmwalletd` API. Sorry for the confusion we should've tested the Raspiblitz integration again after finalizing v0.0.3 of the web UI.

More details in https://github.com/joinmarket-webui/joinmarket-webui/issues/121.

**TL;DR:** Integration in Umbrel required us to enable Basic Auth in order to protect certain API endpoints and the web UI in general. To avoid a conflict between Basic Auth and the JoinMarket bearer token in the `Authorization` HTTP header, we changed the web UI to send the JoinMarket token in the `x-jm-authorization` header. This makes it possible for webservers to enable Basic Auth but also means that webservers must rewrite the `Authorization` header after Basic Auth so that the `Authorization` header then includes the JoinMarket token originally sent in `x-jm-authorization` before proxying the request to `jmwalletd`.

This is a minimal fix to make requests work again. We could, of course, enable Basic Auth on Raspiblitz as well just to be safe.

### DIFF
--- a/home.admin/assets/nginx/sites-available/joinmarket_webui_ssl.conf
+++ b/home.admin/assets/nginx/sites-available/joinmarket_webui_ssl.conf
@@ -19,6 +19,10 @@ server {
 
     location /api/ {
         include /etc/nginx/snippets/ssl-proxy-params.conf;
+
+        proxy_set_header Authorization $http_x_jm_authorization;
+        proxy_set_header x-jm-authorization "";
+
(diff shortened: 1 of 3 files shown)


openoms:
Ok, nginx still up and the webUI is functional with the changes so merging.

Thank you for the fix!

PR #2990, opened 2022-02-23
Fix unit for C-Lightning fee report on info screen

### LINKED ISSUE #2989: Info screen shows C-Lightning routing fee report in sats instead of msat
Running CL, the Fee Report on the info screen is displayed as sat, but the value presented actually is in msat. So the fee report overstates earnings by 1000x. And I actually haven't made millions of sats :)

kev-vl:
Fixes #2989


openoms:
thanks for the report and fix, just needs testing especially if it fits the LCD with the extra letter.

IS your display looking ok with the change?

kev-vl:
I don't have an LCD on my blitz, so I'm not able to test if it fits. Caught the error while running the INFO option in an SSH session. 

Would be nice if someone with an LCD display could check.

openoms:
a better approach might be to simply divide the CL provided value with 1000 to get sats and benefit from having 3 more digits and an `m` worth of space before it wraps

home.admin/00infoBlitz.sh
```diff
@@ -236,7 +236,7 @@ if [ "${lightning}" != "" ]; then
     else
       source <(/home/admin/_cache.sh meta ln_${lightning}_${chain}net_fees_total)
       ln_totalfees="${value}"
-      ln_feeReport="Fee Report: ${color_green}${ln_totalfees} ${color_gray}${netprefix}sat"
+      ln_feeReport="Fee Report: ${color_green}${ln_totalfees} ${color_gray}${netprefix}msat"
```
openoms, line 239: could use:
```
ln_feeReport="Fee Report: ${color_green}$((ln_totalfees/1000)) ${color_gray}${netprefix}sat"
```

openoms:
We need to do either. That line is not limited, can take millions of sats even if displayed as msats.

We can stick to more accurate measure and show the power of unit bias by displaying the earnings in msats.

What do you think @rootzoll ?
The PR could be just merged for a quick fix.

kev-vl:
FWIW I prefer msats for @openoms' arguments, and because it's consistent with the units used in channel fees. 

Perhaps it could use some readability tweaks (11054220 msat vs. 11,054,220 msat) but that's a different discussion.

openoms:
@kev-vl thanks for the comment. Good idea about readability, but indeed currently we just use the number as c-lightning provides it. Would only use spaces as commas-s vs dots can be confusing depending on the location.

Merging this now to be able to fix with simply running `patch`

PR #2992, opened 2022-02-23
no dns spam by netstat / use netstat -n

PatrickScheich:
When electrs is indexing this `netstat` spams dns requests, use `netstat -n` instead


rootzoll:
Thanks. Also cherry-picked into dev branch.

Issue #2995, opened 2022-02-25
armbian-config non-exsitance breaks armbian SDcard build


harvhat:
line 293 of  build_sdcard.sh

'''
[ "${baseimage}" = "armbian" ] && armbian_dependencies="armbian-config" # add armbian-config
'''

However there is no package called "armbian-config" so the whole 'apt install' line fails while the build script continues on anyway, unsuccessfully of course.

I'm thinking armbian-config should be a variable that points to a list of packages, perhaps?

nyxnor:
We haven't tested armbian for a while. That was an old line if you follow the blames.

Can you check any package with a similar name on armbian? Look at their source repositories.

harvhat:
I didn't realize armbian was a OS distro. The comment in the code "# experimental: fallback for all debian on arm" had me thinking it was just what a moniker for any debian on arm, hence my confusion. Indeed there appears to be a armbian-config package in armbian, however I was not trying to install raspiblitz that armbian so didn't test it. I'm using raspios_arm64 but it didn't get picked up because the hostname was not "raspberrypi" (I already have other raspberrypi's on my network after all!)

That, however, is a different issue. (I added an option to force a particular baseimage which defaults to "auto" to solve that)

More related to this issue though is the build process breaking when even 1 of the items on the apt lengthy install list isn't found, for then _none_ of them are installed, yet the build continues full steam ahead instead if exiting.

I solved this by writing a apt_install() function, which will exit if it fails, and calling that everywhere ```apt install -y``` is called.

Though maybe just a ```|| exit 1``` is all you really need.

nyxnor:
Make a merge request so we can see.
For 1 package to be found would require people running on different systems.

harvhat:
Sorry nyxnor I'm not sure which code you want to see, or even what you mean about finding one package on different systems.

harvhat:
Ok @nyxnor I made a pull request for the apt_install() function bit (#3002) and the "added an option to force a particular baseimage" bits are in the setting-base-os branch of my fork. I don't plan on making a PR for that as I think OS detection needs a more major overhaul anyways. But it's there if you want to see.

nyxnor:
> However there is no package called "armbian-config" so the whole 'apt install' line fails while the build script continues on anyway, unsuccessfully of course.

When using armbian, there is no package called armbian-config, maybe we should update that package name to the new one.

[...]

PR #3002, opened 2022-02-27
Wrap apt install -y into a function and exit if it fails

harvhat:
Exit when apt fails to install packages, so the user can see the error message while still on the screen.
Otherwise the script rolls on only to fail at the end.


nyxnor:
This function fails because it will quit if any package failed to install independent of the base os.
E.g.: base os armbian should install armbian-config, others shouldn’t. If running on other system, the script will quit.

build_sdcard.sh
```diff
@@ -291,7 +301,7 @@ general_utils="policykit-1 htop git curl bash-completion vim jq dphys-swapfile b
 python_dependencies="python3-venv python3-dev python3-wheel python3-jinja2 python3-pip"
 server_utils="rsync net-tools xxd netcat openssh-client openssh-sftp-server sshpass psmisc ufw sqlite3"
 [ "${baseimage}" = "armbian" ] && armbian_dependencies="armbian-config" # add armbian-config
-sudo apt install -y ${general_utils} ${python_dependencies} ${server_utils} ${armbian_dependencies}
+sudo apt_install ${general_utils} ${python_dependencies} ${server_utils} ${armbian_dependencies}
```
nyxnor: If using raspbian or debian or whatever else, the script will quit here.

Better to do a separate call of apt_install for different base os dependencies

build_sdcard.sh
```diff
@@ -335,9 +345,10 @@ fi
 
 # special prepare when Raspbian
 if [ "${baseimage}" = "raspios_arm64" ] || [ "${baseimage}" = "debian_rpi64" ]; then
-
-  echo -e "\n*** PREPARE RASPBERRY OS VARIANTS ***"
-  sudo apt install -y raspi-config
+    
+  echo -e "\n*** PREPARE RASPBERRY OS VARIANTS ***"    
+  #sudo apt install -y raspi-config
+  apt_install raspi-config
```
nyxnor, line 354: Same

harvhat:
The script SHOULD fail there, this is the point.  As it stands now the script fails at the end instead, which is *much* less helpful.
Maybe read the issue #2995 again to understand the problem.

build_sdcard.sh
harvhat: this bit is only running for RASPBERRY OS VARIANTS
If raspi-config isn't there this it isn't a raspberrypi variant. something is very wrong so it should fail right here, instead of at the end like it does.

build_sdcard.sh
harvhat: "armbian-config"  is only added if ```["${baseimage}" = "armbian"]``` so if baseimage isn't really ambian is *should* fail right here. This is what is needed, waiting until the end of the script to fail is not very helpful.

harvhat:
TL;DR version is if the script incorrectly detects the OS and tries to install the wrong packages you want it to fail as soon as possible, not at the end. This is the goal of this patch.

The whole OS detection code is not very robust and probably needs a major refactoring.

build_sdcard.sh
nyxnor: concern solved

build_sdcard.sh
nyxnor: concern solved

rootzoll:
Thanks. Marked for v1.8 release.

PR #3019, opened 2022-03-06
Remove white spaces

openoms:
Only cosmetic changes, should not change any functionality, but PR-d to `dev` as many files are affected.

Was simply using:
```
find . -type f -name '*.py' -exec sed --in-place 's/[[:space:]]\+$//' {} \+
find . -type f -name '*.sh' -exec sed --in-place 's/[[:space:]]\+$//' {} \+
```


rootzoll:
We have quite some PRs waiting for v1.8 ... adding such massive cosmetic changes might break lots of them. Not sure its worse to do.

openoms:
Not a priority indeed, Will draft it indefinitely and wait for a time with less conflicts.

YBHQ:
#

openoms:
Going forward we should encourag to use shellcheck and be stricter about formatting. Maybe a formatting github action could help all new PRs.

PR #3036, opened 2022-03-22
chantools update to v0.10.3

guggero:
Bumps the version of `chantools` to v0.10.3 to fix a bug in the key derivation that lead to keys sometimes (with a chance of 1:256) not matching `lnd`'s keys.

Unfortunately I cannot test the install and signature verification of the latest version, as I don't have a RaspiBlitz running at the moment. But I didn't change anything with respect to PGP keys or how the signature is created, so everything should still work the same as in `v0.10.1`.


guggero:
Actually, there still seems to be a problem with the key derivation, even in this version. Will fix first and re-open the PR later.

openoms:
updating to v0.10.4 in https://github.com/rootzoll/raspiblitz/pull/3095

PR #3038, opened 2022-03-23
chantools update to v0.10.4

guggero:
Bumps the version of `chantools` to v0.10.4 to fix a bug in the key derivation that lead to keys sometimes (with a chance of 1:256) not matching `lnd`'s keys.

Unfortunately I cannot test the install and signature verification of the latest version, as I don't have a RaspiBlitz running at the moment. But I didn't change anything with respect to PGP keys or how the signature is created, so everything should still work the same as in `v0.10.1`.


rootzoll:
Marked for v1.8 release

openoms:
This can be closed as updated channel tools in https://github.com/rootzoll/raspiblitz/pull/3095 in the `dev` branch

openoms:
Also verification was working ok. Thank you for the update @guggero!

PR #3057, opened 2022-04-05
Fix for broken link

ktecho:
It seems the title of that section got updated, but the link wasn't


rootzoll:
Marked for v1.8 release.

PR #3063, opened 2022-04-10
add thousands separator on balances

vv01f:
simply for better readability, assuming the numbers are used for display only, not for calculations.


rootzoll:
Will include for v1.8 release ... can you change the base for this PR to `dev` branch?

PR #3070, opened 2022-04-23
Fix missing brace in UPS section

HiLivin:
Add a missing curly brace which potentially affected `00infoBlitz.sh` script result.


rootzoll:
Good catch thanks .. merging for v1.8 release.

PR #3074, opened 2022-04-26
Fix passing UPS battery info between scripts

HiLivin:
The background scan sourced the value of `$upsBattery` from the result of `blitz.ups.sh status` but didn't pass it further to the cache (from which the `00infoBlitz.sh` could look up the value).
Attached commit fixes this and enables proper display of voltage/capacity on the status screen.


rootzoll:
Thanks - this slipped on implementing cache info. Will be part of next release.

PR #3079, opened 2022-04-30
Update README.md

davisv7:
Links for basic setup were broken.


rootzoll:
@nyxnor is this an edge case that is not correctly handled by the automated TOC generation?

rootzoll:
@davisv7 thx - merging for the v1.7 branch

nyxnor:
> is this an edge case that is not correctly handled by the automated TOC generation?

Yes, its the dot in the link which should be removed.

PR #3082, opened 2022-05-01
Move color definitions above their use cases

HiLivin:
After this fix the UPS status color is displayed as intended.


rootzoll:
Makes sense to define the colors first ;) thx, merged for v1.8 release

Issue #3054, opened 2022-04-04
CLN 0.11.2 update


openoms:
The v0.11.0rc1 is about to be tagged so good to start testing: https://github.com/ElementsProject/lightning/pull/5155

openoms:
v0.11.0rc2 installed ok on running RPi and and64 as well with:
```
config.scripts/cl.install.sh update v0.11.0rc2
```

- [x] The install script will need to be refactored due to poetry being used to install the python dependencies now: https://github.com/ElementsProject/lightning/blob/master/doc/INSTALL.md#to-build-on-ubuntu

openoms:
the update to the final v0.11.0.1 also works well.

- [ ] Will also need to expose the new GRPC interface and
- [ ] Expose Runes (likely a future method to connect Zeus too)

rootzoll:
@fusion44 is GRPC/Runes also something to use with the API/library?

fusion44:
GRPC definitely. If I knew the GRPC API would come so soon, I'd have waited with the cln implementation. Saves a lot of time compared to the current rpc API.

openoms:
Note the rebrand of C-lightning to Core Lightning  (CLN) https://blog.blockstream.com/en-c-lightning-is-now-core-lightning/  also for the webgui @fusion44 
- [x] will change the naming in the ssh menu

rootzoll:
New CLN install does not run thru on sd card build (based on new RaspiOS).  @openoms can you see why `mrkd` is missing here and how to make it available?

```
- Install to /usr/local/bin/
mkdir -p /usr/local/bin
mkdir -p /usr/local/libexec/c-lightning
mkdir -p /usr/local/libexec/c-lightning/plugins
mkdir -p /usr/local/share/man/man1
mkdir -p /usr/local/share/man/man5
mkdir -p /usr/local/share/man/man7
mkdir -p /usr/local/share/man/man8
mkdir -p /usr/local/share/doc/c-lightning
cp tools/hsmtool tools/lightning-hsmtool
install cli/lightning-cli lightningd/lightningd tools/lightning-hsmtool /usr/local/bin
install lightningd/lightning_channeld lightningd/lightning_closingd lightningd/lightning_connectd lightningd/lightning_dualopend lightningd/lightning_gossipd lightningd/lightning_hsmd lightningd/lightning_onchaind lightningd/lightning_openingd lightningd/lightning_websocketd /usr/local/libexec/c-lightning
[...]

PR #3086, opened 2022-05-02
CLN update to v0.11.0.1, apply poetry install, rebrand to CLN

openoms:
will change C-lightning to Core Lightning / CLN as part of the rebrand

the install needs testing


rootzoll:
Basically looks good. Will merge into dev ... needs to be tested in context with sd card build and other updates anyway.

openoms:
related #3054

Issue #1058, opened 2020-02-29
i2c /dev entries driver


Lockejoe:
Hello all,

Hope someone can help me.
I tried to update my Raspiblitz from 1.3 to 1.4.

Now I got following Issue with the new image, while booting.
_i2c dev entries driver_

System stuck at this point.
I tried different sd cards, tried different images, downloads, nothing helped.

Any ideas how to solve it?
Also tried to google, but I couldn't find a solution, which fits for me. (I'm also not that familiar with linux, especially with kernel problems.)

Thanks in advance

rootzoll:
The "i2c dev entries driver" is your last message on the LCD, correct?
Can you still ssh in? There are some reports that the LCD is just freezing on boot, but the rest works. Also do you have the ShimFan? Ist it also not booting if you remove the shimfan?

Lockejoe:
Hello,
Thanks for your answer.
Yes, this is my last message, but I do not have a LCD, I also do not have a ShimFan.
I only use with ssh.
After the update it wasn‘t possible to connect via ssh, neither with wifi nor with ethernet.
So I cecked with a „normal“ monitor to find out the root cause.
And it always stuck i2c /dev entries driver, as last message.

openoms:
@Lockejoe simply need to unplug the monitor from the HDMI and boot without it. Login via ssh and follow the guide.
During the initial setup using the HDMI freezes the boot process. Later it will be possible to use. Look for HDMI in the issues to get the script helping you to change to that from the default GPIO LCD output.

Lockejoe:
Hello @openoms 
Thanks for your answer. Ok maybe my first explanation was not that good.
What I did:
1. Updated my sd card (tried with different sd cards)
2. started my rasperry pi (plugged in) - there was no display connected, only my hdd and ethernet (tried also wifi, because this was my privious setup)
3. waited for ca. 10-15 min (and also trying with known ips to ssh...)
4. checked my network to identify correct ip (if the old ip is still correct) - wondering, that I couldn‘t find an active one for the raspberry pi
5. connected a monitor to identify the root cause why couldn‘t ssh the pi

Hope this helps

ZukoTFL:
[...]

PR #3087, opened 2022-05-03
Add support for X708 UPS HAT

HiLivin:
This adds a possibility to install X708 scripts with `blitz.ups.sh on x708`.

Succesfully tested the code on the `dev` branch with Raspberry Pi 4B+ and  X708 v1.2 UPS HAT.

- [x] The [`blitz.shutdown.sh`](https://github.com/rootzoll/raspiblitz/blob/v1.7/home.admin/config.scripts/blitz.shutdown.sh) is executed before cutting off the power to the Raspberry Pi

- [x] The I2C is being enabled to allow getting data from the on-board battery management chip. I haven't found any issues as referenced by #1058 

- [x] The battery voltage and AC state are correctly fetched by [`_background.scan.sh`](https://github.com/rootzoll/raspiblitz/blob/v1.7/home.admin/_background.scan.sh) and later displayed on the status window

- [x] Still, I'd like to know if it's alright to link to my repo for downloading the scripts or maybe it would be better to pack them and put somewhere along with other RaspiBlitz code.

### DIFF
--- a/home.admin/config.scripts/blitz.ups.sh
+++ b/home.admin/config.scripts/blitz.ups.sh
@@ -7,6 +7,7 @@ source /mnt/hdd/raspiblitz.conf 2>/dev/null
 if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
  echo "Configure a UPS (Uninterruptible Power Supply)"
  echo "blitz.ups.sh on apcusb"
+ echo "blitz.ups.sh on x708"
  echo "blitz.ups.sh status"
  echo "blitz.ups.sh off"
  exit 1
@@ -21,7 +22,7 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
   echo "Turn ON: UPS"
 
   if [ "$2" = "apcusb" ]; then
-   
+
     # MODEL: APC with USB connection
     # see video: https://www.youtube.com/watch?v=6UrknowJ12o
 
@@ -48,12 +49,36 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
     sudo sed -i "s/^WALL=.*/#WALL=wall/g" /etc/apcupsd/apccontrol
     sudo systemctl enable apcupsd
     sudo systemctl start apcupsd
-    
+
     # set ups config value (in case of update)
     /home/admin/config.scripts/blitz.conf.sh set ups "apcusb"
 
     echo "OK - UPS is now connected"
     echo "Check status/connection with command: apcaccess"
+
+  elif [ "$2" = "x708" ]; then
+
+    # MODEL: Geekworm/Suptronics X708 v1.2
+
+    # enable I2C interface
+    sudo raspi-config nonint do_i2c 0
+
+    # install prerequisites
+    sudo apt-get install -y python3-smbus i2c-tools
+
+    # clone git repository with the X708 scripts
+    git clone https://github.com/HiLivin/x708blitz.git /home/admin/x708blitz
+    cd /home/admin/x708blitz
+    git checkout 2ed6caffedef3bd523732287234ebe9d200fd92f
+
+    # run install script
+    sudo bash /home/admin/x708blitz/install.sh
+
+    # set ups config value (in case of update)
+    /home/admin/config.scripts/blitz.conf.sh set ups "x708"
+
+    echo "OK - X708 UPS HAT is now connected"
+    echo "Please, perform restart to apply changes."
 
   else
     echo "FAIL: unknown or missing second parameter 'UPSTYPE'"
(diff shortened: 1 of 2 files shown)


rootzoll:
Thats an interesting integration. I see if I can get a hands on those boards to test your integration. I added it to the v1.8.2 milestone for possible merge .. a bit up the line because v1.8.0 & 1.8.1 wil be very WebUI focused.

> Still, I'd like to know if it's alright to link to my repo for downloading the scripts or maybe it would be better to pack them and put somewhere along with other RaspiBlitz code.

You can keep things in your repo and install it on demand, but then after cloning the repo within RaspiBlitz script it should checkout to a certain commit hash for security reasons.

HiLivin:
It would be great to have someone double-check the whole thing. 

- [x] I've added the `git checkout` with a proper hash to the install script and squashed the commits

rootzoll:
Progress looks great. If you think you have everything in there you can remove `[WIP]` from title and mark it with the tag `final testing`. Targeting here for release with 1.8.1

HiLivin:
Done changing title. I think only repo maintainers can add tags/labels.

HiLivin:
Bumped the repo version to checkout. A major improvement was added.

> Normally the x708 board waits only for ~50 seconds for the Pi to shutdown. If Pi shutdowns later than that, then the UPS hat doesn't turn itself off and the power to the Pi isn't cut. In case of power loss it wouldn't be a much problem - the board would eventually turn off because of low battery.

Now, to do it right, when the x708 starts waiting for Pi to shutdown a background loop is started with a specific timeout after which the UPS is triggered to wait for Pi shutdown again. This way the RaspiBlitz shutdown script can take very long and the hat will properly cut off the power afterwards.

HiLivin:
Bumped again.
The install script now uses `patch` on `00mainMenu.sh` instead of changing code lines with `sed` - this will improve compatibility with future versions of menu script.

The functionality seems fully developed. It would be nice to get some feedback from others using this UPS board with the scripts.

rootzoll:
@HiLivin if you prepare a tweet asking for broader feedback on this feature (Call for Testers) .. I am happy to share.

holgern:
I have installed the X708 V2.0 and I have applied this PR to 1.7.2. After reboot, it is doing now every time a shutdown.
I have two 18650  (connected in parallel with the holder) and I power the x708 board through USB.

I will wait what will happen when the battery are fully loaded. (They are now at 25%)

holgern:
After reaching MIN_VOLTAGE = 4.0, my raspi could start.
sudo ./config.scripts/blitz.ups.sh status returns now:
upsStatus='ONLINE'
upsBattery='4.22V'

HiLivin:
Great! So that's a new board.

The automatic shutdown should be called when `battery voltage is low` AND `AC power loss is detected` (PLD pin high). 
There could possibly be some hardware issues with PLD when batteries are low and the power supply is inefficient - high voltage drop might be recognized as AC power fail.

home.admin/config.scripts/blitz.ups.sh
```diff
@@ -48,13 +49,37 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
...
+    # enable I2C interface
+    sudo raspi-config nonint do_i2c 0
+
+    # install prerequisites
+    sudo apt-get install -y python3-smbus i2c-tools
+
+    # clone git repository with the X708 scripts
+    git clone https://github.com/HiLivin/x708blitz.git /home/admin/x708blitz
+    cd /home/admin/x708blitz
+    git checkout ed3612617365ba485141b98c595bdba67be95804
```
holgern: This needs to change to the newest commit
```
git checkout 2ed6caffedef3bd523732287234ebe9d200fd92f
```

YBHQ:
#

home.admin/config.scripts/blitz.ups.sh
HiLivin: Done in 7782c1935f06c48ba6feb76e048547ed2d70bc60.

YBHQ:
~~~~

smamamo:
I have installed the X708 V2.0 on my raspiblitz V. 018.0 and all works . 
Auto start - Start when power cam back works as well 
many thanks

rootzoll:
Perfect - so I am merging to be included in v1.8.1 release.

smamamo:
A suggestion would be the integration of a simple monitoring of the battery charge. - x708.info.py 
And a simple adjustment of the fan speed. - x708.fan.py -

My battery setting

MIN_VOLTAGE = 3.7 # (volts) shutdown when voltage drops below this value

I optimised my fans at

HIGH_THRESHOLD = 56 # (degrees Celsius) Fan running at high speed at this temperature.
LOW_THRESHOLD = 50 # (degrees Celsius) Fan running at low speed at this temperature.
SLEEP_INTERVAL = 5 # (seconds) How often do we check the core temperature.

rootzoll:
@smamamo if you like to add furthe rimprovements, it best to start a new PR on it

PR #3089, opened 2022-05-03
Update bonus.helipad.sh

SpencerPearson:
Reflects the latest Helipad version


SpencerPearson:
(I will soon write a way to just auto fetch the latest version so I don't have to keep setting it manually 😅)

PR #3090, opened 2022-05-04
fix toc dot

nyxnor:
related to pull https://github.com/rootzoll/raspiblitz/pull/3079

the `.` dot was being passed as regex, not string. Now fixed.


rootzoll:
thx for the fix. merging for v1.8 release.

Issue #3055, opened 2022-04-04
BTCPay - NBXPlorer update needs database pinned or migrated to postgesql


openoms:
For now to keep the old database format it seems that need to add to the 
`Execstart=` line in the nbxplorer systemd service.
```
--nomigrateevts --nomigraterawtxs
```

https://github.com/dgarage/NBXplorer/blob/master/docs/Postgres-Migration.md

rootzoll:
When was this change introduced? Does this effect already people running the v1.7.2?

ghost:
This occurred for me when updating BTCPay from 1.7.2

openoms:
> When was this change introduced? Does this effect already people running the v1.7.2?

no, it is in the last few versions of NBXplorer, not in the one shipped with v1.7.2.

@Bucherbp for now you can reinstall BTCPay from SERVICES to reset to the pinned version.

ghost:
> @Bucherbp for now you can reinstall BTCPay from SERVICES to reset to the pinned version.

Yep, already done, thanks.

rootzoll:
As PR #3096 is mergd - closing this issue.

PR #3096, opened 2022-05-06
Update btcpayserver to 1.5.1

cercatrova21:
Tested, no issues so far


openoms:
Thanks. So there is the option to leave NBXplorer at the version it was on v1.7.2 and only update BTCpay.

So this can be merged. 
Maybe deactivate the updatiing nbxplorer also in the update part until this is solved: https://github.com/rootzoll/raspiblitz/issues/3055

Issue #2905, opened 2022-01-21
[Research] ZRAM to use compressed swap in memory (increase performance when memory usage is high)


openoms:
ZRAM is a ~3x compressed swapfile in memory. Helps to avoid using the swap on the disk which even with an SSD can lead to a complete halt when the computer is running out of memory.
Article:
https://haydenjames.io/raspberry-pi-performance-add-zram-kernel-parameters/
Generic install script:
https://github.com/foundObjects/zram-swap

Relevant for 1-2GB RAM models already now.

The best test would be to install ZRAM on a device which already reproducibly experiencing slowdowns and/or freezes and see if it makes a difference.

GoetzGoerisch:
RPI3, HDD during IBD:

`v1.7.2rc1`

**Before:**
 
```shell
admin@192.168.178.5: ₿ free -m
               total        used        free      shared  buff/cache   available
Mem:             959         213          40           1         706         687
Swap:           3137         349        2788
admin@192.168.178.5: ₿ sudo cat /proc/swaps 
Filename                                Type            Size            Used            Priority
/mnt/hdd/swapfile                       file            1984508         354540          -2
/dev/zram0                              partition       1228548         4452            15
admin@192.168.178.5: ₿ vmstat -s
       982840 K total memory
       211788 K used memory
       206780 K active memory
       629852 K inactive memory
        52120 K free memory
         8592 K buffer memory
       710340 K swap cache
      3213056 K total swap
       361296 K used swap
      2851760 K free swap
        77607 non-nice user cpu ticks
            2 nice user cpu ticks
        66127 system cpu ticks
       573418 idle cpu ticks
       251014 IO-wait cpu ticks
            0 IRQ cpu ticks
         7894 softirq cpu ticks
            0 stolen cpu ticks
     19807001 pages paged in
      1990289 pages paged out
        33186 pages swapped in
       107756 pages swapped out
     33451365 interrupts
      6807495 CPU context switches
   1642783597 boot time
        62686 forks
```
        
**After:**
``` shell
admin@192.168.178.5:~ ₿ free -m
               total        used        free      shared  buff/cache   available
Mem:             959         317          49           3         592         581
Swap:           3137         343        2794
admin@192.168.178.5:~ ₿ sudo cat /proc/swaps 
Filename                                Type            Size            Used            Priority
/mnt/hdd/swapfile                       file            1984508         0               -2
/dev/zram0                              partition       1228532         355328          15
[...]

PR #3102, opened 2022-05-10
add ZRAM

openoms:
Related: #2905, #2924


openoms:
ready to merge - ZRAM will be switched for all devices during the first boot in the _provision_.sh

PR #3109, opened 2022-05-11
add the cln-grpc plugin

openoms:
releated #3054

### DIFF
--- a/home.admin/config.scripts/cl-plugin.cln-grpc.sh
+++ b/home.admin/config.scripts/cl-plugin.cln-grpc.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+
+# command info
+if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ];then
+  echo
+  echo "Install the cln-grpc plugin for CLN"
+  echo "Usage:"
+  echo "cl-plugin.cln-grpc.sh [on|off] <testnet|mainnet|signet>"
+  echo
+  exit 1
+fi
+
+source <(/home/admin/config.scripts/network.aliases.sh getvars cl $2)
+
+function buildGRPCplugin() {
+  echo "- build the cln-grpc plugin"
+  if -f /home/bitcoin/cl-plugins-available/cln-grpc/debug/cln-grpc; then
+    cd /home/bitcoin/lightning/plugins/grpc-plugin
+    sudo -u bitcoin /home/bitcoin/.cargo/bin/cargo build \
+     --target-dir /home/bitcoin/cl-plugins-available/cln-grpc
+  fi
+}
+
+if [ "$1" = install ]; then
+  buildGRPCplugin
+
+elif [ "$1" = on ]; then
+  buildGRPCplugin
+  sudo ln -s /home/bitcoin/cl-plugins-available/cln-grpc/debug/cln-grpc \
+   /home/bitcoin/${netprefix}cl-plugins-enabled/
+  #blitz.conf.sh set [key] [value] [?conffile] <noquotes>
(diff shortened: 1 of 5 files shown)


home.admin/config.scripts/cl-plugin.cln-grpc.sh
```diff
@@ -0,0 +1,43 @@
...
+  sudo ln -s /home/bitcoin/cl-plugins-available/cln-grpc/debug/cln-grpc \
+   /home/bitcoin/${netprefix}cl-plugins-enabled/
+  #blitz.conf.sh set [key] [value] [?conffile] <noquotes>
+  /home/admin/config.scripts/blitz.conf.sh set grpc-port "7777" ${CLCONF} noquotes
+  /home/admin/config.scripts/blitz.conf.sh set ${netprefix}cln-grpc-port "7777"
+
+elif [ "$1" = off ]; then
+    sed -i "/^grpc-port/d" ${CLCONF}
+    rm -rf /home/bitcoin/${netprefix}cl-plugins-enabled/cln-grpc
+    /home/admin/config.scripts/blitz.conf.sh set ${netprefix}cln-grpc-port "off"
```
openoms, line 38: @rootzoll question: is there a benefit to use the port number here or should it be just 
`${netprefix}cln-grpc=on|off` in the `raspiblitz.conf`?

Also the port number is completely arbitrary, suggestions are welcome.
For example on the 4*3 phone keyboard GRPC is 4772.

home.admin/config.scripts/cl-plugin.cln-grpc.sh
```diff
@@ -0,0 +1,43 @@
...
+if [ "$1" = install ]; then
+  buildGRPCplugin
+
+elif [ "$1" = on ]; then
+  buildGRPCplugin
+  sudo ln -s /home/bitcoin/cl-plugins-available/cln-grpc/debug/cln-grpc \
+   /home/bitcoin/${netprefix}cl-plugins-enabled/
+  #blitz.conf.sh set [key] [value] [?conffile] <noquotes>
+  /home/admin/config.scripts/blitz.conf.sh set grpc-port "7777" ${CLCONF} noquotes
+  /home/admin/config.scripts/blitz.conf.sh set ${netprefix}cln-grpc-port "7777"
```
openoms: @rootzoll question: is there a benefit to use the port number here or should it be just 
`${netprefix}cln-grpc=on|off` in the `raspiblitz.conf` and in redis?

Also the port number is completely arbitrary, suggestions are welcome.
For example on the 4*3 phone keyboard GRPC is 4772.

openoms:
building the `grpc-plugin` takes a good few minutes. I suggest it should be part of the "fatpack" SDcard image and can be on by default just like the experimental-features of CLN.

home.admin/config.scripts/cl-plugin.cln-grpc.sh
rootzoll: I mean its nice to have the port numer as an option - the prestart script should take care of always setting it number. Only point is that its a bit more fuzzy as testing against "on" to check if the service shoudl be active. So now we would test if NOT "" - but that should be basically fine.

rootzoll: port numbner 4772 sounds good with me - nice idea with the GRPC phonekeyboard coding :)

rootzoll:
Please merge if you see this PR as done. Intense testing will be done as part of RC1

openoms:
The cln-grpc plugin will be the part of the image release as it will be built after CLN in the build_sdcard.sh. Meging to `dev` for testing.

PR #3113, opened 2022-05-11
fix electrs is ready for btc-rpc-explorer prestart

PatrickScheich:
After reboot btc-rpc-explorer service gets trapped in a reboot loop as the check if elelctrs is running and ready always fails.
`netstat -n` does not show connections in LISTEN state.  
`netstat -tan` shows all tcp connections `-t` in all states `-a` without resolving names `-n` to prvent dns spam


rootzoll:
thx - good catch

rootzoll:
merging for v1.8.0 release

PR #3031, opened 2022-03-18
Add Alby

secondl1ght:
I have added Alby as a tutorial connection option for RaspiBlitz users. Please let me know if you require any further information, thanks!

PR #3125, opened 2022-05-17
Use special bonus script to show connection details for Alby

bumi:
This PR builds on #3031

[Alby](https://getalby.com/) is a lightning app for desktop browsers and thus it is not possible for users to scan a lndconnect QR code. PR 3031 reused the current bonus.lndconnect.sh script which displays a QR code. This is not really usable for Alby. 
This new bonus.alby.sh script prints Alby connection details that can be copied and pasted into Alby.

Note: this is the first time I am writing some RaspiBlitz code, I am not sure how correct this is. 
I've tried to test the script by c&p it to my running node.

### DIFF
--- a/home.admin/99connectMenu.sh
+++ b/home.admin/99connectMenu.sh
@@ -70,18 +70,7 @@ case $CHOICE in
     exit 0;;
 
   ALBY)
-      websiteLink="https://getalby.com"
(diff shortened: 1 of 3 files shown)


rootzoll:
thx :) merged for v1.8.0 release

PR #3141, opened 2022-05-23
cln-backup: use the version before poetry

openoms:
fix for new installs on v1.7 as the poetry installation breaks the old method

Related issue: https://github.com/lightningd/plugins/issues/363


openoms:
tested ok, merging and cherry-picking to dev

PR #3160, opened 2022-06-09
Adapt to new migration for umbrel 0.5.0 including core lightning data

### LINKED ISSUE #3159: Check Umbrel migration for new 0.5.0 version
https://blog.getumbrel.com/introducing-umbrel-0-5-a-beautiful-personal-server-os-for-self-hosting-dcbee7e23b64

Also now that they have also core lightning - this needs to be part of the migration process, too.

rootzoll:
fixes #3159


rootzoll:
@openoms I works that I can migrate the core lightning data - but their hsm_secret is not encrypted as far as i can see, should we leave this as default after migration?

openoms:
Yes, we are not encrypting it by default either.

PR #3177, opened 2022-06-21
remove TODO in 00infoLCD

woehrer12:


rootzoll:
basically a good catch - but this comment was made when refactoring of the touchscreen was still considered to be part of v1.8.0 - but now git pushed up to 1.8.3. So pushing this PR to v1.8.3 milestone (will take a bit before merge)

PR #3183, opened 2022-06-23
set LANG=en_US.utf8 when using grep 'Good signature'

openoms:
avoid the edge cases when `grep 'Good signature'` fails in non-English environments

Related discussion: https://github.com/bitcoin/bitcoin/issues/25395#issuecomment-1163594630


openoms:
As a next step we should move to exit code based decisions everywhere, but for now this applies to all where `grep 'Good signature'` is used

PR #3188, opened 2022-06-24
reduce lnd workers to number of cores if -lt 8

### LINKED ISSUE #3187: Reduce load by limiting LND to use 4 cores only
New options in the lnd.conf to be tweaked accrodingly:
```
[workers]

; Maximum number of concurrent read pool workers. This number should be
; proportional to the number of peers. (default: 100)
; workers.read=200

; Maximum number of concurrent write pool workers. This number should be
; proportional to the number of CPUs on the host. (default: 8)
; workers.write=8

; Maximum number of concurrent sig pool workers. This number should be
; proportional to the number of CPUs on the host. (default: 8)
; workers.sig=4
```
https://github.com/lightningnetwork/lnd/blob/master/sample-lnd.conf#L1139

Eg for the RPi4 with 4 cores:

```
[workers]
workers.write=4
workers.sig=4
```

Related: https://github.com/fusion44/blitz_api/issues/126

openoms:
fix #3187

### DIFF
--- a/home.admin/config.scripts/lnd.check.sh
+++ b/home.admin/config.scripts/lnd.check.sh
@@ -208,6 +208,31 @@ if [ "$1" == "prestart" ]; then
   setting ${lndConfFile} ${insertLine} "db.bolt.auto-compact-min-age" "672h"
   setting ${lndConfFile} ${insertLine} "db.bolt.auto-compact" "true"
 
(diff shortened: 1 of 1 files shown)


openoms:
tested ok on an RPi4, ready to merge

PR #3198, opened 2022-06-29
stackingsats: allow more than one config, also complementing a base config

vv01f:
enables usage of e.g.
* a base config aside one or more alternative complementing ones
* different stacking amounts
* different accounts (such as friends and family or business and private)
* test config aside a productive one

simply via 2nd cli parameter for `stacksats.sh` and more config files in `/mnt/hdd/app-data/stacking-sats-kraken` that look the same as `.env` but must be named `.env-param2` when calling e.g. `stacksats.sh "stack" "param2"`


rootzoll:
Thanks :) Will be part of v1.8.0rc3 for testing.

PR #3226, opened 2022-07-13
Fixed x64 checksum

ktecho:
Tried creating SD card from dev and checksum for x64 doesn't match. I checked the other 2 files for other architectures and they seem ok.

Please, take the correct one in this PR


ktecho:
I tried it pointing the install script to my "raspiblitz" repo and it work flawlessly.

rootzoll:
Good catch. Thx :)

PR #3228, opened 2022-07-13
Adding fdisk to dependencies

ktecho:
While installing Raspiblitz on Ubuntu 22.04, it seems like fdisk is needed to prepare the data disk.

I'm thinking if this could be moved to other place, near the use of fdisk, because it's really not needed to create the SD and it could not be used if you're using a backup of the blockchain data... What do you think?


rootzoll:
Makes sense to have it there - thanks. Merging for v1.8.1 release.

PR #3239, opened 2022-07-21
don't delete old rescue files

openoms:
After going through a few recovery scenarios I think it is safer to keep all backups. 
The risk of filling up the SDcard is much less then losing potentially valuable data.
On download the rescue-files are identified with the md5sum, so should not cause any confusion:

https://github.com/rootzoll/raspiblitz/blob/cd2867d3f5c89ff54d054723a583f27e5d959d9e/home.admin/config.scripts/cl.backup.sh#L166

https://github.com/rootzoll/raspiblitz/blob/cd2867d3f5c89ff54d054723a583f27e5d959d9e/home.admin/config.scripts/lnd.backup.sh#L174


rootzoll (approved):

rootzoll:
Also merged in v1.8

PR #3293, opened 2022-08-15
RTL v0.13.0 update

openoms:
tested ok updating to v0.13.0 on CLN - it is ok since only changes the binary an restarts the default instance
ignores RC-s


openoms:
could be merged to dev and v1.8, solving #3299 also

YBHQ:
~~~~

PR #3310, opened 2022-09-05
working prototype of boltcard

dipunm:
This is currently a draft.

A PoC boltcard service for programming your boltcard to pay from your own server.

TODO:
- What should DB Password be (PasswordB?)
- Allow customizing spend allowance for boltcard instances
- Remove DB when turning service off or avoid dropping DB when re-installing boltcard?
- Allow customizing `AES_DECRYPT_KEY` or generate a random one?
- Implement `menu` and `update`

This is a request for comment as this is sort of large in scope.

Relevant PRs:
- https://github.com/boltcard/boltcard/pull/13
- https://github.com/breez/breezmobile/issues/977


rootzoll:
> What should DB Password be (PasswordB?)

Password B is always a good choice. We started in general to add a prestart script to the systemd-services of additional apps. That script will then be executed before every (re)start of the app and makes sure the config is uptodate - so for example it will always set the lastest passwordB even if it got changed.

To see an example of a prestart script see --> https://github.com/rootzoll/raspiblitz/blob/dc7f5f36453e58fc75bacdc53c626edaebfd87c1/home.admin/config.scripts/bonus.template.sh#L358  

Also in that `bonus.template.sh` you will find further infos on how to best add a new bonus script to raspiblitz.

dipunm:
This is cool... I think I'll port over my code to this template :)

dipunm:
@rootzoll I noticed that this project uses postgres but postgres is installed within the bonus.btcpayserver.sh file.

Is it a good idea to extract the postgres installation into its own bonus file or is there a better way to get both files to depend on postgres without conflicting?

AFAIK, both will work on the same version (looks like we use whatever is the latest version)

dipunm:
I am going to create a bonus.postgres.sh file as part of this PR. It will have only `on` and `off` commands and the off command will be discouraged with a warning. Even though I have tested uninstalling and re-installing postgres and the data stays in tact, according to the docs, if postgres updates a major version, then data will likely be lost: https://www.postgresql.org/docs/9.0/install-upgrading.html.

Now I am wondering about macaroon files, BoltCard suggests creating a single purpose macaroon that can only do `uri:/routerrpc.Router/SendPaymentV2`. There are already so many existing macaroons, but some of these I can't for the life of me find out what they do like `lndwalletkit` and `lndsigner`. If any of these are basically the same, then I might as well symlink or copy it but if not, then it looks like this will be the first bonus app to use the `bakemacaroon` command directly. There is a bunch of code that copies macaroons to different locations and syncs them (eg. lnd.credentials.sh) so I am looking for advice here as to what script files I should and should not be modifying and which folder I should be creating my macaroon into.

home.admin/00settingsMenuServices.sh
```diff
@@ -202,6 +204,29 @@ else
   echo "Specter Desktop Setting unchanged."
 fi
 
+# Boltcard process choice
+choice="off"; check=$(echo "${CHOICES}" | grep -c "s")
```
rootzoll: Its testing here for `s` while Boltcard is above added as `z` - seems not consistent

rootzoll:
> Is it a good idea to extract the postgres installation into its own bonus file or is there a better way to get both files to depend on postgres without conflicting?

Yes that makes sense and looks good so far

rootzoll:
about the macaroons I thik it makes most sense that you bake them yourself .. will give you more freedoms in the future with spending limits, etc.

rootzoll:
Also let me know how far you see this PR .. do you think it will be ready for v1.8.1 end of october?

home.admin/00settingsMenuServices.sh
dipunm: That'll probably be why it isn't working 😁

dipunm:
I have it mostly working, I want to see Breez get this released first so I can test the full e2e flow https://github.com/breez/breezmobile/issues/977

Also, boltcard has already updated with a new feature which I don't understand fully yet, but there is no migration script. I think I'll descope that for this pr, but I think I'll probably make a migration script for them too later.

Honestly, I think end of October is achievable but it mostly depends on Breez which depends on fiatjaf releasing a new version of their library (the fix is already in master)

dipunm:
I'm happy for this to go into the next next release if it helps you plan, better.

dipunm:
regarding the macaroons, I don't think there are examples of this outside of things like loop/lit (which I guess is being done within the loop executable). -- which by the way uses its access to the admin macaroon to do it.

I also noticed that there are macaroons created in multiple locations, and there is a script to regenerate them etc.

Basically, I feel like I need a bit of guidance on where to create copies of a new macaroon, and how to ensure I am covering all the edge cases.

rootzoll:
About the macaroons - you should always fin the latest in this directory: 
Bildschirmfoto 2022-10-06 um 16 52 11

Insgtead of one time  copying zthem in install you should either "link" (ln unix command) and add the user of your app the fitting unix group (e.g. grozp `lndreadonly` to be able to acces the `readonly.macaroon') OR you copy the needed macaroon everytime on systemd start with the prestart script (your app user also needs to be part of the fitting group so that the prestart script can accesss the macaroon for copy). Hope that helps.

dipunm:
Just a heads up, my SSD blew up, like, I swear it just spontaneously became a decorative item that just heats up when plugged in and does nothing more. I happened to be away at the time (of course).

Anyway, good news is that the change to Breez is getting released soon. Bad news is I can't really test it myself until I get a new SSD sorted and recover everything (Here's hoping I didn't have much in zombie channels).

I doubt this will reach the next release unless someone wants to help with testing.

tlindi:
Breez seems now done their part. Hows your progress?

dipunm:
Hi @tlindi, unfortunately I am still without a lightning node right now (need to buy a new SSD).

I will do so early next year and continue.

This PR does have functioning code though.

If you'd like to help by testing, then I can try to complete this with your guidance.

tlindi:
I could try to test. But I'm not actually coder/dev. So getting this running my raspiblitz can be PITA to me. But I'm interested. Unfortunately I yet have no 424 NFC cards. And I'd like to delay testing until it can be done on raspiblitz 1.9. Somewhere in the begin of year 2023 it could be the time...

rootzoll:
Now that LNbits & BTCpay provide BOLT card integration we think it should be recommended to use those implementaions with the raspiblitz. Closing PR for now.

Issue #3230, opened 2022-07-14
[Feature Request] Make installation and configuration of LNDg possible


md-weber:
I would love to have the possibility to install and configure LNDg right from the main menu. 

https://github.com/cryptosharks131/lndg

openoms:
There is a PR, but not-functional currently: https://github.com/rootzoll/raspiblitz/pull/3153
Might need a new one.

mark0st83:
I've installed lndg with 1.7.2 and a virtual setting on debian (proxmox). I've used @openoms [documentation ](https://github.com/openoms/lightning-node-management/blob/en/hardware/raspiblitz/lndg.md)for this.
I had to change a few things in the nginx.sh and a few commands from @openoms 

But it works

allyourbankarebelongtous:
I'm working on this now, see https://github.com/allyourbankarebelongtous/raspiblitz/tree/integrate_lndg. Still hasn't been tested yet, should get to that this weekend. After successful testing I'll submit a PR.

rootzoll:
PR #3313 is candidate to be merged in v1.9.0

rootzoll:
Merging #3313 into feature branch `3313-lndg` to test and finish last smal things.

rootzoll:
- [x] Make sure to still fix this: https://github.com/rootzoll/raspiblitz/pull/3313#discussion_r1024539631

rootzoll:
OK recovery test worked - merging `3313-lndg` branch. And to be final tested during RC phase.

Issue #3003, opened 2022-02-28
letsencrypt script syntax error


allyourbankarebelongtous:
Background: upgraded to 1.7.2 from 1.7.1, re-enabled vpn and hybrid mode, and was trying to redo letsencrypt for btcpayserver. Every time I tried to enable letsencrypt it errored out and returned to the raspiblitz main screen very quickly. A screenshot of the error showed:

'# running: python /home/admin/config.scripts/blitz.subscriptions.letsencrypt.py create-ssh-dialog
'   File "/home/admin/config.scripts/blitz.subscriptions.letsencrypt.py", line 70
'       print(*args, file-sys.stderr, **kwargs)
'                ^
'SyntaxError: invalid syntax
'# redis-cli del ln_default_locked:out
'(integer) 0

There was more but the screenshot didn't capture it. The problem is that calling the letsencrypt.py script using python didn't default to using python3 and the script has python3 specific functions in it (print as a function, for one). I exited to the command line and ran the same command that is run from the gui but with python3 instead of python and it worked fine. Recommend in addition to separating letsencrypt tls.cert and tls.key from the other nginx services which are not using letsencrypt (as described in issue #2980) calling the letsencrypt script via python3 explicitly.

rootzoll:
just checked on my 1.7.2 testnode: `python --version` --> `Python 3.9.2`

Will add to v1.8 milestone to recheck the letsencrypt routine.

rootzoll:
Checked v1.8.0 - also python is python defaults to 3.9.2 ... pushing it to 1.8.1 to check again in combination with: #2247

rootzoll:
With #4579 the letsencrypt ACME was updated and tested. Will be part of v1.11.1rc1 ... reopen if problem persist after that release.

PR #3313, opened 2022-09-10
Integrate LNDg to Raspiblitz

allyourbankarebelongtous:
Integrates LNDg (popular auto-rebalancer and auto-fee setting node manager) to the Raspiblitz via the SERVICES menu. bonus.lndg.sh runs a complete install of LNDg, including an nginx server and systemd backend services. bonus.lndg.sh along with included .conf files in /home/admin/assets/nginx/sites-available/ includes support for ssl and ssl over tor similar to other Raspiblitz install scripts. bonus.lndg.sh also moves the LNDg database to /mnt/hdd/app-data/lndg and simlinks it for a persistent database over updates, and _provision_.sh combined with bonus.lndg.sh ensures that the update will not change LNDg. bonus.lndg.sh automatically includes tor if tor is set in raspiblitz.conf. bonus.lndg.sh installs LNDg in /home/lndg/lndg under user lndg. Also includes the ability to update LNDg from the UPDATE menu, and includes an INFO menu when LNDg is installed, similar to other Raspiblitz services.

Fix Required: Currently changes the way python is called, resulting in other scripts breaking (see comment below). Working on fixing this. -- FIXED

TODO: 
Future: Integrate signature check if NGU ever starts signing his LNDg updates. 
Future: Enable bonus.lndg.sh to update to a particular version instead of just the latest version.
Working on: Give users the option of keeping or deleting database (currently always keeps database and always uses it on next install). --DONE
Working on: Integrate into blitzapi
Working on: Allow users to pick username on install


allyourbankarebelongtous:
After further checking, there seems to be a problem with the install. The install breaks other instances on the Raspiblitz that call scripts using "python" instead of "python3". See [issue 3003](https://github.com/rootzoll/raspiblitz/issues/3003). For example, my blitzapi.service didn't work until I edited the .service file from "/usr/bin/python -m uvicorn app.main:app --port 11111 --host=0.0.0.0 --root-path /api" to "/usr/bin/python3 -m uvicorn app.main:app --port 11111 --host=0.0.0.0 --root-path /api". It seems that the method for installing dependencies for the uwsgi server makes "python" not automatically python3. This is the same issue as issue 3003 (I had installed lndg at the time using a similar method). Help would be appreciated.

Example:
 ₿ python --version
Python 2.7.18
 ₿ python3 --version
Python 3.9.2

From my Raspiblitz after install

rootzoll:
Thanks for the PR - marked to be part of v1.8.1 release.

FYI best to base PRs on `dev` branch to keep up with other changed happening.

rootzoll:
Note: Linked to issue #3230

allyourbankarebelongtous:
Will do. I found a fix for the python issue as well. Switched from uwsgi to gunicorn for app server, runs only on python 3 with no dependencies. I'll recommit using dev branch with new changes.

allyourbankarebelongtous:
Ok, this commit fixes the issue with python by using gunicorn instead of uwsgi as the application server. Seems to be super stable with low overhead. Runs using systemd and installs with no dependencies. This commit also allows users to either delete or keep the database on uninstall, and allows users to use an existing database (if one exists) or delete it and restart from scratch on install. Tests good, feedback welcome.

MEdGalo:
Hi, I seems that after the installation of lndg  it the blocks the shutdown of LND when using the reboot or shutdown option from the raspiblitz menu. Could you please check if this indeed the case?

allyourbankarebelongtous:
I'll run some troubleshooting over the weekend if I can, but it does appear to do that. May require a modification to the shutdown/reboot scripts.

allyourbankarebelongtous:
New push fixing shutdown script. Checks if LNDg is installed and if so, shuts down services prior to shutdown of raspiblitz. Tests good from my end.

MEdGalo:
This fixed it for me as well.

MEdGalo:
My failed HTLCs page doesn't seem to update, does it work for you?

allyourbankarebelongtous:
> My failed HTLCs page doesn't seem to update, does it work for you?

Works for me, try `sudo systemctl status htlc-stream-lndg.service` and see what it says.

MEdGalo:
htlc-stream-lndg.service - Run HTLC Stream For Lndg
     Loaded: loaded (/etc/systemd/system/htlc-stream-lndg.service; enabled; vendor preset: enabled)
     Active: inactive (dead)

I started it now with sudo systemctl start  htlc-stream-lndg.service, but I don't know why it didn't start in the first place.

allyourbankarebelongtous:
Can you paste the content of the service file here?
`sudo nano /etc/systemd/system/htlc-stream-lndg.service` and copy/paste?

MEdGalo:
[Unit]
Description=Run HTLC Stream For Lndg
Requires=lnd.service
After=lnd.service
[Service]
User=lndg
Group=lndg
ExecStart=/bin/bash /home/lndg/lndg/htlc_stream.sh
StandardError=append:/var/log/lnd_htlc_stream_error.log
Restart=always
RestartSec=60s
[Install]
WantedBy=multi-user.target

allyourbankarebelongtous:
Looks good. I know the script starts it when it's installed and I haven't had mine not start again on boot. If it becomes a persistent issue let me know.

home.admin/config.scripts/bonus.lndg.sh
```diff
@@ -0,0 +1,380 @@
...
+    sudo ln -s /mnt/hdd/app-data/lnd/ /home/lndg/.lnd
+
+    # download and install
+    sudo -u lndg git clone https://github.com/cryptosharks131/lndg.git /home/lndg/lndg/
+    cd /home/lndg/lndg/ || exit 1
+    sudo apt install -y virtualenv
+    sudo -u lndg virtualenv -p python3 .venv
+    sudo -u lndg .venv/bin/pip install -r requirements.txt
+    PASSWORD_B=$(sudo cat /mnt/hdd/bitcoin/bitcoin.conf | grep rpcpassword | cut -c 13-)
+    sudo -u lndg .venv/bin/python initialize.py -pw $PASSWORD_B
```
rootzoll, line 146: Password B is set just during install .. this means that if the user changes the password later here it stays the old one. The simple way would be to change it in the password script - like other apps do it here: https://github.com/rootzoll/raspiblitz/blob/15b460098c4efc841a9ee86fdb5bc91c12361e34/home.admin/config.scripts/blitz.passwords.sh#L351 ... or a bit more complex do a prestart systemd script that on every start sets passwordB correct.

rootzoll:
So this really looks good - install tests ran fine :) Lets add it in the next release (called 1.9.0) for SSH install.

TODO: CHANGES.md & README.md (in the listing of available apps)

allyourbankarebelongtous:
Thanks for the feedback! Do you want me to update CHANGES.md and README.md? Anything else?

rootzoll:
CHANGES.md and README if you like - I can do too. The "passwordB" (see review above) would be a missing solution to implement.

allyourbankarebelongtous:
> CHANGES.md and README if you like - I can do too. The "passwordB" (see review above) would be a missing solution to implement.

PasswordB is already used. Don't see any issues with it in the reviews/comments, am I missing something?

I'll hack on CHANGES.md and README.md over the weekend.

rootzoll:
> PasswordB is already used. Don't see any issues with it in the reviews/comments, am I missing something?

Again what I mean is how to change the admin password to lndg by command line? Can I get a `bonus.lndg.sh setpassword [NEWPASSWORD]` which will set a new password after install?

Also I dont think we should offer to change the login username - like printed in info menu `To change username, login using default username and go to: http://192.168.178.61:8889/lndg-admin` .. later the WebUI wants to show the user info like: "Use usernam 'lndg-admin' and your PasswordB" .. but if the user can change the admin user name this is the wrong information than.

allyourbankarebelongtous:
> > PasswordB is already used. Don't see any issues with it in the reviews/comments, am I missing something?
> 
> Again what I mean is how to change the admin password to lndg by command line? Can I get a `bonus.lndg.sh setpassword [NEWPASSWORD]` which will set a new password after install?
> 
> Also I dont think we should offer to change the login username - like printed in info menu `To change username, login using default username and go to: http://192.168.178.61:8889/lndg-admin` .. later the WebUI wants to show the user info like: "Use usernam 'lndg-admin' and your PasswordB" .. but if the user can change the admin user name this is the wrong information than.

I'll edit the script to add a setpassword flag, no problem. I'll also edit the menu to just say "login with username lndg-admin and passwordB". New commit incoming this weekend!

Note: because lndg uses a Django interface, people will still be able to change their username and password by going to http://192.168.178.61:8889/lndg-admin, but the menu won't suggest it. I did try to find a way to get the username so the new username is always displayed by the menu but the username is only stored encrypted in the database, so I couldn't figure out how to do that.

allyourbankarebelongtous:
Ok, this last commit adds the ability to change the password using the script with `bonus.lndg.sh setpassword mynewpasswordB`. 

Note: To do this I had to create a new python file: the install script creates it when you install it, so if you have already installed and want to test this feature, either manually create the changepassword.py file or uninstall and reinstall LNDg using this updated script.

Also fixed the menu so it doesn't suggest changing the username and password via the Django web interface.

I don't have the most up to date CHANGES.md and README.md, so I'll defer to Rootzoll to update those, but I'm happy to provide inputs if needed.

Script tests good on my blitz.

allyourbankarebelongtous:
On further testing, I found that uninstalling and reinstalling using the same database resulted in an error due to an update that LNDg pushed. My database was v1.3.1, and LNDg was updated to v1.4.0, so when I reinstalled it installed v1.4.0 on top of an older database. My script didn't account for that, so I had to run the update script to migrate the changes. I added a migrate line to any install that uses an existing database, which should fix that.

I also modified the setpassword script to check for changepassword.py, and create it if it doesn't exist, making this work out of the box without an uninstall/reinstall to create the file. It still creates the file on install. 

Tests good.

allyourbankarebelongtous:
One more commit that fixes an issue that was discovered by another user attempting to update. I missed this because I uninstalled and reinstalled instead of updating through the update menu. 

This latest LNDg update made changes to settings.py which is auto-generated on initializing the app. This change makes it so that all updates will reinitialize settings.py in case future updates also require it.

Tests good.

rootzoll (approved):

rootzoll:
Ok will merge into `3313-lndg` feature branch and continue dev on this issue:
https://github.com/rootzoll/raspiblitz/issues/3230#issuecomment-1344516763

home.admin/config.scripts/bonus.lndg.sh
allyourbankarebelongtous: Is it too late for me to work it into the blitz.passwords.sh? I can submit a commit for that tonight if not.

rootzoll: All good - did the password change already in the branch `3313-lndg`.

allyourbankarebelongtous: Awesome, thanks! First PR ever for me.

Issue #3274, opened 2022-08-04
CLN v0.12.1 update


openoms:
This is to document an issues and changes needed related to the update.
Needed to use for example commando/runes and poncho (#3269)

update CLN to the latest rc found in https://github.com/ElementsProject/lightning/tags eg. 'v0.12.0rc1':
```
# upgrade pip first
sudo pip install --upgrade pip
# update CLN
config.scripts/cl.install.sh update v0.12.0rc1
```

BitromNO:
when running v0.12.0rc1 i cannot get it to run in clearnet, unchecked tor in the menu but it still advertises onion adress.

openoms:
Note that rc2 is now available.

>i cannot get it to run in clearnet, unchecked tor in the menu but it still advertises onion adress.

this is unrelated to the CLN version. You will need to set the cln config manually.
See https://lightning.readthedocs.io/TOR.html for the different options.

rootzoll:
done in other issue. closing.

PR #3319, opened 2022-09-13
CLN update to v0.12.1

openoms:
fixes: #3274 

the existing install process worked ok after setting the right keys


openoms:
CLN now at v0.12.1

can remove mako and mrkd dependencies

openoms:
Building v0.12.1 tested ok, ready to merge.

Issue #3324, opened 2022-09-21
Make sure Proxmox guide is linked in documentation


rootzoll:
https://cercatrova.blog/en/install-raspiblitz-on-proxmox/

cercatrova21:
I can copy the content to a md file here on github. I'll create under alternative.platforms a new folder called "Proxmox" and save the guide there. Ok?

rootzoll:
Sounds good with me, thx.

rootzoll:
merged and noted in CHANGES

rootzoll:
@cercatrova21 is there also an update strategy when a new version comes out? Build new VM with new image and then change external SSD to the new image?

cercatrova21:
Yes this would be my update strategy. Will test with the 1.8.1 release and add that to the guide

robinvonmises:
Not sure if anyone else has tested this yet, but I did and ran into the problem, that RTL is not working in my VM, anything else is.

At first I migrated the node through the ssd, RTL was installed. HTTP not working at all, HTTPS calling a bad gateway.

A fresh install did not let me install RTL by ssh menu. It failed, asking me to install it manually, which I did. This was leading to the errors above. Guess this is not priority at all, but maybe someone has a fast idea. 

EDIT: Maybe this causes the trouble, c-lightning REST is not able to start

```
Okt 04 10:38:39 RaspiBlitz systemd[1]: clrest.service: Scheduled restart job, restart counter is at 85.
Okt 04 10:38:39 RaspiBlitz systemd[1]: Stopped c-lightning-REST daemon for mainnet.
Okt 04 10:38:39 RaspiBlitz systemd[1]: Started c-lightning-REST daemon for mainnet.
Okt 04 10:38:39 RaspiBlitz node[1616508]: node:internal/modules/cjs/loader:936
Okt 04 10:38:39 RaspiBlitz node[1616508]:   throw err;
Okt 04 10:38:39 RaspiBlitz node[1616508]:   ^
Okt 04 10:38:39 RaspiBlitz node[1616508]: Error: Cannot find module 'express'
Okt 04 10:38:39 RaspiBlitz node[1616508]: Require stack:
Okt 04 10:38:39 RaspiBlitz node[1616508]: - /home/bitcoin/c-lightning-REST/bitcoin/app.js
Okt 04 10:38:39 RaspiBlitz node[1616508]: - /home/bitcoin/c-lightning-REST/bitcoin/cl-rest.js
[...]

PR #3328, opened 2022-09-23
Add Proxmox installation guide

cercatrova21:


rootzoll:
fixes #3324 

@cercatrova21 PR ready to merge?

cercatrova21:
@rootzoll yes ready to merge

MEdGalo:
Hi, how would the raspiblitz VM handle a shutdown signal from the proxmox VE (in case of low laptop/UPS battery power for example)? The desired behavior would be that it would trigger blitz.shutdown.sh to run, but I don't know if this is the case currently. 

Maybe a solution could be to run blitz.shutdown.sh with systemd right before shutdown of the raspiblitz VM? [()?](https://www.golinuxcloud.com/run-script-with-systemd-before-shutdown-linux/)

Issue #3329, opened 2022-09-24
[Discussion] Initial Block Download over clearnet, I2P and Tor by default


openoms:
Aware of this is a sensitive topic, but discussing is necessary. 

## The problem

The environment keeps changing with Tor currently practically broken: https://status.torproject.org/

Having report of IBD on RPi4+SSD taking 2 weeks and misidentifying the bottleneck being the RPi, when it is actually Tor.

See for comparison the 3 years old tweet about the full sync over Tor-only in 80 hours: https://twitter.com/openoms/status/1147393450171674624

Also the bitcoin network has some segregation risk if may nodes are only connecting to other Tor-only nodes.

Possible improvements:

## Hybrid mode for Bitcoin Core

 Simply need to delete the line:
```
onlynet=onion
```
from your bitcoin.conf (menu - SYSTEM - bitcoinCONF) to sync connect to peers over clearnet also.
Just be aware that your ISP will see that you are using bitcoin - it is obvious anyway even over from the traffic and download spikes on the new blocks and is ok in most countries.

## I2P as an alternative network
can achieve significant speed improvements, but more peers are needed.
See how to set up easily in the command line:
https://github.com/rootzoll/raspiblitz/issues/2413#issuecomment-1013855483

## what can we do?
- start nodes in hybrid mode (for bitcoin core, not lightning) by default with a warning
- implement I2P to be used by default as well as Tor
- wonder if adding a warning and an option to the web-gui to avoid exposing the clearnet traffic is reasonably easy @cstenglein?
- if there is a synced full node on your LAN or on the VPN the node is connected to can make the download happen over the fast, private network by adding the address of the synced node with:
  ```
  bitcoin-cli addnode “IP_ADDRESS:8333” “onetry”
  ```

Comments and suggestions are most welcome.

ghost:
Instead of removing `onlynet=onion`, we could add `onlynet=i2p` so that Tor and i2p are used for outbound connections.

In case this doesn't help, IPv4 and IPv6 peers are required, a free VPN like [riseup](https://riseup.net/en/vpn/linux) could be used.

Emzy:
As you said, the only downside is that your ISP would see more clearly you run a Bitcoin Core node.
It is very important for your LN node to get all new blocks and don't get stuck.
I think running more than one network (Clearnet + Tor) is the safer option anyway. 
Clearly this should be an option, or even the default option.

openoms:
[...]

Issue #2413, opened 2021-07-07
I2P - [ FEATURE REQUEST]


nyxnor:
https://landchad.net/i2p

https://twitter.com/jonatack/status/1366764964896075776
https://twitter.com/openoms/status/1412879980254711809

https://bitcoin.stackexchange.com/questions/103402/how-can-i-use-bitcoin-core-with-the-anonymous-network-protocol-i2p

https://github.com/bitcoin/bitcoin/blob/master/doc/i2p.md
https://en.wikipedia.org/wiki/I2P

https://geti2p.net/en/
https://geti2p.net/en/comparison/tor

I2P nodes to connect https://twitter.com/emzy/status/1385610565305999362
https://github.com/bitcoin/bitcoin/commit/e01f173fb9b9d35729c700199ba6cf1c028fcaaf
https://github.com/bitcoin/bitcoin/blob/master/contrib/seeds/nodes_main.txt

Implement I2P as an network option.

* I2P: Invisible Internet Project: a project meant to provide an anonymity layer, so user can communicate anonymously using a range of applications.
* Router: The core I2P software, which routes encrypted packets on the I2P network. All routers by default participate in the network, which both helps the network and provides cover traffic for any clients or servers connecting to the I2P network through the router.
* RouterIdentity: A collection of information required to communicate directly with a router, such as its IP address and listening port, public signing and encryption keys etc.
* Tunnel: An anonymous communication pathway between a client or server and the I2P network. Tunnels are unidirectional, so any one client or server must have at least two Tunnels - one for inbound traffic and one for outbound traffic.
* Destination: The cryptographic identity of a tunnel. These are the identities of clients and servers within the I2P network, and are analogous to the IP:port of a computer on the normal internet.
*LeaseSet: A collection of information required to communicate with a client or server at a particular Destination, such as the gateways of the inbound Tunnels for that Destination.

**What is I2P?**

The Invisible Internet Project (I2P) is a fully encrypted private network layer. It protects your activity and location. Every day people use the network to connect with people without worry of being tracked or their data being collected. In some cases people rely on the network when they need to be discrete or are doing sensitive work.

**I2P Cares About Privacy**

[...]

PR #3330, opened 2022-09-24
feat: add I2P

openoms:
Fixing: #2413 

Related discussion: #3329

There could a be a nice menu in a separate PR with:

- i2pd webconsole over LAN
- i2pd.conf file edit
- bitcoin network peers list
- add all i2p seed nodes with onetry


home.admin/config.scripts/bonus.i2pd.sh
```diff
@@ -0,0 +1,103 @@
+#!/bin/bash
+
+if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
+  echo "# bonus.i2pd.sh on       -> install the i2pd"
+  echo "# bonus.i2pd.sh off      -> uninstall the i2pd"
+  echo "# bonus.i2pd.sh addseednodes -> Add all I2P seed nodes from: https://github.com/bitcoin/bitcoin/blob/master/contrib/seeds/nodes_main.txt"
```
kristapsk: Why? To update node list from newer version without upgrading bitcoind? Because from `contrib/seeds/nodes_main.txt` `src/chainparamseeds.h` is generated, Bitcoin Core already knows about them. https://github.com/bitcoin/bitcoin/tree/master/contrib/seeds

rootzoll:
> TODO: edit blitz.conf.sh to be able to add multiple lines with the same key

Do you mean the value of a key could be a multiline string? Or do you need a key that can have multiple values (list)?

kristapsk:
Sume users with limited Internet speed / bandwidth might also want to either limit I2P bandwidth or disable relaying public I2P traffic. https://github.com/raspibolt/raspibolt/issues/1086

home.admin/config.scripts/bonus.i2pd.sh
openoms: how do I connect to the seed nodes without adding the peers manually with the bitcoin-cli or listi.g them in the bitcoin.conf?
Or would they just connect automatically if there are no 10 peers already?

openoms:
> > TODO: edit blitz.conf.sh to be able to add multiple lines with the same key
> 
> Do you mean the value of a key could be a multiline string? Or do you need a key that can have multiple values (list)?

needs keys which are repeated with different values like:
```
debug=tor
debug=i2p
```
and
```
onlynet=tor
onlynet=i2p
```
this will make bitcoind connect to both tor and i2p peers, but not clearnet. For clearnet (hybrid mode) can add 
```
onlynet=ipv4
```
or remove all the `onlynet` entries

home.admin/config.scripts/bonus.i2pd.sh
kristapsk: Should connect automatically, they are in https://github.com/bitcoin/bitcoin/blob/master/src/chainparamsseeds.h, compiled into bitcoind. Besides, these are seed nodes, bitcoind will connect to them, get list of peers and disconnect. Not use them as peers. Also, sooner or later you will get information about I2P peers from connected onion peers too.

openoms: in case you add them manually they stay as peers. 
I think it can be a good option to force start bootstrapping, but won't make it run by default. Might just leave it as a CLI option only.

openoms:
> Sume users with limited Internet speed / bandwidth might also want to either limit I2P bandwidth or disable relaying public I2P traffic. [raspibolt/raspibolt#1086](https://github.com/raspibolt/raspibolt/issues/1086)

It seems that if you don't specify and forward the port there is no bandwith used for relaying: https://github.com/seth586/guides/blob/master/FreeNAS/bitcoin/freenas_3_tor.md#optional-relay

openoms:
The script feels ready, but more testing is needed.
On an RPi my Bitcoin Core doesn't pick up i2p peers and on x86 I get a segmentation fault.

Also thinking of what menu options would be needed if any.

openoms:
moved to the description

rootzoll:
> needs keys which are repeated with different values like:
> ```
> onlynet=tor
> onlynet=i2p
> ```

@openoms so basically you need a list, correct? I think instead of "multiple same keys" I would recommend "comma seperated values" on one key - like:

```
onlynet=tor,i2p
```

So in bash scripts (if they simply import the value) you could still check for the value with:
```
if [[ $onlynet =~ "i2p" ]]; then
  echo "I2P is active"
fi
```

I could add some features to the `blitz.conf.sh` script to make it easiert to add & remove list values. Would that make sense to you?

openoms:
> > needs keys which are repeated with different values like:
> > ```
> > onlynet=tor
> > onlynet=i2p
> > ```
> 
> @openoms so basically you need a list, correct? I think instead of "multiple same keys" I would recommend "comma seperated values" on one key - like:
> 
> ```
> onlynet=tor,i2p
> ```
> 
> So in bash scripts (if they simply import the value) you could still check for the value with:
> 
> ```
> if [[ $onlynet =~ "i2p" ]]; then
>   echo "I2P is active"
> fi
> ```
> 
> I could add some features to the `blitz.conf.sh` script to make it easiert to add & remove list values. Would that make sense to you?

it is already covered in:
https://github.com/rootzoll/raspiblitz/pull/3330/commits/04bbc16b5c52afabe95847124cb9739acc2adfcc

`blitz.conf.sh add`
adds a line with `tee -a` rather than change the existing value with `sed` when using `blitz.conf.sh set`

but if you would prefer doing it differently happy to change it.
It gets a little complicated when we want to remove one entry.
Currently the script removes all entries with the same key with `delete` then can use  `set` to make one entry and `add` to have more.

rootzoll:
> but if you would prefer doing it differently happy to change it.

I merged #3334 into dev which introduced `list-add` & "list-remove" actions to `blitz.conf.sh` .. I think this way we can handle list values much cleaner. List items are basically words seperated by spaces now. It takes care that list values are uniquie - so even if you add an item twice it will only be in the list once.

So now you can still source the config file in a bash script and check if list item is active with:
```
if [[ $onlynet =~ "i2p" ]]; then
  echo "I2P is active"
fi
```

@openoms if this works for you, please use this new functions in this PR (remove your changes to ´blitz.conf.sh´)

home.admin/config.scripts/bonus.i2pd.sh
```diff
@@ -211,12 +210,10 @@ if [ "$1" = "0" ] || [ "$1" = "off" ]; then
...
 
   echo "# Remove settings from bitcoind"
-  /home/admin/config.scripts/blitz.conf.sh delete debug /mnt/hdd/bitcoin/bitcoin.conf
-  /home/admin/config.scripts/blitz.conf.sh set debug tor /mnt/hdd/bitcoin/bitcoin.conf
+  /home/admin/config.scripts/blitz.conf.sh list-delete debug i2p /mnt/hdd/bitcoin/bitcoin.conf
   /home/admin/config.scripts/blitz.conf.sh delete i2psam /mnt/hdd/bitcoin/bitcoin.conf
   /home/admin/config.scripts/blitz.conf.sh delete i2pacceptincoming /mnt/hdd/bitcoin/bitcoin.conf
-  /home/admin/config.scripts/blitz.conf.sh delete onlynet /mnt/hdd/bitcoin/bitcoin.conf
-  /home/admin/config.scripts/blitz.conf.sh set onlynet tor /mnt/hdd/bitcoin/bitcoin.conf
+  /home/admin/config.scripts/blitz.conf.sh list-delete onlynet i2p /mnt/hdd/bitcoin/bitcoin.conf
```
rootzoll: its `list-remove`

home.admin/config.scripts/bonus.i2pd.sh
```diff
@@ -112,12 +112,11 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
 
   sudo systemctl enable i2pd
 
-  /home/admin/config.scripts/blitz.conf.sh set debug tor /mnt/hdd/bitcoin/bitcoin.conf noquotes
-  /home/admin/config.scripts/blitz.conf.sh add debug i2p /mnt/hdd/bitcoin/bitcoin.conf noquotes
+  /home/admin/config.scripts/blitz.conf.sh list-add debug tor /mnt/hdd/bitcoin/bitcoin.conf noquotes
```
rootzoll: noquotes has just an effect on `set` - not on `list-add`

openoms:
This PR is functional with I2P running in the background from the start, but no menu entries yet. Those can be added separately.

MEdGalo:
HI, I tried installing I2p with bonus.i2pd.sh on, but this resulted in:
'#' Starting i2pd service ...
Job for i2pd.service failed because the control process exited with error code.
See "systemctl status i2pd.service" and "journalctl -xe" for details.
'#' Restart bitcoind ...
i2pd: error while loading shared libraries: libboost_filesystem.so.1.74.0: cannot open shared object file: No such file or directory
'#' i2pd is not installed

Since then my bitcoind is also not able to start anymore. I tried bonus.i2pd.sh off, but that didn't help. 

Any idea why the I2p install script is not working for me and how I can get my bitcoind to work again?

openoms:
the `blitz.conf.sh list-add` option will need to be tested again. Changing to draft until done.

openoms:
@rootzoll we cannot use a list in the bitcoin.conf

This is not doing what is needed
```
/home/admin/config.scripts/blitz.conf.sh set debug tor /mnt/hdd/bitcoin/bitcoin.conf noquotes
/home/admin/config.scripts/blitz.conf.sh list-add debug i2p /mnt/hdd/bitcoin/bitcoin.conf noquotes
```
results in:
```  
debug='tor i2p'
```
instead of:
```
debug=tor 
debug=i2p
```
we need to have multiple entries of the same key, that's how bitcoin.conf is interpreted correctly

openoms:
Added what is needed in a simple function in [cf84042](https://github.com/rootzoll/raspiblitz/pull/3330/commits/cf84042709c03dfd3830cc155157a285c435cd49). The script is working again.

rootzoll:
@openoms when the I2P service is installed during build sd card - there is no passwordB set. So in `/etc/i2pd/i2pd.conf` there is an empty password and the "i2pd web console" cannot be loged in. I think something like a "prestart script" could make sense here - then also in case of PasswordB change, the service just need to restart.

rootzoll:
@openoms also the changes of `bitcoin.conf` dont work during sd card install - which is a bit harder to solve with a `prestart script`. Maybe like with lnd/cln we need a seperate `install` (during sd card) & `on` (during provision) phase.

home.admin/config.scripts/bonus.i2pd.sh
```diff
@@ -0,0 +1,300 @@
...
+  echo
+}
+
+
+echo "# Running: 'bonus.i2pd.sh $*'"
+source /mnt/hdd/raspiblitz.conf
+
+if [ "$1" = "install" ]; then
+
+  if apt list i2pd; then
```
rootzoll: this install check with "apt list" seems not to work - best to remove it and just run install, otherwise install mit not work at all.

home.admin/_provision.setup.sh
```diff
@@ -134,6 +134,10 @@ do
 done
 echo "OK ${network} startup successful " >> ${logFile}
 
+# I2P
```
rootzoll: does not need to be called in provision.setup.sh script .. because its already called in _provision_.sh which is always executed.

rootzoll:
OK I gave it another test build (sd card without fatbuild) and setup a node with just bitcoin. 

Checking `bonus.i2pd.sh status` gave me:

```
2022-12-07T11:53:16Z Config file arg: debug="i2p"
2022-12-07T11:53:16Z Config file arg: i2pacceptincoming="1"
2022-12-07T11:53:16Z Config file arg: i2psam="127.0.0.1:7656"
2022-12-07T11:54:07Z i2paccept thread start
2022-12-07T11:54:07Z [i2p] Creating persistent SAM session 25230017af with 127.0.0.1:7656
2022-12-07T11:54:07Z [i2p] Error listening: Cannot connect to 127.0.0.1:7656
2022-12-07T11:54:07Z [i2p] Control socket error: not connected
```

It seems the install of i2pd went wrong:
```
admin@192.168.178.147:~ ₿ sudo cat /etc/i2pd/i2pd.conf
cat: /etc/i2pd/i2pd.conf: No such file or directory
admin@192.168.178.147:~ ₿ sudo systemctl status i2pd
Unit i2pd.service could not be found.
```

rootzoll:
OK I think it was still because of the install detection - I changed and will give it another try. Also renamend script to `blitz.i2pd.sh` - because its more an feature of the blitz core instead of a bonus app.

rootzoll:
Ok now sd card build looks better. The i2pd get is istalled and is running with correct conf (passwordB is set). I can also call the web console now:
Bildschirmfoto 2022-12-07 um 14 11 10

rootzoll:
OK added i2p also to the fallbacklist & kickstart mechanism - and background process is now also triggers kickstart of i2p if i2p is turned on and i2p-peers is zero

rootzoll (approved):

PR #3336, opened 2022-09-30
Calculate LN amounts in lnd.balance.sh using msats

b-ryan:
This changes the `lnd.balance.sh` script so that the ln invoices, payments, and fees totals are calculated using msats and then rounded. Before, with the rounding happening before the aggregation, there is a loss of precision.

I discovered this as I was auditing my balances and saw that the fees I was manually calculating differed from what raspiblitz said by a few sats. I saw that there is a warning about this in the output, but I wanted to know more and eventually discovered this code. It looked like a straightforward enough thing to change, so I thought I would throw this PR out there.


rootzoll:
Thanks for the accurate improvement :)

rootzoll:
Will be part of v1.8.1 release.

PR #3347, opened 2022-10-07
chantools update to v0.10.5

guggero:
Updates `chantools` to `v0.10.5` which is compatible with `lnd v0.15.1-beta`.
See https://github.com/guggero/chantools/releases/tag/v0.10.5.


rootzoll:
test worked. merging to be part of next release.

PR #3353, opened 2022-10-10
lnd update to v0.15.2

openoms:
urgent bugfix update to be used by all new builds


AdamISZ:
Not familiar with the setup here, but doesn't the comment on line 6 say you should also update `lnd.update.sh`  at the same time?

openoms:
> Not familiar with the setup here, but doesn't the comment on line 6 say you should also update `lnd.update.sh` at the same time?

thanks for checking. That `lndUpdateVersion` is basically does add the option to apply the fix with PGP verification when someone patched scripts, but not the whole SDcard image yet (aka dirty update - not recommended)
When a new SDcard image will be released (which is the plan here) the setting is irrelevelant again as the image will contain the latest LND version.

PR #3365, opened 2022-10-12
fix(bitcon.update): unary operator expected

OliverOffing:
Fixes syntax error

### DIFF
--- a/home.admin/config.scripts/bitcoin.update.sh
+++ b/home.admin/config.scripts/bitcoin.update.sh
@@ -137,7 +137,7 @@ if [ "${mode}" = "tested" ]||[ "${mode}" = "reckless" ]||[ "${mode}" = "custom" 
   
   displayInfo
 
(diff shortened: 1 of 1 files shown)


home.admin/config.scripts/bitcoin.update.sh
```diff
@@ -137,7 +137,7 @@ if [ "${mode}" = "tested" ]||[ "${mode}" = "reckless" ]||[ "${mode}" = "custom"
   
   displayInfo
 
-  if [ $installedVersion = $bitcoinVersion ];then
+  if [ $installedVersion = $bitcoinVersion ]; then
```
openoms: adding a space won't make any difference there.
Can add quotes to avoid the error when the variable is not defined:
```suggestion
  if [ "$installedVersion" = "$bitcoinVersion" ]; then
```

rootzoll:
also cherry-picked over intoi dev branch

PR #3384, opened 2022-10-24
Redirect nginx HTTP to HTTPS and change Port

ChuckNorrison:
Cause of in example LNBits service which is already set to 443, set WebUI on 4443

Never allow port 80 cause of missing encryption. Redirect it to HTTPS.


ChuckNorrison:
Does not work on TOR Nodes.

Issue #3294, opened 2022-08-17
Electrs menu doesn't show the Tor address


openoms:
Reported on TG.

Workaround for  now:

Run:
https://github.com/rootzoll/raspiblitz/blob/7074455380eec200bfcaa35cd6cca066fcf9bef3/home.admin/config.scripts/bonus.electrs.sh#L441

or 
```
sudo cat /mnt/hdd/tor/electrs/hostname
```

TCP port 50001
SSL 50002

PR #3388, opened 2022-10-26
Fix localIP and torAddress gathering

LeoSpyke:
I fixed the commands used to gather the local ip addresses in `bonus.btc-rpc-explorer.sh` and `bonus.homer.sh` because previously, they were not able to retrieve the network cards with a prefix different from `eth0\|wlan0\|enp0` (in my case it was indeed `enp7s0` on a Debian amd64 machine).
Actually, many other scripts are already using this "new" logic.

Finally, in the `bonus.electrs.sh` I just added the second flag `showAddress` in order to show the tor address in the "status" menu.

### DIFF
--- a/home.admin/config.scripts/bonus.btc-rpc-explorer.sh
+++ b/home.admin/config.scripts/bonus.btc-rpc-explorer.sh
@@ -85,7 +85,7 @@ if [ "$1" = "status" ]; then
     echo "installed=${installed}"
 
     # get network info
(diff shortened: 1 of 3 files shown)


openoms:
Thanks, I also found that this method of gathering the IP address is much more reliable. Merging.
Fixes #3294

openoms:
Cherry picking to v1.8 as it is such a common question re the Electrs Tor address.

PR #3389, opened 2022-10-26
LNBits branch master does not exist

ChuckNorrison:
githubRepo lnbits-legends gets redirected to lnbits

the default branch of lnbits is main


openoms:
better defaults, thanks

PR #3398, opened 2022-11-01
Electrs update and menu

ChuckNorrison:
This should allow to update electrs


ChuckNorrison:
mixed with some old pull requests... :(

PR #3399, opened 2022-11-01
Add update for electrs v0.9.9

ChuckNorrison:
update method and menu entry added

lnd hotfix for electrs


home.admin/config.scripts/bonus.electrs.sh
```diff
@@ -510,5 +510,18 @@ if [ "$1" = "0" ] || [ "$1" = "off" ]; then
   exit 0
 fi
 
+if [ "$1" = "update" ]; then
+  cd /home/electrs/electrs || exit 1
+  sudo -u electrs git fetch
+  sudo -u electrs git reset --hard $ELECTRSVERSION
```
openoms: this only works if we manually update the version or run a patch. See how the github tags are parsed in other scripts like in: https://github.com/rootzoll/raspiblitz/blob/dev/home.admin/config.scripts/bonus.btcpayserver.sh#L747-L785

ChuckNorrison: For electrs this method does not work. So i changed it to use the tags instead. Please test :)

openoms:
tACK [460d659](https://github.com/rootzoll/raspiblitz/pull/3399/commits/460d659562036e818242b980b398964b1ba4b90c) by updating v0.9.7 to v0.9.9 on amd64.

We are aiming for v1.8.1 so should be rebased on the `dev` branch and ready to merge there.

ChuckNorrison:
i will make a new PR in dev

home.admin/config.scripts/lnd.install.sh
```diff
@@ -4,6 +4,7 @@
 ## based on https://raspibolt.github.io/raspibolt/raspibolt_40_lnd.html#lightning-lnd
 ## see LND releases: https://github.com/lightningnetwork/lnd/releases
 ### If you change here - make sure to also change interims version in lnd.update.sh #!
+
```
openoms, line 7: no change here

PR #3407, opened 2022-11-02
Fix: Disabling fulcrum accidentally removes Electrs Tor hidden service

steepdawn974:
This fixed the bug that `bonus.fulcrum.sh off` removes accidentally the hidden service for electrs, instead of fulcrum


openoms:
Thanks!

PR #3411, opened 2022-11-03
LNBits SQLite to PostgreSQL migration

ChuckNorrison:
New installations with PostgreSQL configuration, existent data with SQLite.

convert script needs a full start of LNBits prior migrating the old sqlite db's.

User can migrate from raspiblitz menu.


ChuckNorrison:
Was able to migrate on one machine but not on another with a very old database. 

The conversion script stucks with 

```
admin@192.168.8.160:/home/lnbits/lnbits((0.9.4)) ₿ sudo -u lnbits ./venv/bin/python tools/conv.py
Selected path:  /mnt/hdd/app-data/LNBits
Traceback (most recent call last):
  File "/home/lnbits/lnbits/tools/conv.py", line 208, in <module>
    check_db_versions(file)
  File "/home/lnbits/lnbits/tools/conv.py", line 61, in check_db_versions
    raise Exception(
Exception: sqlite database version (1) of tpos doesn't match postgres database version 3
```
Maybe not all databases are easy to migrate or outdated in some way that lnbits cant fix it yet.

A manual solution to migrate documented here:
https://github.com/lnbits/lnbits/issues/1092#issuecomment-1307119310

rootzoll:
Have to check if postgresql is just an option or somewhere mandatory. Because once its podtsgresql we might cannot migrate back, right?

rootzoll:
TODO: a quick disclaimer in the cli hlep there should be a disclaimer: "migrate back to sqllite might not be possible"

ChuckNorrison:
It is possible to migrate back. In the last commit i added a `migrateMsg` function with commands to execute to revert the migration back to the old backuped sqlite database. This message gets printed at finish.

```
sudo systemctl stop lnbits
sudo mv /mnt/hdd/app-data/LNBits /mnt/hdd/app-data/LNBits_backup
sudo tar -xf /mnt/hdd/app-data/LNBits_sqlitedb_backup.tar
sudo sed -i "/^LNBITS_DATABASE_URL=/d" /home/lnbits/lnbits/.env
sudo sed -i "/^LNBITS_DATA_FOLDER=/d" /home/lnbits/lnbits/.env
sudo bash -c "echo 'LNBITS_DATA_FOLDER=/mnt/hdd/app-data/LNBits' >> /home/lnbits/lnbits/.env"
sudo chown lnbits:lnbits -R /mnt/hdd/app-data/LNBits
sudo systemctl start lnbits
```

This could be further improved... maybe a dialogue can be added to accept the migration process with the hints on how to revert if something goes wrong

The postgresql migration is not mandatory but highly recommended by lnbits, see [here](https://github.com/lnbits/lnbits/blob/main/docs/guide/installation.md#basic-installation). 

I did a lot of research on how to migrate my two year old database, cause it has some database corruptions shown with use of extensions. 
The migrate progress is a good opportunity to fix any existent database corruptions and maybe is better supported for future use than the sqlite solution.

Not tested yet:
- reflash sd card with existent LNBits on SQLite
- reflash sd card with existent LNBits on PostgreSQL
- install and remove BTCPayServer (it runs on PostgreSQL)

Successful tested:
- Migrate existent LNBits from SQLite to PostgreSQL
- Revert migration with manual steps

ChuckNorrison:
Did a lot more improvements for several circumstances i tested. Should be kind of bullet proof at this point. No risk at all and a beautiful dialog behaviour :)

Test Error behaviour:
- Set your usr string as LNBITS_ADMIN_USERS in .env
- Increase your balance amount to 21000000
- Start migrate with Tag 0.9.4

This got fixed [here](https://github.com/lnbits/lnbits/pull/1030). Hopefully we can see a new LNBits Tag 0.9.5. soon. We should not migrate prior v0.9.5.

rootzoll:
@arcbtc or @callebtc ... an lnbits v0.9.5 is planned within the next weeks, correct?

Also while we talk about versioning of lnbits for raspiblitz - do we want to use raspiblitz specific tags - that you can move in case of an important bug fix?

I would propose tags like `raspiblitz_v1.8`, `raspiblitz_v1.9` ...   the raspiblitz lnbits install script will check the github API if a tag for the latest release branch is available thru: https://api.github.com/repos/lnbits/lnbits/tags .. if its not yet set it will fallback to a fixed version release. So if you have an emergency update in the future - you can simply add the tag and ask users to uninstall & reinstall lnbits to update. Would that make sense?

rootzoll (approved):

rootzoll:
Merged into `3411-lnbits` feature branch further testing on issue #3505

PR #3419, opened 2022-11-06
Add staticIP config and fix 127.0.0.1

ChuckNorrison:
Add possibility of a staticIP config in Raspiblitz.conf to override globalIP and publicIP detection.

Skip dyndns in case of static ip usage

if curl fails accidentaly it could result in the wrong IP. Only prefer if publicIP is empty.


rootzoll (approved):

rootzoll:
OK test worked. Merging.

Issue #3423, opened 2022-11-07
Issue detecting SSD versus Mechanical Drive


rek79:
https://github.com/rootzoll/raspiblitz/blob/b20293a0c90f531477774c9a810b5491c81d0166/home.admin/config.scripts/blitz.datadrive.sh#L166

On Odroid HC2 the code referenced above incorrectly returns "1" (or true for rotational) even though a solid state drive is attached.  I've found an alternate method (see https://ostechnix.com/how-to-find-if-the-disk-is-ssd-or-hdd-in-linux/) that accurately returns "Solid State Device".  It does require that "smartctl is installed on the system:

Command:
`sudo smartctl -a /dev/sda | grep 'Rotation Rate'`
if SSD, returns:
`Rotation Rate:    Solid State Device`
if mechanical HD, returns:
`Rotation Rate:    5400 rpm`

We could modify this to check for the word "Solid State", for example:
`isSSD=$(sudo smartctl -a /dev/${hdd} | grep 'Rotation Rate' | grep -c "Solid State")`

PR #3424, opened 2022-11-08
fix localIP assignment and improve SSD detection

rek79:
**Fix localip assignment**:
use `localip=$(hostname -I | awk '{print $1}')` instead of `localip=$(ip addr | grep 'state UP' -A2 | grep -E -v 'docker0|veth' | grep 'eth0\|wlan0\|enp0' | tail -n1 | awk '{print $2}' | cut -f1 -d'/')`.  The former assignment seems to work on a wider range of hardware.
related: https://github.com/rootzoll/raspiblitz/commit/5d61d0b216b281ff7b99e0488c39390d1d5e038c#commitcomment-89042523

**Improve reliability of solid state drive detection**:
smartctl seems to be more compatible with a wider array of hardware.
`isSSD=$(cat /sys/block/${hdd}/queue/rotational 2>/dev/null | grep -c 0)` incorrectly sets SSD to false for Odroid HC2.  `isSSD=$(sudo smartctl -a /dev/sda | grep 'Rotation Rate' | grep -c "Solid State")` correctly identifies the SSD.
Related issue: https://github.com/rootzoll/raspiblitz/issues/3423#issue-1439092806

### DIFF
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -163,7 +163,7 @@ if [ "$1" = "status" ]; then
     fi
 
     # try to detect if its an SSD 
-    isSSD=$(sudo smartctl -a /dev/sda | grep 'Rotation Rate' | grep -c "Solid State")
+    isSSD=$(sudo smartctl -a /dev/${hdd} | grep 'Rotation Rate' | grep -c "Solid State")
     echo "isSSD=${isSSD}"
 
     # display results from hdd & partition detection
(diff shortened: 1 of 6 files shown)


rek79:
@openoms @rootzoll this is follow on work to my related merged PR #2432 under [v1.7.1](https://github.com/rootzoll/raspiblitz/releases/tag/v1.7.1).  The solid state drive fix is not related but included here since it seems to be an improvement for hardware compatibility.  I can separate that out if we think that will take longer to test.  Please let me know.

ChuckNorrison (commented):
smartctl: command not found

This needs to be installed

rek79:
@ChuckNorrison check for package "smartmontools" added.  It should work now if not already installed.

ChuckNorrison:
Looks good so far, but what if a device is unsupported by smartmontools like described [here](https://www.smartmontools.org/wiki/Unsupported_USB-Devices)?

Should we preserve the command `cat /sys/block/${hdd}/queue/rotational` as a fallback?

rek79:
Fallback added.  If "Rotation Rate:" is found then` smartctl` method is used to detect the SSD.  Otherwise `cat /sys/block/${hdd}/queue/rotational` is used.
@ChuckNorrison in the link of unsupported devices, it seems that some of those devices still report rotation so this seems to be a robust way forward.  Thanks for the input!

rootzoll:
PR basically looks good for final testing but is based on v1.8 release ... @rek79 could you re-base it on `dev` branch?

rek79:
Opened a new pull request under #3439 which is re-based to `dev` branch.  I'll close this out.

PR #3430, opened 2022-11-09
lit update to v0.8.3-alpha

openoms:
The qustion here if really should stop running Lit over Tor?

This was to torify the Lightning Connect tunnel, but given the current state of Tor this might not make sense any more. 

Remember discussing with @kaloudis that the LN node ID is not exposed to the ligthning connect proxy server so it would not be assosiated with the IP address, right?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,8 @@
 ## What's new in Version 1.8.1 of RaspiBlitz?
 
 - New: Support of X708 UPS HAT [details](https://github.com/rootzoll/raspiblitz/pull/3087)
-- Update: LND v0.15.3 [details](https://github.com/lightningnetwork/lnd/releases/tag/v0.15.3-beta
+- Update: LND v0.15.4 [details](https://github.com/lightningnetwork/lnd/releases/tag/v0.15.4-beta
+- Update: Lightning Terminal v0.8.3-alpha [details](https://github.com/lightninglabs/lightning-terminal/releases/tag/v0.8.3-alpha)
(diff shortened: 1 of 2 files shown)


rootzoll:
First Testrun failed with:

```
manifest-v0.8.3-alpha.sig   100%[========================================>]     566  --.-KB/s    in 0s

2022-11-15 16:16:47 (6.89 MB/s) - ‘manifest-v0.8.3-alpha.sig’ saved [566/566]

--2022-11-15 16:16:47--  https://github.com/ellemouton.gpg
Resolving github.com (github.com)... 140.82.121.4
Connecting to github.com (github.com)|140.82.121.4|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 3150 (3.1K) [text/plain]
Saving to: ‘ellemouton.gpg’

ellemouton.gpg              100%[========================================>]   3.08K  --.-KB/s    in 0s

2022-11-15 16:16:47 (13.4 MB/s) - ‘ellemouton.gpg’ saved [3150/3150]

# check gpg finger print
gpg: can't open './pgp_keys.asc': No such file or directory

# BUILD WARNING --> LiT PGP author not as expected
Should contain PGP: D7D916376026F177
PRESS ENTER to TAKE THE RISK if you think all is OK
```

rootzoll:
OK test worked now. Merging for next release.

kaloudis:
@openoms the LN public key isn't used as an identifier. A 64-byte stream ID is created on initial handshake.

PR #3439, opened 2022-11-14
Improve localIP assignment and SSD detection

rek79:
**Rebased PR #3424 from [rootzoll/raspiblitz](https://github.com/rootzoll/raspiblitz) to [dev ](https://github.com/rootzoll/raspiblitz/tree/dev) branch**

**1. Fix localip assignment:**
use `localip=$(hostname -I | awk '{print $1}')` in place of `localip=$(ip addr | grep 'state UP' -A2 | grep -E -v 'docker0|veth' | grep 'eth0\|wlan0\|enp0' | tail -n1 | awk '{print $2}' | cut -f1 -d'/')`.
The former assignment seems to work on a wider range of hardware.
related: https://github.com/rootzoll/raspiblitz/commit/5d61d0b216b281ff7b99e0488c39390d1d5e038c#commitcomment-89042523

**2. Improve reliability of solid state drive detection:**
smartctl seems to be more compatible with a wider array of hardware.
`isSSD=$(cat /sys/block/${hdd}/queue/rotational 2>/dev/null | grep -c 0)` incorrectly sets SSD to false for Odroid HC2. `isSSD=$(sudo smartctl -a /dev/${hdd} | grep 'Rotation Rate:' | grep -c "Solid State")` correctly identifies the SSD.
Related issue: https://github.com/rootzoll/raspiblitz/issues/3423#issue-1439092806

### DIFF
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -39,12 +39,12 @@ if [ ${btrfsInstalled} -eq 0 ]; then
 fi
 
 # install smartmontools if needed
-smartmontoolsInstalled=$(dpkg -s smartmontools 2>/dev/null | grep -c "install ok installed")
+smartmontoolsInstalled=$(apt-cache policy smartmontools | grep -c 'Installed: (none)' | grep -c "0")
 if [ ${smartmontoolsInstalled} -eq 0 ]; then
   >&2 echo "# Installing smartmontools ..."
   apt-get install -y smartmontools 1>/dev/null
 fi
-smartmontoolsInstalled=$(dpkg -s smartmontools 2>/dev/null | grep -c "install ok installed")
+smartmontoolsInstalled=$(apt-cache policy smartmontools | grep -c 'Installed: (none)' | grep -c "0")
 if [ ${smartmontoolsInstalled} -eq 0 ]; then
   echo "error='missing smartmontools package'"
   exit 1
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -177,6 +177,7 @@ if [ "$1" = "status" ]; then
 
     # try to detect if its an SSD
     isSMART=$(sudo smartctl -a /dev/${hdd} | grep -c "Rotation Rate:")
+    echo "isSMART=$(isSMART)"
     if [ ${isSMART} -gt 0 ]; then
     	#detect using smartmontools (preferred)
         isSSD=$(sudo smartctl -a /dev/${hdd} | grep 'Rotation Rate:' | grep -c "Solid State")
@@ -406,6 +407,7 @@ if [ "$1" = "status" ]; then
 
     # try to detect if its an SSD
     isSMART=$(sudo smartctl -a /dev/${hdd} | grep -c "Rotation Rate:")
+    echo "isSMART=$(isSMART)"
     if [ ${isSMART} -gt 0 ]; then
     	#detect using smartmontools (preferred)
         isSSD=$(sudo smartctl -a /dev/${hdd} | grep 'Rotation Rate:' | grep -c "Solid State")
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -38,6 +38,18 @@ if [ ${btrfsInstalled} -eq 0 ]; then
   exit 1
 fi
 
+# install smartmontools if needed
+smartmontoolsInstalled=$(dpkg -s smartmontools 2>/dev/null | grep -c "install ok installed")
+if [ ${smartmontoolsInstalled} -eq 0 ]; then
+  >&2 echo "# Installing smartmontools ..."
+  apt-get install -y smartmontools 1>/dev/null
+fi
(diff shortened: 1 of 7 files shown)


rek79:
@rootzoll should be all set to test now that the PR is based on `dev` .

ChuckNorrison:
[wifilocalIP](https://github.com/rootzoll/raspiblitz/pull/3439/commits/3e48433322bdecbba2353fe509e275ea7851b80c) is still wrong and with the wildcard it will print the LAN IP instead nothing, if wifi is not used

rek79:
> [wifilocalIP](https://github.com/rootzoll/raspiblitz/pull/3439/commits/3e48433322bdecbba2353fe509e275ea7851b80c) is still wrong and with the wildcard it will print the LAN IP instead nothing, if wifi is not used
@ChuckNorrison 
Is it as simple as changing it to `hostname -I | awk '{print $1}'` ?

I have my odroid wired with no wireless adapter so I can't test the wifi connection.

ChuckNorrison:
Just remove the wildcard (*) from the IP addr command to get the wifi IP. You can check if it prints your lan ip, it should not print your lan ip. Should be empty without wifi used

rek79:
@ChuckNorrison Okay, reverted back to original.  Should be all set now.

home.admin/config.scripts/blitz.datadrive.sh
```diff
@@ -38,6 +38,18 @@ if [ ${btrfsInstalled} -eq 0 ]; then
   exit 1
 fi
 
+# install smartmontools if needed
+smartmontoolsInstalled=$(dpkg -s smartmontools 2>/dev/null | grep -c "install ok installed")
```
rootzoll: Is there a special reason ti use `dpkg` instead of `apt-get`? `apt-get` would also check for missing dependencies (may be needed on other base OS) and so be more error blocking prone.

home.admin/config.scripts/blitz.datadrive.sh
```diff
@@ -162,8 +175,15 @@ if [ "$1" = "status" ]; then
...
+    # try to detect if its an SSD
+    isSMART=$(sudo smartctl -a /dev/${hdd} | grep -c "Rotation Rate:")
+    if [ ${isSMART} -gt 0 ]; then
+    	#detect using smartmontools (preferred)
+        isSSD=$(sudo smartctl -a /dev/${hdd} | grep 'Rotation Rate:' | grep -c "Solid State")
+    else
+    	#detect using using fall back method
+	isSSD=$(cat /sys/block/${hdd}/queue/rotational 2>/dev/null | grep -c 0)
+    fi 
     echo "isSSD=${isSSD}"
```
rootzoll, line 188: Would recommend to also echo "isSMART" here for better debug.

home.admin/config.scripts/blitz.datadrive.sh
```diff
@@ -384,7 +404,15 @@ if [ "$1" = "status" ]; then
...
+    # try to detect if its an SSD
+    isSMART=$(sudo smartctl -a /dev/${hdd} | grep -c "Rotation Rate:")
+    if [ ${isSMART} -gt 0 ]; then
+    	#detect using smartmontools (preferred)
+        isSSD=$(sudo smartctl -a /dev/${hdd} | grep 'Rotation Rate:' | grep -c "Solid State")
+    else
+    	#detect using using fall back method
+	isSSD=$(cat /sys/block/${hdd}/queue/rotational 2>/dev/null | grep -c 0)
+    fi
     echo "isSSD=${isSSD}"
```
rootzoll, line 418: Would recommend to also echo "isSMART" here for better debug.

rootzoll (changes requested):

home.admin/config.scripts/blitz.datadrive.sh
rek79: > 

I was attempting to follow the same coding pattern as:
https://github.com/rootzoll/raspiblitz/blob/b8a9c84845d40ea786e48f94cac9e0d927c4b026/home.admin/config.scripts/blitz.datadrive.sh#L29-L39

I could just call `apt-get install` and do away with pre-checking if the package is installed.  Are you suggesting we go that route? Or, is there a way to pre-check if the package is installed using apt-get?

home.admin/config.scripts/blitz.datadrive.sh
rek79: ACK, done.

home.admin/config.scripts/blitz.datadrive.sh
rootzoll: basically - because blitz.datadrive.sh is a very basic always needed script - the dependency can also move into the sd card build script. https://github.com/rootzoll/raspiblitz/blob/8e3fef2144f0cc2810d0278b62fb08eb36d88ab2/build_sdcard.sh#L305

rek79: Replaced `dpkg -s smartmontools 2>/dev/null | grep -c "install ok installed"` with `apt-cache policy smartmontools | grep -c 'Installed: (none)' | grep -c "0"`
Added smartmontools to general_utils.

rootzoll (approved):

rootzoll (approved):

PR #3444, opened 2022-11-15
Add commands confirmation message

ChuckNorrison:
for interrupting commands, ask first

### DIFF
--- a/home.admin/99updateMenu.sh
+++ b/home.admin/99updateMenu.sh
@@ -182,7 +182,7 @@ patch()
           whiptail --title "ERROR" --msgbox "${error}" 8 30
         fi
       fi
-      patch
+      patch all
       exit 0
       ;;
     BRANCH)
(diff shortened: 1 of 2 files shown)


rootzoll:
Please not the `patch` command ... its used in some development flows and always a yes/no confirmation would slow that down.

ChuckNorrison:
> Please not the `patch` command ... its used in some development flows and always a yes/no confirmation would slow that down.

found only the 99updateMenu.sh file who would use the command. Should we change the first parameter from "" to "all" if called from there? Than in _commands.sh we could just check for "" and then print the confirmation message.

// confirmation message for `patch` command removed

rootzoll:
> Than in _commands.sh we could just check for "" and then print the confirmation message.

Yeah that can be a good compromise. If there this no further parameter it wants a confirmation. If there is a second paramter (all, code, ...) it will assume the user knows already and there will be no Y/N to not further interrupt the workflow.

rootzoll (approved):
tests look good

PR #3459, opened 2022-11-20
Update cl.update.sh menu entries

steepdawn974:
The CLVERSION in `cl.install.sh` is already at v0.12.1, so no need to pin 0.10.2 in the patch script


rootzoll:
thx good find

rootzoll (approved):

Issue #3537, opened 2022-12-17
LightningTipBot and LNDg SETTINGS conflict


allyourbankarebelongtous:
In the dev branch, both LightningTipBot and LNDg use option "g" to select for install in 00settingsMenuServices.sh.

  OPTIONS+=(g 'LightningTipBot' ${lightningtipbot})
fi

# just available for LND
if [ "${lightning}" == "lnd" ] || [ "${lnd}" == "on" ]; then
  OPTIONS+=(r 'LND RTL Webinterface' ${rtlWebinterface})
  OPTIONS+=(t 'LND ThunderHub' ${thunderhub})
  OPTIONS+=(l 'LND LIT (loop, pool, faraday)' ${lit})
  OPTIONS+=(g 'LND LNDg (auto-rebalance, auto-fees)' ${lndg})

This results in installing LightningTipBot when you want to install LNDg, and then it showing that both are installed. PR incoming to fix today.

rootzoll:
Thx. PR merged. Will be part of v1.9 release :)

PR #3538, opened 2022-12-17
Update 00settingsMenuServices.sh

allyourbankarebelongtous:
Fixes a conflict between LightningTipBot and LNDg install on the SERVICES menu. Both LNDg and LightningTipBot were using "g" as the character to identify selection, resulting in the installation of both services when one is selected. Changes the identifier for LNDg to a ü.

### DIFF
--- a/home.admin/00settingsMenuServices.sh
+++ b/home.admin/00settingsMenuServices.sh
@@ -66,7 +66,7 @@ if [ "${lightning}" == "lnd" ] || [ "${lnd}" == "on" ]; then
   OPTIONS+=(r 'LND RTL Webinterface' ${rtlWebinterface})
   OPTIONS+=(t 'LND ThunderHub' ${thunderhub})
   OPTIONS+=(l 'LND LIT (loop, pool, faraday)' ${lit})
-  OPTIONS+=(g 'LND LNDg (auto-rebalance, auto-fees)' ${lndg})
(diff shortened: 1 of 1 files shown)


allyourbankarebelongtous:
Fixes [Issue #3537](https://github.com/rootzoll/raspiblitz/issues/3537)

Tests good

openoms:
Thanks for catching this. As we are running out of characters I think it would be better to use double ASCII characters for the menu identification rather than starting to introduce non-ASCII letters. It can make thing complicated that `ü` is not on the most widely used English keyboards.

Would need to double all letters like adding `a` to the end of all of them to start with, then `b` as necessary like here it could be `ga` for LightningTipBot and `gb` for LNDg.

allyourbankarebelongtous:
> Thanks for catching this. As we are running out of characters I think it would be better to use double ASCII characters for the menu identification rather than starting to introduce non-ASCII letters. It can make thing complicated that `ü` is not on the most widely used English keyboards.
> 
> Would need to double all letters like adding `a` to the end of all of them to start with, then `b` as necessary like here it could be `ga` for LightningTipBot and `gb` for LNDg.

Incoming!

allyourbankarebelongtous:
Changed as suggested. Tests good on the nuc.

openoms (approved):
LGTM

rootzoll (approved):

PR #3574, opened 2022-12-24
Proxmox installation guide update to 1.9

cercatrova21:
tested on my proxmox host, works


rootzoll (approved):

rootzoll:
@cercatrova21 is there a reason why not use
```
sudo bash build_sdcard.sh -f 0 -d headless
```
so it will run without fatpack & without any display ... I mean without fatpack also the WebUI would not be available for setup (can be activated later) but for a mor pro setup like with proxmox I would recommend the SSH update anyway for a time being.

cercatrova21:
Good point, will add that as optional point in the guide

PR #3597, opened 2023-01-09
Update LNDg to latest version 1.5.0

allyourbankarebelongtous:
PR to update the LNDg install script bonus.lndg.sh to version 1.5.0. Tests good for me.


allyourbankarebelongtous:
Second commit fixes an issue a user had who had added lndg with the PR prior to it being incorporated. When the user went to update raspiblitz to v1.9, lndg installed without an issue but couldn't read the database at /mnt/hdd/app-data. After troubleshooting, the problem was that the owner of /mnt/hdd/app-data/lndg was mempool for some reason. Adds a couple of commands to set the user and permissions of old databases in case they are messed up prior to installing lndg when doing a new install but using an existing database.

allyourbankarebelongtous:
Last commit fixes indents

rootzoll (approved):
LGTM - still needs one quick install test

Issue #3468, opened 2022-11-22
Electrs 0.9.11


maximotodev:
![Screenshot from 2022-11-22 11-03-23](https://user-images.githubusercontent.com/94984249/203363145-641347b1-80ad-461e-a948-b2f0bc91a8a4.png)
![Screenshot from 2022-11-22 11-02-04](https://user-images.githubusercontent.com/94984249/203363150-15b957e2-d1eb-4a3f-9589-579e2eef87c1.png)

cwgriswold:
Update to 0.9.9 or 0.9.10

rootzoll:
Electrs 0.9.10 will be part of next RaspiBlitz 1.9.0 release. Merged. Final testing.

ShogunWarlord:
> Update to 0.9.9 or 0.9.10

Can you show me how to update it?

maximotodev:
> Update to 0.9.9 or 0.9.10

Any documentation on how to do it?

willheisenberg:
But still if you do that and I try to connect electrs, i got this error:

######## How to Connect to Electrum Rust Server #######

error='missing second parameter - see help'


so i dont get the onion address

rootzoll:
TODO: Test connection

jukeboxnode:
having same issues and wondering is there a upgrade script we could try out?

rootzoll:
So if anyone cant wait for electrs v0.9.10 coming with the next release - you can try the following manual command (not tested):
```
cd /home/electrs/electrs
sudo -u electrs git fetch
sudo -u electrs git pull -p
sudo -u electrs git reset --hard v0.9.10
sudo -u electrs /home/electrs/.cargo/bin/cargo build --locked --release
sudo systemctl start electrs
```

cwgriswold:
Update to 0.9.10 worked

rootzoll:
TODO: Had on quicktesting 1.9.0RC1 a problem to connect to ElectrumApp (did not had time to check deeper) - so up for retest for RC2.

Tracachang:
Updated manually to Electrs v0.9.11, tested with Electrum and Sparrow without issues. We may set 0.9.11 in 1.90 release?

rootzoll:
testing v0.9.11 install right now

rootzoll:
PR now open for v0.9.11 update: #3613

rootzoll:
On testmachine I get status after Index is build:
```
[...]

PR #3613, opened 2023-01-15
#3468 Update electrs to 0.9.11

rootzoll:
see https://github.com/rootzoll/raspiblitz/issues/3468#issuecomment-1382839908

worked on my test machine


openoms:
Install looks good on Rpi and amd64

Thinking that we could add verification to the update as well after the git reset:
```
    sudo -u electrs /home/admin/config.scripts/blitz.git-verify.sh \
     "${PGPsigner}" "${PGPpubkeyLink}" "${PGPpubkeyFingerprint}" || exit 1
```

Roman is very consistent on signing all the release commits and tags: https://github.com/romanz/electrs/releases

openoms (approved):

PR #3615, opened 2023-01-16
suez update and improvements

openoms:


rootzoll (approved):
Install & Quick funtion test worked on my testmachine.

PR #3702, opened 2023-03-08
update lnbits to 0.10.2 and use poetry instead of venv

dni:
as the title states, updating lnbits to 0.10.2 and installing/running it with poetry now
i did not test it on a raspiblitz yet ;(

LNbits version 0.10.2 will be releases very soon.

PR #3684, opened 2023-02-17
cln update to v23.02, backup-plugin update, add poetry

openoms:
fixing: #3675 #3676 #3677

on to testing

Issue #3632, opened 2023-01-25
Update lnbits to 0.10.2


rootzoll:
https://github.com/lnbits/lnbits/releases/tag/0.9.6

rootzoll:
Test update with data - especially with new dab scheme

PR #3703, opened 2023-03-08
update lnbits to 0.10.2 and use poetry instead of venv

dni:
follow up on https://github.com/rootzoll/raspiblitz/pull/3702

### DIFF
--- a/home.admin/config.scripts/bonus.lnbits.sh
+++ b/home.admin/config.scripts/bonus.lnbits.sh
@@ -631,7 +631,11 @@ if [ "$1" = "install" ]; then
   cd /home/lnbits/lnbits  || exit 1
 
   # check if poetry in installed, if not install it
-  sudo -u lnbits which poetry || sudo -u lnbits curl -sSL https://install.python-poetry.org | sudo -u lnbits python3 -
+  if ! sudo -u lnbits which poetry; then
+    echo "# install poetry"
+    sudo pip3 install --upgrade pip
+    sudo pip3 install poetry
+  fi
   # do install like this
   sudo -u lnbits poetry install
 
--- a/home.admin/config.scripts/bonus.lnbits.sh
+++ b/home.admin/config.scripts/bonus.lnbits.sh
@@ -724,33 +724,6 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
     exit 1
   fi
 
-<<<<<<< Updated upstream
-=======
-  # add lnbits user
-  echo "*** Add the 'lnbits' user ***"
-  sudo adduser --disabled-password --gecos "" lnbits
-
-  # get optional github parameter
-  githubUser="lnbits"
-  if [ "$3" != "" ]; then
-    githubUser="$3"
-  fi
-
-  # https://github.com/lnbits/lnbits/releases
-  tag="0.10.2"
-  if [ "$4" != "" ]; then
-    tag="$4"
-  fi
-
-  # install from GitHub
-  echo "# get the github code user(${githubUser}) branch(${tag})"
-  sudo rm -r /home/lnbits/lnbits 2>/dev/null
-  cd /home/lnbits
-  sudo -u lnbits git clone ${githubRepo} lnbits
-  cd /home/lnbits/lnbits
-  sudo -u lnbits git checkout ${tag} || exit 1
-
->>>>>>> Stashed changes
   # prepare .env file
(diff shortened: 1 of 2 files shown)


dni:
https://github.com/lnbits/lnbits/releases/tag/0.10.2

home.admin/config.scripts/bonus.lnbits.sh
```diff
@@ -627,24 +622,18 @@ if [ "$1" = "install" ]; then
...
+  sudo -u lnbits git clone https://github.com/${githubUser}/lnbits lnbits
   cd /home/lnbits/lnbits || exit 1
   sudo -u lnbits git checkout ${tag} || exit 1
 
   # to the install
   echo "# installing application dependencies"
   cd /home/lnbits/lnbits  || exit 1
 
+  # check if poetry in installed, if not install it
+  sudo -u lnbits which poetry || sudo -u lnbits curl -sSL https://install.python-poetry.org | sudo -u lnbits python3 -
```
openoms: we use poetry in more apps already namely CLN, backup-plugin and Suez (channel visualization tool). While updating those had difficulties with poetry being installed for different users as system side and poetry venv packages were mixed in unpredictably. I found that since we have python and pip installed under root we should have poetry too to not interfere with permissions. The changes have been merged recently in #3684

so would suggest to use:
```suggestion
  if ! sudo -u lnbits which poetry; then
    echo "# install poetry"
    sudo pip3 install --upgrade pip
    sudo pip3 install poetry
  fi
```

this change doesn't affect the `sudo -u lnbits poetry install` or `poetry run` lines below.

dni: i think its a great change, and checking the binary with `which` is the best thing to fall back to, imho.

dni: but still ```curl -sSL https://install.python-poetry.org | python3 -```is the recommended way to install poetry from their website.

openoms: Yes, this is a problem since the official script installs in a  folder not in the PATH in Debian and we need find a way to reliably add it to the PATH for all users (if installed for root) or for the dedicated user (which is even more problematic to test).

example to install for the `bitcoin` user where it is used already:
```
admin@192.168.3.175:~ ₿ curl -sSL https://install.python-poetry.org | sudo -u bitcoin python3 -
Retrieving Poetry metadata

# Welcome to Poetry!

This will download and install the latest version of Poetry,
a dependency and package manager for Python.

It will add the `poetry` command to Poetry's bin directory, located at:

/home/bitcoin/.local/bin

You can uninstall at any time by executing this script with the --uninstall option,
and these changes will be reverted.

Installing Poetry (1.4.0): Done

Poetry (1.4.0) is installed now. Great!

To get started you need Poetry's bin directory (/home/bitcoin/.local/bin) in your `PATH`
environment variable.

Add `export PATH="/home/bitcoin/.local/bin:$PATH"` to your shell configuration file.

Alternatively, you can call Poetry explicitly with `/home/bitcoin/.local/bin/poetry`.

You can test that everything is set up by executing:

`poetry --version`
```
Testing the PATH settings:
```
admin@192.168.3.175:~ ₿ sudo -u bitcoin which poetry; echo $?
1

admin@192.168.3.175:~ ₿ export PATH="/home/bitcoin/.local/bin:$PATH"
admin@192.168.3.175:~ ₿ sudo -u bitcoin which poetry; echo $?
1

admin@192.168.3.175:~ ₿ echo 'export PATH="/home/bitcoin/.local/bin:$PATH"' | sudo tee -a /home/bitcoin/.bashrc
admin@192.168.3.175:~ ₿ sudo -u bitcoin which poetry; echo $?
1
```

This would not make poetry accessible for the lnbits user PATH either.

openoms: the only working alternative seems that we would need to use full paths for poetry everywhere

(maybe a custom symlink when poetry is installed for root, but I am not sure if this way the dependencies of all users running poetry would be kept separate at all?)

rootzoll (approved):
PR installs on my test machine LGTM

rootzoll:
also relates to #3632

ktecho:
Any way to install this version into v1.8? Thanks!

rootzoll:
We hope to get the v1.9.0 out soon - thats what our time focus is on right now.

Issue #1186, opened 2020-05-16
Full node interface for standard banking software via FinTS


ghost:
**A full node is the equivalent of a private bank.**

It has different bank accounts depending on which wallets or programs are used on the node, e.g:
1 wallet bitcoin core
1 wallet lightning lnd onchain
1 wallet lightning lnd offchain
1 wallet btcpayserver internal
etc.

Especially for merchants easy bookkeeping is essential if they want to use bitcoin payments, but also as a private user it´s easy to get lost with different wallets and balances.

BTCPay server has options to export invoices (manual actions most bookkeepers charge expensive) or to connect to QuickBooks Online, which is quite nice:
https://docs.btcpayserver.org/Accounting/

But there are lots of different home banking and accounting programs (e.g. https://shop.lexware.de/persoenliche-finanzen or https://shop.lexware.de/buchhaltung). In Europe all of these use a standard protocol FinTS to connect to the bank and accounts, manage transfers, get balances etc.:
https://en.wikipedia.org/wiki/FinTS
https://www.hbci-zka.de/spec/4_1.htm (sorry it´s in german)

To give merchants, bookkeepers and normal end users the same user experience like they have with their ordinary bank accounts they should be able to connect to their private bank (full node) via FinTS get a list of accounts and be able to import all movements from these accounts into their finance software (sending transfers would be even better, but we leave that for later).

So what is needed is a server side integration of FinTS (normally done by banks), which connects to the different application like bitcoin core, BTCPay or lnd.

On the client side there are different open source projects of the FinTS protocol available:
https://python-fints.readthedocs.io/en/latest/
https://github.com/mrklintscher/libfintx
https://github.com/nemiah/phpFinTS

But on the server side there doesn´t seem to be an open source solutions, only commercial ones like this: https://subsembly.com/fints-server.html

Something like this could be implemented maybe in a hackathon?
Supported applications on the server side could be added step by step, maybe starting with lnd and btcpayserver?

A FinTS protocol interface for full nodes would be huge step to better integrate Bitcoin/Lightning in standard finance software.

rootzoll:
It would really be great to be able to manage your node within your standard bookkeeping software just like a normal foreign currency account by supporting the FinTS. For merchants that want to make Bitcoin on Lightning a normal part of their business - such interfaces are highly needed, so that accounting can be automated as much as possible. 

[...]

PR #3704, opened 2023-03-10
#1186 FinTS/HBCI interface

rootzoll:
First draft of install script - still testing, but should be part 1.9.0RC3


rootzoll:
merging what we have so far for rc3 - to be improved for final release

PR #3708, opened 2023-03-16
Remove extra open bracket in line 31 of bonus.go.sh

flyerbear:
Line 31, `elif [ ${goOSversion{} = "arm64" ]; then` has an extra open bracket causing the script to fail if arm64 is the local architecture and this line is executed. Removing the extra bracket will fix the code.


openoms:
Thank you, but this is already fixed in the `dev` branch (also where the new PRs should point to).

flyerbear:
Understood, sorry for applying to the wrong branch. I'll stick with `dev` in the future.

PR #3711, opened 2023-03-18
fix apt update Key error for influx repo

PatrickScheich:
when installing telegraf via `bonus.telegraf.sh` the occurs an error during `apt update` as the key of the influx repo is unknown
further more, when using smartmontools within telegraf the telegraf-user needs to be in sudoers group


rootzoll (approved):

rootzoll:
thx will be part of v1.9.0rc3 release

PR #3712, opened 2023-03-18
fix missing timeout value for nc cmnd

PatrickScheich:
minor fix: some `nc` cmnd had a missing timeout `-w` value


rootzoll (approved):

rootzoll:
good catch. thx. will be part of v1.9.0rc3

Issue #3706, opened 2023-03-13
Ship a patched CLN 23.02


Semisol:
Due to CLN requiring the raw description to an invoice that has a description hash, LNbits and BTCPayServer are breaking.

There are 3 solutions:
1. Patch CLN. This is the most straightforward method. I have also tested it.
2. Patch LNbits and BTCPayServer. Would be too time consuming, and still, it requires you have access to the description field, which 99% of the time the app doesn't.
3. Ship an outdated version. This is not good for compatibility and security reasons, and makes you miss out on new features.

Here's a tested patch:
```diff
diff --git a/plugins/pay.c b/plugins/pay.c
index 268e4fca1..bcfdc9a88 100644
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1073,7 +1073,7 @@ static struct command_result *json_pay(struct command *cmd,
                                                    JSONRPC2_INVALID_PARAMS,
                                                    "Invalid bolt11: missing description");
                        }
-                       if (!description)
+                       if (!description && false)
                                return command_fail(cmd,
                                                    JSONRPC2_INVALID_PARAMS,
                                                    "bolt11 uses description_hash, but you did not provide description parameter");
```

openoms:
I suggest to use `sed` to change that line after downloading and PGP verifying the code for maximum transparency (don't maintain a patched fork).

callebtc:
Great to see this. The breaking change is [here](https://github.com/ElementsProject/lightning/commit/2afe7a1856b04177898e2dea6b1050d4c8477d87).

openoms:
The
``` 
sudo -u bitcoin sed      .....            plugins/pay.c
```
line can be put here:
https://github.com/rootzoll/raspiblitz/blob/1ae636b265a21edf2102787e5bbff0984727fc1c/home.admin/config.scripts/cl.install.sh#L137

so either change can be tested and PRs are welcome.

Patch to the `dev` branch and after the install script is edited uninstall the CLN binaries:
```
config.scripts/cl.install.sh off purge
```
install and switch the service on:
```
config.scripts/cl.install.sh on
```

Semisol:
> The
> 
> ```
> sudo -u bitcoin sed      .....            plugins/pay.c
> ```
> 
> line can be put here:
> 
> https://github.com/rootzoll/raspiblitz/blob/1ae636b265a21edf2102787e5bbff0984727fc1c/home.admin/config.scripts/cl.install.sh#L137
[...]

PR #3716, opened 2023-03-21
#3706 Update CLN v23.02.2

rootzoll:


home.admin/config.scripts/cl.install.sh
```diff
@@ -374,6 +374,7 @@ alias ${netprefix}clconf=\"sudo nano ${CLCONF}\"
   fi
 
   # if this is the first lightning mainnet turned on - make default
+  [ "${lightning}" == "none" ] && lightning=""
```
openoms, line 377: something is missing here

rootzoll: its a shortform - only if the first is true the second will be executed. worked on testing, but if its not valid to some standard we con change to long form

openoms: ok, got it. it is just catching the case when lightning would be set to `none` and uses an empty string as the value instead

openoms (approved):

PR #3738, opened 2023-04-04
LIT update to v0.9.2-alpha

jbrill:
Updates our LIT release to use `v0.9.2-alpha`


jbrill:
@openoms @rootzoll one thing I noticed is `rpcmiddleware.enable=true` must be set now in the lnd.conf file for autopilot to successfully be enabled.

home.admin/assets/lnd.bitcoin.conf
```diff
@@ -24,6 +24,9 @@ rpclisten=0.0.0.0:10009
 # REST open to all connections on Port 8080
 restlisten=0.0.0.0:8080
 
+# RPC middlware enabled
+rpcmiddleware.enable=1
+
```
openoms: Please remove this:
```suggestion
```
It doesn't belong to the default config and only should be added when the autopilot is enabled if needed. Thank you for flagging it, but consider a separate PR for this.

Also as there are headers used it would need to be:
```
[rpcmiddleware]
# Enable the RPC middleware interceptor functionality.
rpcmiddleware.enable=true
```

openoms:
Please rebase to the dev branch as we are not updating v1.8, but moving from `dev` to `v1.9` for the new release.

home.admin/assets/lnd.bitcoin.conf
jbrill: Appreciate the review! I chatted with the team, and the flag should be okay to enable by default. Happy to chat more if you need more details.

More details on the interceptor [here](https://docs.lightning.engineering/lightning-network-tools/lnd/rpc-middleware-interceptor)
Also the PR to enable the flag in Umbrel by default: https://github.com/getumbrel/umbrel-apps/pull/249

openoms: The complication is that we are saving the existing lnd.conf so it needs to be changed for existing users as well.
If it is needed for the autopilot it can be changed only when autopilot activated or in the lnd.check.sh for everyone.
If you want to change these please address it in a separate PR as it is not related to the LIT update.

jbrill: It appears that `v-0.9.0-alpha` cannot be deployed without this flag set (see https://github.com/lightninglabs/lightning-terminal/issues/508). In this case, would it make sense to add the flag within this PR, perhaps to `lnd.check.sh`? Happy to tackle that in a follow-up commit. Thanks 👍

openoms:
lit v0.9.2 is out

home.admin/assets/lnd.bitcoin.conf
openoms: yes the entry with the header:
```
[rpcmiddleware]
rpcmiddleware.enable=true
```
needs to be added to lnd.check.sh (for all nodes) and to the default lnd.conf (for new nodes)

jbrill:
@openoms where can I find the default lnd.conf? I updated the lnd.check.sh with some initial bash checks

home.admin/assets/lnd.bitcoin.conf
```diff
@@ -37,3 +37,7 @@ autopilot.allocation=0.6
 [bolt]
 db.bolt.auto-compact=true
 db.bolt.auto-compact-min-age=672h
+
+[rpcmiddleware]
+# Enable the RPC middleware interceptor functionality.
```
openoms: the default lnd.conf is already changed here

home.admin/config.scripts/lnd.check.sh
```diff
@@ -305,6 +305,39 @@ fi
...
+  fileLines=$(wc -l ${lndConfFile} | cut -d " " -f1)
+  echo "# fileLines(${fileLines})"
+  if [ ${fileLines} -lt ${insertLine} ]; then
+    echo "# adding new line for inserts"
+    echo "
+  " | tee -a ${lndConfFile}
+  fi
+
+  # SET/UPDATE rpcmiddleware.enable
+  setting ${lndConfFile} ${insertLine} "rpcmiddleware\.enable" "true"
```
openoms: does the dot need to be escaped? it is not done elsewhere (see `tor.streamisolation` above)

jbrill: Updated without the escaped dot

rootzoll:
OK test worked now - merging vor v1.9.0rc4

rootzoll (approved):

Issue #3750, opened 2023-04-13
joinmarket: install stops after bitcoind is restarted


openoms:
```
# Installed JoinMarket
# Run: 'sudo /home/admin/config.scripts/bonus.joinmarket.sh on' to configure and switch on
# Enable the multiwallet feature in bitcoin core mainnet and specify wallet.dat
# Switching the bitcoin core wallet on
# Restarting bitcoind
# Create a non-descriptor wallet.dat

```
the install stopped here when the wallet was not present yet and bitcoind restarted.

Will need to wait with running bitcoin-cli until bitcoind is up again

openoms:
could be related to a problem on this particlular (amd64 with ZFS) machine: 
```
2023-04-13T07:05:05Z Using BerkeleyDB version Berkeley DB 4.8.30: (April  9, 2010)
2023-04-13T07:05:05Z Using wallet /mnt/hdd/bitcoin/wallet.dat/wallet.dat
2023-04-13T07:05:05Z BerkeleyEnvironment::Open: LogDir=/mnt/hdd/bitcoin/wallet.dat/database ErrorFile=/mnt/hdd/bitcoin/wallet.dat/db.log
2023-04-13T07:05:05Z init message: Loading wallet…
```

PR #3752, opened 2023-04-13
joinmarket: check for the bitcoin core wallet last

openoms:
#3750 is likely a unique issue, but just moving the wallet check last so it would cause less problems if would happen again


rootzoll (approved):

rootzoll:
LGTM @openoms merge if ready

PR #3758, opened 2023-04-14
Refactor the bitcoin update script with new keys

openoms:
As Wladimir van der Laan is retired from maintaining Bitcoin Core new defaults are needed.
Added Peter Wuille (sipa) as a fallback and Emzy as another example.

Using the method discussed in: https://github.com/bitcoin/bitcoin/issues/25395#issuecomment-1164066943


rootzoll (approved):

rootzoll:
LGTM - but cannot test because all my test nodes are on latest bitcoin verison. Will need to test in the wild.

PR #3767, opened 2023-04-20
minor fixes I found as I ran scripts manually

dipunm:

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -625,14 +625,14 @@ sudo -H /usr/bin/python -m pip install "/home/admin/raspiblitz/home.admin/BlitzP
 
 # make sure lndlibs are patched for compatibility for both Python2 and Python3
 file="/home/admin/config.scripts/lndlibs/lightning_pb2_grpc.py"
-! grep -Fxq "from __future__ import absolute_import" "${file}" && sed -i -E '1 a from __future__ import absolute_import' "${file}"
-! grep -Eq "^from . import.*" "${file}" && sed -i -E 's/^(import.*_pb2)/from . \1/' "${file}"
+! grep -Fxq "from __future__ import absolute_import" "${file}" && sudo sed -i -E '1 a from __future__ import absolute_import' "${file}"
+! grep -Eq "^from . import.*" "${file}" && sudo sed -i -E 's/^(import.*_pb2)/from . \1/' "${file}"
 
 # add /sbin to path for all
 sudo bash -c "echo 'PATH=\$PATH:/sbin' >> /etc/profile"
 
 # replace boot splash image when raspbian
-[ "${baseimage}" = "raspios_arm64" ] && { echo "* replacing boot splash"; sudo cp /home/admin/raspiblitz/pictures/splash.png /usr/share/plymouth/themes/pix/splash.png; }
+[ -d /usr/share/plymouth ] && [ "${baseimage}" = "raspios_arm64" ] && { echo "* replacing boot splash"; sudo cp /home/admin/raspiblitz/pictures/splash.png /usr/share/plymouth/themes/pix/splash.png; }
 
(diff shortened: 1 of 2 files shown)


build_sdcard.sh
```diff
@@ -625,14 +625,14 @@ sudo -H /usr/bin/python -m pip install "/home/admin/raspiblitz/home.admin/BlitzP
 
 # make sure lndlibs are patched for compatibility for both Python2 and Python3
 file="/home/admin/config.scripts/lndlibs/lightning_pb2_grpc.py"
-! grep -Fxq "from __future__ import absolute_import" "${file}" && sed -i -E '1 a from __future__ import absolute_import' "${file}"
-! grep -Eq "^from . import.*" "${file}" && sed -i -E 's/^(import.*_pb2)/from . \1/' "${file}"
+! grep -Fxq "from __future__ import absolute_import" "${file}" && sudo sed -i -E '1 a from __future__ import absolute_import' "${file}"
+! grep -Eq "^from . import.*" "${file}" && sudo sed -i -E 's/^(import.*_pb2)/from . \1/' "${file}"
```
dipunm: Although these lines didn't apply to my setup, I ran them anyway and noticed that they wouldn't work.

All `sudo` changes are actually made because one of my attempts to run this build_sdcard.sh file, I decided to do it as follows:

```bash
> wget https://raw.githubusercontent.com/rootzoll/raspiblitz/v1.8/build_sdcard.sh
> bash build_sdcard.sh
```

I forgot to use `sudo` upon execution and half the script ran, the other did not, and the script did not warn about requiring `sudo` to execute successfully.

build_sdcard.sh
```diff
@@ -625,14 +625,14 @@ sudo -H /usr/bin/python -m pip install "/home/admin/raspiblitz/home.admin/BlitzP
...
 
 # add /sbin to path for all
 sudo bash -c "echo 'PATH=\$PATH:/sbin' >> /etc/profile"
 
 # replace boot splash image when raspbian
-[ "${baseimage}" = "raspios_arm64" ] && { echo "* replacing boot splash"; sudo cp /home/admin/raspiblitz/pictures/splash.png /usr/share/plymouth/themes/pix/splash.png; }
+[ -d /usr/share/plymouth ] && [ "${baseimage}" = "raspios_arm64" ] && { echo "* replacing boot splash"; sudo cp /home/admin/raspiblitz/pictures/splash.png /usr/share/plymouth/themes/pix/splash.png; }
```
dipunm: When starting with raspios lite 64bit, this gave an error because plymouth was never installed. 
That being said, considering that plymouth is uninstalled further up, I'm not sure why this line is useful.

home.admin/config.scripts/blitz.web.sh
```diff
@@ -72,7 +72,7 @@ EOF
   sudo mkdir -p /var/www/letsencrypt/.well-known/acme-challenge
   sudo chown -R admin:www-data /var/www/letsencrypt
   sudo cp -a /home/admin/assets/nginx/www_public/ /var/www/public
-  sudo chown www-data:www-data /var/www/public
+  sudo chown -R www-data:www-data /var/www/public
```
dipunm, line 75: Considering we place files into this folder, I suspect that the `-R` flag is desired here, but curious to learn why not.

build_sdcard.sh
openoms: it is generally expected to run the build script with sudo
if anything we could aim to remove the unnecessary sudo commands and check if the script is called by root

openoms:
Thanks for these! As a release is close we might only change these after the SDcard image for v1.9 is finalized.

build_sdcard.sh
rootzoll: I will remove the sudos and add a sudo check at the beginning

rootzoll (approved):

PR #3770, opened 2023-04-23
[v1.8] fix c-lightning installation

dipunm:
- [ ] That its based against the `dev` brach - not the default release branch. 

^^ NO: This is a bug in the current release, I intend for this to go to the current branch.

Since a security update in git, `sudo make install` will fail due to a line in the makefile at the top that tries to work out the version of the git repository before executing the named command.

Related issue here: https://github.com/ElementsProject/lightning/issues/5189

I looked at a few options:
- ```git config --system --add safe.directory `pwd`'/*'```: This line does not work as git treats `*` as a special character, not a glob.
- ```git config --system --add safe.directory '*'```: works, but feels like a sledge hammer and is hard to reverse without making assumptions about what is already configured.

This PR simply `chown`s the directory (recursively) to `root` and then reverses the action after to `bitcoin`. Git will not reject commands if the folder is owned by the caller of git which in this case is `root`.

I also delayed the `exit 1` step to allow re-assigning `chown` to `bitcoin` before exiting.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -787,6 +787,7 @@ if ${fatpack}; then
 
   # set build code as new default
   sudo rm -r /home/admin/assets/nginx/www_public
+  mkdir -p /home/admin/assets/nginx/www_public
   sudo cp -a /home/blitzapi/blitz_web/build/* /home/admin/assets/nginx/www_public
   sudo chown admin:admin /home/admin/assets/nginx/www_public
   sudo rm -r /home/blitzapi/blitz_web/build/*
--- a/home.admin/config.scripts/cl-plugin.cln-grpc.sh
+++ b/home.admin/config.scripts/cl-plugin.cln-grpc.sh
@@ -35,6 +35,8 @@ function buildGRPCplugin() {
       echo "* Adding Core Lightning ..."
       /home/admin/config.scripts/cl.install.sh install || exit 1
     fi
+    # install required dependency
+    sudo apt-get install protobuf-compiler -y
     # rust for cln-grpc, includes rustfmt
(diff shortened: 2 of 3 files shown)


dipunm:
New commit, fixes the CLN plugin as this is also blocking the completion of `build_sdcard.sh` in `v1.8`. This is fixed in dev: https://github.com/rootzoll/raspiblitz/pull/3684/files#diff-90bd4dc3bc79a0696472ef5b04b46572d610800eb301a1b5a8ff90365b493076R39

openoms:
thanks, not long before we switch to v1.9 as the default but good to maintain the current branch

`utACK`

rootzoll:
Please base PR against `dev` branch - which wil be merged in to v1.9 by us next week and v1.9 will then soon become next default/release branch. Please beware that the fatpack section moved in dev/v1.9 into a seperate script.

rootzoll:
@dipunm With #3778 merged - this PR to be closed?

dipunm:
Personally I think it has value, my position is that it would be good to have non-breaking commits at each version tag, but that isn't practical for ALL tags as things we depend on update, evolve, disappear.

So if you prefer to rely on "the latest tag is the only supported tag", then close this after releasing v1.9 if you are doing that in the next few days.

rootzoll:
Basically I also can see value in fixing old tags - but before merging something into a main version branch I would like to test it I am missing the time ressources now to setup an old build environment based - especially while v1.9.0 release is already in such a delay I need to focus on that. 

So I mark this to be closed once v1.9.0 is out.

openoms:
closing as v1.8 won't be updated any more

PR #3771, opened 2023-04-23
lnproxy update with new specs

openoms:
no webui is installed locally any more


rootzoll:
install test LGTM - inlcuding in v1.9.0rc4

rootzoll (approved):

PR #3774, opened 2023-04-25
lnd-v0.16.1

openoms:
updates of existing nodes wroked ok


rootzoll (approved):

rootzoll:
Thats at least one good outcome of our delayed raspiblitz release, that we already have the bugfixed lnd v0.16.1 as standard.

PR #3776, opened 2023-04-27
fix reverse proxy issue, bind to 0.0.0.0 on ExecStart

dni:
lnbits needs to start and bind on 0.0.0.0, else it will only be reachable fia local and wont work with reverse proxies.

at the moment  host defaults to 127.0.0.1

https://github.com/lnbits/lnbits/issues/1658


openoms:
Can change this, but note that the reverse proxies (eg nginx- already used) is connecting on the localhost.

This change helps users accessing lnbits on HTTP on their LAN.

openoms (approved):

PR #3788, opened 2023-05-05
Update LNbits to 0.10.5

dni:
this is very important update!


dni:
ah, already a PR open, sorry

Issue #3803, opened 2023-05-07
Specter: pin the last working version with Python <3.10 and disable update


openoms:
need to keep  a lower version pinned until https://github.com/rootzoll/raspiblitz/pull/3709 is solved

rootzoll:
is this just a reminder or a TODO for v1.9.0?

openoms:
it is a TODO as uprading to Specter 2.0.0 does not work

PR #3709, opened 2023-03-18
specter update to 2.0.4

openoms:
Adding specter desktop 2.0.0 which needs python 3.10. As it is not available for Debian currently it is installed from the deadsnakes repo and not made default.

- [x] tested the update so far
- [x] need to test the new install

The install runs ok, but needs 
- [x] https://github.com/cryptoadvance/specter-desktop/issues/2393
- [x] https://github.com/cryptoadvance/specter-desktop/pull/2399
to be fixed.

PR #3836, opened 2023-05-15
disable Specter update in menu

openoms:
fixing #3803
Related: #3709

The latest Specter releases depend on Pyhton 3.10 which Debian doesn't have yet: https://github.com/rootzoll/raspiblitz/issues/3803


rootzoll:
On testing this was a v1.9.0rc5 base I still get a python3 error:

Bildschirmfoto 2023-05-15 um 21 31 26

openoms:
we might need to leave specter on the version as is and just disable the update for now

rootzoll (approved):

Issue #3838, opened 2023-05-15
CLN recovery mode is not switched off when finished


openoms:
There are changes needed because unlike with LND the `cl.backup.sh recoverymode off` is not run ever.

rootzoll:
Waiting PR to merge for final v1.9.0

PR #3839, opened 2023-05-15
fix: switch cl recovery mode off when finished

openoms:
fixes: #3838


rootzoll (approved):

rootzoll (approved):

openoms:
This might not cover all edge cases, but don't want it to be a blocker, merging now.

Edit: Test successful after adding a peer to my CL node.

Issue #3864, opened 2023-05-23
Suez  Command not found


openoms:
```
# Showing the channels of cl mainnet - consider reducing the font size (press CTRL- or CMD-)
Command not found: ./suez
```
same problem as in: #3856

rootzoll:
will be merged into v1.9 brach later and be available thru PATCH UPDATE

PR #3865, opened 2023-05-23
fix: use the absolute path of suez

openoms:
fixes: #3864


rootzoll (approved):

rootzoll:
will be merged into v1.9 brach later and be available thru PATCH UPDATE

PR #3975, opened 2023-06-24
Copy lnd rescue files after creating backup device

adevoss:

Copy lnd rescue files after creating backup device


rootzoll:
the scb file is always there - the lnd-rescue back would need to be created first ... risking to fill up the backup drive (when its a small thumb drive) I would not to this automatically here.

PR #3977, opened 2023-06-24
Timestamp in lnd backup file

adevoss:


rootzoll:
The name of the lnd-rescue-backup is already quite long. I can see a benefit in also having a timestamp on there but I would recommend to do that manually if needed by for example archiving the file for example in a diretcory with the date. In normal cases your file system will also give you metadata when the file was created.

PR #3980, opened 2023-06-24
Backup script

adevoss:


rootzoll:
Basically what you are tyring todo here (create a ZIP of all data of the SSD excluding blockchain data) that is what the script `blitz.migration.sh` is doing. https://github.com/raspiblitz/raspiblitz/blob/dev/home.admin/config.scripts/blitz.migration.sh

Issue #3885, opened 2023-05-31
After Update to Raspiblitz 1.9.0, LND stucks at "prepare syny"; RPC service not available


Marcoblue:
I did update Raspiblitz to 1.9.0 with new SD-card and FATPACK today.

After restart I got stuck at (waited 2h):
![grafik](https://github.com/rootzoll/raspiblitz/assets/71592699/1b47d96a-a9d0-468a-b67b-1bf934f95320)

At lndlog I see every second:
2023-05-31 15:30:24.072 [DBG] RPCS: [/lnrpc.Lightning/GetInfo] requested
2023-05-31 15:30:24.072 [ERR] RPCS: [/lnrpc.Lightning/GetInfo]: waiting to start, RPC services not available

My debug log:
https://l.termbin.com/7u0n

Can you help me to find the problem? Thanks.

Marcoblue:
journalctl -u lnd:
May 31 14:17:46 raspberrypi systemd[1]: Starting Lightning Network Daemon...
May 31 14:37:47 raspberrypi systemd[1]: lnd.service: start operation timed out. Terminating.
May 31 15:01:28 raspberrypi lnd[11525]: error creating wallet config: unable to set up wallet password listeners: shutting down
May 31 15:01:28 raspberrypi systemd[1]: lnd.service: Main process exited, code=exited, status=1/FAILURE
May 31 15:01:28 raspberrypi systemd[1]: lnd.service: Failed with result 'timeout'.
May 31 15:01:28 raspberrypi systemd[1]: Failed to start Lightning Network Daemon.
May 31 15:01:28 raspberrypi systemd[1]: lnd.service: Consumed 9min 12.287s CPU time.
May 31 15:02:28 raspberrypi systemd[1]: lnd.service: Scheduled restart job, restart counter is at 1.
May 31 15:02:28 raspberrypi systemd[1]: Stopped Lightning Network Daemon.
May 31 15:02:28 raspberrypi systemd[1]: lnd.service: Consumed 9min 12.287s CPU time.
May 31 15:02:28 raspberrypi systemd[1]: Starting Lightning Network Daemon...
May 31 15:22:29 raspberrypi systemd[1]: lnd.service: start operation timed out. Terminating.
May 31 15:40:27 raspberrypi lnd[254908]: error creating wallet config: unable to set up wallet password listeners: shutting down
[...]

PR #3986, opened 2023-07-02
lnd: Restart=always and clean config

openoms:
Related: #3885

### DIFF
--- a/home.admin/config.scripts/lnd.check.sh
+++ b/home.admin/config.scripts/lnd.check.sh
@@ -332,8 +332,9 @@ if [ "$1" == "prestart" ]; then
 " | tee -a ${lndConfFile}
   fi
 
-  # remove erroneous entries: "  [rpcmiddleware]"
+  # remove erroneous entries
   sudo sed -i '/^  \[rpcmiddleware\]/d' ${lndConfFile}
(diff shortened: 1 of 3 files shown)


home.admin/config.scripts/lnd.check.sh
```diff
@@ -332,6 +332,9 @@ if [ "$1" == "prestart" ]; then
 " | tee -a ${lndConfFile}
   fi
 
+  # remove erroneous entries: "  [rpcmiddleware]"
+  sudo sed -i '/^  \[rpcmiddleware\]/d' ${lndConfFile}
```
cstenglein, line 336: I think this won't remove the [[Rr]pcmiddleware] lines in the config file:

cstenglein (changes requested):

home.admin/config.scripts/lnd.check.sh
openoms: Thanks, can't hurt if we make sure those are gone too.
adding:
```
  sudo sed -i '/^  \[\[Rr\]pcmiddleware\]/d' ${lndConfFile}
```

cstenglein (approved):

rootzoll (approved):

PR #3992, opened 2023-07-06
prevent caching of index.html, add cache-control header for other req…

cstenglein:
When upgrading the RaspiBlitz, some users get an error with the webUI since the cached index.html in the browser tries to load resources which don't exist anymore, see e.g. https://github.com/cstenglein/raspiblitz-web/issues/633

The JS & CSS filenames have the hash attached to them, so they are no problem, but the index.html always has the same filename, thus the browser wants to use the cached version.

This PR adds `Cache-Control` header to prevent any caching of the index.html (`no-store`) and tells the browser for other resources, that it should revalidate them (`no-cache, must-revalidate`). Don't be confused with `no-cache` - it does caching, it's a misnomer, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

### DIFF
--- a/home.admin/assets/nginx/sites-available/public.conf
+++ b/home.admin/assets/nginx/sites-available/public.conf
@@ -32,6 +32,13 @@ server {
 		root /var/www/letsencrypt;
 	}
 
+	location = /index.html {
+		# internal means that it can't be called from outside
+		internal;
+		# add no-store to prevent caching of index.html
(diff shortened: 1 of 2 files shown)


cstenglein:
@rootzoll would like to have this included in the next version please

rootzoll (approved):

PR #4010, opened 2023-07-11
Adjust lnd.conf for healthcheck parameters, allowing for longer timeout intervalls for rpc connection

TrezorHannes:

We've observed a number of lnd.service restarts due to health monitor shutting down.
eg [link1](https://l.termbin.com/fkdeo), {to be complemented by others}
This is presumably happening due to high CPU / Memory load, LND loses it's connection with bitcoind and shutting down.

With the proposed PR, we're introducing a more lax yet not reckless latency allowance for LND to 
 - have a longer timeout window
 - try again 3 times
 - have an extended interval time between those 3 attempts to wait out potential temporary CPU / harddisk peak times

This PR intentionally omits a necessary `-/home/admin/config.scripts/lnd.check.sh` adjustment, I want to propose the changes first, and leave the bash magic to alter the sed changes for someone who knows his / her stuff :wink: 

We've had 3-4 raspi-noderunners successfully turning an unstable LND into a stable instance by adding those settings. Happy to debate and take on questions / challenging the settings.


fre33o:
With the addition of the healthcheck parameters in lnd.conf, the crash problems have gone away. LND runs again without problems. Thanks

openoms:
Hopefully the newer LND versions will have less CPU load, but these are welcome changes for the resource constrained environments (RPi-s) and shouldn't make matters worse in others.

`conceptACK`

needs `lnd.check.sh` additions as discussed

PR #4032, opened 2023-07-26
Nodejs update to v20.x with apt

openoms:
node and yarn binaries are in the PATH now:
```
₿ which node
/usr/bin/node
₿ which yarn
/usr/bin/yarn
```
affects the webUI install


cstenglein:
Will check on the webUI repo with node 20 and report back

cstenglein:
Webui build works with node 20  👍

rootzoll (approved):
We will have to see if the rest of WebApps will work during RC phase.

Issue #3983, opened 2023-06-29
add overcommit_memory to redis config


cstenglein:
From the redis logs under `/var/log/redis/redis-server.log`:

```
WARNING overcommit_memory is set to 0! Background save may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
```

See also: https://redis.io/docs/management/admin/

fusion44:
I think this only applies if we were persisting the DB to disk, which AFAIK we're not.

rootzoll:
_When overcommit_memory is set to 0 (the default), the kernel uses a heuristic algorithm to determine whether to allow a process to allocate more memory than is available. This can lead to situations where a process is allowed to allocate more memory than is physically available, which can cause the system to become unstable or crash._

_When overcommit_memory is set to 1, the kernel will always allow a process to allocate memory up to the limit of its address space, regardless of how much physical memory is available. This can prevent the system from becoming unstable or crashing due to memory allocation issues, but it can also cause processes to fail if they try to allocate more memory than is available._

To set overcommit_memory=1 seems like the more stable option.

rootzoll:
@openoms can you recommend the changes in #4041

openoms:
Agree, individual processes failing is the less evil. They can restart and allow time to diagnose the memory overallocation problem.

If we can we should add warnings if anything crashes due to OOM (out of memory) events.

rootzoll:
PR merged - now on final testing

PR #4033, opened 2023-07-26
ci: UEFI image generation and build updates

openoms:
switches the `amd64-lean-image-build` gh action to build an UEFI bootable image with Gnome desktop

adds a `amd64-lean-legacyboot-image-build` action to build a headless image with legacy boot for older hardware

refactored the amd64 build heavily, converted from json to a more standard and flexible pkr.hcl file

updates to Debian 12 base for amd64

ci readme updates

fixes: https://github.com/raspiblitz/raspiblitz/issues/4029
fixes: https://github.com/raspiblitz/raspiblitz/issues/3509
related: https://github.com/raspiblitz/raspiblitz/issues/3053

PR #4041, opened 2023-07-29
vm.overcommit_memory=1 & increase fs.inotify.max_user_instances

rootzoll:
Change some system settings to prevent freeze/crash of system:

- minimum is now 2GB RAM on RP4 (since 2020) so I thing we can double the fs.inotify.max_user_instances
- activate `vm.overcommit_memory = 1` as discussed in #3983


openoms:
Don't worry about the amd64 build failing, Debian is on 12.1 now. Updated in #4033

openoms (approved):

openoms:
rebased on the updated image builds

PR #3979, opened 2023-06-24
Copy lnd backup file to backup device after backup

adevoss:
Thank you for your contribution to RaspiBlitz. Before submitting this PR, please make sure:
- [x] That its based against the `dev` brach - not the default release branch.

PR #4053, opened 2023-08-01
lndbackup also using backup device if connected

rootzoll:
final merge into dev from #3979 with added free space check


rootzoll:
tested & ready to merge

PR #4109, opened 2023-08-16
(background.scan): only init values if they do not exist

cstenglein:
~~Not yet tested.~~ Script works.

Just came across this. The `_background.scan.sh` always inits the values, thus resetting them every time and afterwards sets them again.

Thus I updated `_cache.sh` to enable the `NX` parameter (see https://redis.io/commands/set/) to only set these initial values if they do not exist already.

Since saving these values is not a really expensive operation, I'm not sure if this change is meaningful enough.

The `_cache.sh` update is kinda hacky, don't know how to make it better without changing too much.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
 <!-- omit in toc -->
-# ![RaspiBlitz](pictures/raspilogo_tile_400px.png) 
+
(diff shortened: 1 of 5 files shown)


rootzoll:
I rewrote the NX extra parameter to "_cache.sh init" - merged for testing.

PR #4114, opened 2023-08-17
swapfile size increase to 10GB

openoms:
Proposing to increase the swapfile on the SSD to 10GB as the SSD space is cheap and filling up the memory and the swap can lead to the unexplained freezes seen on some 4GB RPi-s.

### DIFF
--- a/home.admin/config.scripts/blitz.datadrive.sh
+++ b/home.admin/config.scripts/blitz.datadrive.sh
@@ -1577,9 +1577,9 @@ if [ "$1" = "swap" ]; then
       sed -i "s/^CONF_SWAPFILE=.*/CONF_SWAPFILE=\/mnt\/hdd\/swapfile/g" /etc/dphys-swapfile  
     fi
     sed -i "s/^CONF_SWAPSIZE=/#CONF_SWAPSIZE=/g" /etc/dphys-swapfile 
-    sed -i "s/^#CONF_MAXSWAP=.*/CONF_MAXSWAP=3072/g" /etc/dphys-swapfile
+    sed -i "s/^#CONF_MAXSWAP=.*/CONF_MAXSWAP=10240/g" /etc/dphys-swapfile
     >&2 echo "# Creating SWAP file .."
-    dd if=/dev/zero of=$externalSwapPath count=3072 bs=1MiB 1>/dev/null
+    dd if=/dev/zero of=$externalSwapPath count=10240 bs=1MiB 1>/dev/null
     chmod 0600 $externalSwapPath 1>/dev/null
     >&2 echo "# Activating new SWAP"
     mkswap $externalSwapPath


cstenglein:
Wouldn't setting `CONF_SWAPFACTOR=2` the better option?

According to the docs:

```
# set size to computed value, this times RAM size, dynamically adapts,
#   guarantees that there is enough swap without wasting disk space on excess
```

Setting it to 2 would result in `8gb` for the 4GB pi and `16gb` for the 8gb model on maximum cause it dynamically adapts.
We could also set this to 3 on the 4gb model, so it's always 10gb+ but dynamically.


===

The docs on the `CONF_MAXSWAP` says it's absolute and restricted: 

```
# restrict size (computed and absolute!) to maximally this limit
#   can be set to empty for no limit, but beware of filled partitions!
#   this is/was a (outdated?) 32bit kernel limit (in MBytes), do not overrun it
#   but is also sensible on 64bit to prevent filling /var or even / partition
```

openoms:
Given the same apps and usage the swap size would need to be higher the on the smaller RAM boards, not proportional.

Giving it a fixed value is the simplest option. 10GB is 1% of the typical 1TB disk.

Also I am not sure of the behaviour of the dynamic allocation. Does it increase on demand? Should still leave the maximum allocated space available so the outcome is the same as reserving it.

Generally having less than 10% space available on the disk would warrant an upgrade.

cstenglein (approved):
I'm fine with either solutions :+1:

home.admin/config.scripts/blitz.datadrive.sh
```diff
@@ -1577,9 +1577,9 @@ if [ "$1" = "swap" ]; then
       sed -i "s/^CONF_SWAPFILE=.*/CONF_SWAPFILE=\/mnt\/hdd\/swapfile/g" /etc/dphys-swapfile  
     fi
     sed -i "s/^CONF_SWAPSIZE=/#CONF_SWAPSIZE=/g" /etc/dphys-swapfile 
-    sed -i "s/^#CONF_MAXSWAP=.*/CONF_MAXSWAP=3072/g" /etc/dphys-swapfile
+    sed -i "s/^#CONF_MAXSWAP=.*/CONF_MAXSWAP=10240/g" /etc/dphys-swapfile
     >&2 echo "# Creating SWAP file .."
-    dd if=/dev/zero of=$externalSwapPath count=3072 bs=1MiB 1>/dev/null
```
rootzoll, line 1582: @openoms what will happen to an old swap file on recover .. will it just get overwritten and get bigger?

rootzoll (approved):

home.admin/config.scripts/blitz.datadrive.sh
rootzoll: We will find out in testing - see #4131

PR #4115, opened 2023-08-17
electrs: control the log level in the conf only

openoms:
It is a returning question to enable electrs logs and the systemd setting is unexpected so better to remove. See: https://t.me/raspiblitz/134975

Reducing the level in the electrs config to WARN.

Other options:
https://docs.rs/env_logger/latest/env_logger/#enabling-logging
```
error
warn
info
debug
trace
```


chrisguida:
Why the change to `WARN`? `INFO` is already quite sparse

rootzoll (approved):

chrisguida:
hello?

rootzoll:
If openoms has experience from telegram suipport that this runs more stable for people I merge it - maybe electrs can run in edge cases where its spams logs on info level.

PR #4124, opened 2023-08-21
(build): set LC_ALL to C

cstenglein:
Untested.

This sets `LC_ALL` to `C` and *should* result in a speed-up for all bash scripts using e.g. `grep`.

Most likely needs some testing, since the locale is changed and this could change search results in greps.

See the following resources:

- https://news.ycombinator.com/item?id=37200587
- https://unix.stackexchange.com/questions/87745/what-does-lc-all-c-do (explanation)
- https://www.kurokatta.org/grumble/2019/03/grep-c-locale-speed
- https://www.inmotionhosting.com/support/website/speed-up-grep-searches-with-lc-all/

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -106,7 +106,7 @@ get_arg(){
 }
 
 ## hacky getopts
-## 1. if the option requires argument, and the option is preceeded by single or double dash and it
+## 1. if the option requires an argument, and the option is preceeded by single or double dash and it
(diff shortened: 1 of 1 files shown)


rootzoll:
This changes the sdcard build script - @cstenglein if not critical I will push this to v1.11.0 when new sd card will be build.

cstenglein:
Sure no worries
Just wanted to let you know, most likely it will need some more testing anyway.

openoms:
Running the image with these changes on amd64. No errors seen. Approving.

openoms (approved):

PR #4152, opened 2023-09-05
handle nvme drives when using ext4

Schnitzel:
I ran into an issue that when using NVME drives with ext4 they cannot be partitioned and formatted as the nvme drives are creatd in the format `nvme0n1p1` while the code right now looks for `nvme0n11` (see missing `p`). The attached code fixes this with existing code used for btrfs.


rootzoll (approved):

rootzoll:
Thanks - will merge.

PR #4153, opened 2023-09-05
Fix pip error: This environment is externally managed on Debian 12

Schnitzel:
I'm currently working on installing raspiblitz on a debian12 linux notebook and ran into some issues, mostly around pip and home directories, this PR fixes them and I can install raspi on a fresh debian 12

- remove python lock which prevents pip from installing modules
- make admin home directory readable

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -375,6 +375,11 @@ else
   exit 1
 fi
 
+# remove any debian python protection from pip installing modules
+if [ "${baseimage}" = "debian" ]; then
+  rm /usr/lib/python3.*/EXTERNALLY-MANAGED
+fi
+
 # make sure /usr/bin/pip exists (and calls pip3 in Debian Buster)
 update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 1
 # 1. libs (for global python scripts)
@@ -549,6 +554,9 @@ if [ $(sudo cat /etc/group | grep -c "^admin") -lt 1 ]; then
 else
   echo -e "\nOK group admin exists"
 fi
+# make admin home directly readable by others, the bitcoin user needs access to it
+chmod 755 /home/admin
+
 
 echo -e "\n*** ADDING SERVICE USER bitcoin"
 # based on https://raspibolt.org/guide/raspberry-pi/system-configuration.html
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -376,7 +376,7 @@ else
 fi
 
 # remove any debian python protection from pip installing modules
-if [ "${baseimage}" = "debian" ]; then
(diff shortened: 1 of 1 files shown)


build_sdcard.sh
```diff
@@ -549,6 +554,9 @@ if [ $(sudo cat /etc/group | grep -c "^admin") -lt 1 ]; then
 else
   echo -e "\nOK group admin exists"
 fi
+# make admin home directly readable by others, the bitcoin user needs access to it
+chmod 755 /home/admin
```
openoms: see the more comprehensive solution to this in: https://github.com/raspiblitz/raspiblitz/pull/4156

build_sdcard.sh
```diff
@@ -375,6 +375,11 @@ else
   exit 1
 fi
 
+# remove any debian python protection from pip installing modules
+if [ "${baseimage}" = "debian" ]; then
+  rm /usr/lib/python3.*/EXTERNALLY-MANAGED
```
openoms, line 380: when did you have the errors needing this change?
Would like to reproduce it.

openoms: can confirm the existence of the file:
```
sudo cat /usr/lib/python3.11/EXTERNALLY-MANAGED 
[externally-managed]
Error=To install Python packages system-wide, try apt install
 python3-xyz, where xyz is the package you are trying to
 install.

 If you wish to install a non-Debian-packaged Python package,
 create a virtual environment using python3 -m venv path/to/venv.
 Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
 sure you have python3-full installed.

 If you wish to install a non-Debian packaged Python application,
 it may be easiest to use pipx install xyz, which will manage a
 virtual environment for you. Make sure you have pipx installed.

 See /usr/share/doc/python3.11/README.venv for more information.
```

openoms: this is discussed extensively in:
https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3
and in: https://stackoverflow.com/questions/75602063/pip-install-r-requirements-txt-is-failing-this-environment-is-externally-mana

The other solution could be adding
```
[global]
break-system-packages = true
```
to the same effect.

I think your solution is ok, but don't need to limit to debian.

Simply if the file is there, remove it.

openoms: applied the suggested change

build_sdcard.sh
openoms: removed this change

openoms (approved):
Tested that this change is currently necessary to use Debian 12 as a base.
Approving to be part of the v1.10 release, but not relevant to the RPi image as it is on Debian 11 still.

Can be merged despite the failing builds (they are fixed elsewhere).

Fixes: #4170

PR #4156, opened 2023-09-09
create home folders as 0755 on Debian 12

openoms:
Changing the user creation in every script from:
```
sudo adduser --disabled-password --gecos "" ${APPID}
```
to
```
sudo adduser --system --group --home /home/${APPID} ${APPID}
```

`man adduser`:
>If called with one non-option argument and the --system option, adduser will add a dynamically allocated
       system user, often abbreviated as system user in the context of the adduser package.
       By  default,  system  users are placed in the nogroup group.  To place the new system user in an already
       existing group, use the --gid or --ingroup options.  If the --group is given and the  identically  named
       group does not already exist, it is created with the same ID.
       If  no  home  directory  is specified, the default home directory for a new system user is /nonexistent.
       This directory should never exist on any Debian system, and adduser will never create it automatically.

`man adduser.conf`
>DIR_MODE
              The  permissions  mode  for  home directories of non-system users that are created by adduser(8).
              Defaults to 0700.  Note that there are potential configurations (such as /~user web services,  or
              in-home mail delivery) which will require changes to the default.  See also SYS_DIR_MODE.

>SYS_DIR_MODE
              The  permissions  mode  for home directories of system users that are created by adduser(8).  De‐
              faults to 0755.  Note that changing the default permissions for system users may cause some pack‐
              ages to behave unreliably, if the program relies on the default setting.  See also DIR_MODE.

Fixes: https://github.com/raspiblitz/raspiblitz/issues/4154

These changes are required because of the change in the default permissions in Debian 12. 
Tested the commands work to the same effect on Debian11 as well.

Does not need to be part of v1.10 as the base raspiOS is on Debian 11 still, but would be good to be part of the `dev` branch as that's where from the manual, amd64 and debian12 RPi images are built.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -392,7 +392,7 @@ echo -e "\n*** PREPARE ${baseimage} ***"
 # make sure the pi user is present
 if [ "$(compgen -u | grep -c pi)" -eq 0 ];then
   echo "# Adding the user pi"
-  adduser --disabled-password --gecos "" pi
+  adduser --system --group --home /home/pi pi
   adduser pi sudo
 fi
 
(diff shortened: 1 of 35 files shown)


rootzoll:
It touches the sd card, but it does not add/removes features, it just future improves - so I am good with merging now.

rootzoll (approved):

PR #4160, opened 2023-09-10
ci: fix builds

openoms:
Related: https://github.com/raspiblitz/raspiblitz/issues/4169


talvasconcelos:
Not sure if this helps, but cchardet version is too old, on @fusion44 blitz_api , latest version is `5.2.0` and blitz_api installs `2.1.7`, according to `pyproject.toml`, also apparently the project seems to have been abandoned. Someone forked it and i think it's Python 3.11 compatible: https://github.com/faust-streaming/cChardet

openoms:
Aiming for these changes to be part of the `dev` branch between releases as they are relevant to the amd64 images built in Actions and the manual builds running the `build_sdcard.sh` on Debian12.

PR #4172, opened 2023-09-14
fix adduser + maintenance tasks

openoms:
Follow up of the changes in: https://github.com/raspiblitz/raspiblitz/pull/4156


talvasconcelos:
Carefully watching this one...

openoms:
Tests are looking good here. 
Only one new restriction to take note (recorded in CHANGES):

Info: the users not intended to be logged in will not be available to change into (manage them from admin with sudo)

If want to log in to a system user (which is not used as a cli tool) can run:
```
sudo usermod -s /bin/bash ${USERNAME}
```

or change relevant entry in /etc/passwd 
from
```
/usr/sbin/nologin
```
to
```
/bin/bash
```

rootzoll (approved):

talvasconcelos:
@openoms , for non fatpack installs (run build_sdcard.sh) on amd64 / Debian 12, this fixes the related issues, right?

openoms:
> @openoms , for non fatpack installs (run build_sdcard.sh) on amd64 / Debian 12, this fixes the related issues, right?
> 

yes, the builds in github actions passed

PR #4180, opened 2023-09-19
Update LNBits to 0.10.10

cercatrova21:
Tested, works


rootzoll:
Its a bit too close to raspiblitz 1.10.0 release .. but will mark for v1.10.1

adevoss:
May I suggest LNBits 0.11.0 (latest)?

marssystems:
> May I suggest LNBits 0.11.0 (latest)?

I second that.

PR #4218, opened 2023-10-15
docs: add instructions to prepare a used OSdisk

openoms:


openoms:
Documentation only and tested extensively, merging.

PR #4224, opened 2023-10-23
bos telegram systemd service broken in v1.10.0 - rm working dir 

bliotti:

bos telegram systemd service broken in v1.10.0 - rm working dir fixes issue. 
dir balanceofsatoshis no longer exists.

### DIFF
--- a/home.admin/config.scripts/bonus.bos.sh
+++ b/home.admin/config.scripts/bonus.bos.sh
@@ -130,7 +130,7 @@ After=network-online.target
 
 [Service]
 Type=simple
-WorkingDirectory=/home/bos/balanceofsatoshis
+WorkingDirectory=/home/bos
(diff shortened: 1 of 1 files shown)


home.admin/config.scripts/bonus.bos.sh
```diff
@@ -130,7 +130,7 @@ After=network-online.target
 
 [Service]
 Type=simple
-WorkingDirectory=/home/bos/balanceofsatoshis
+WorkingDirectory=/home/bos
```
openoms: thanks for the fix.
This is because the bos source code is not downloaded any more as it is installed with npm.
This line can be removed altogether.

openoms:
```suggestion
```

openoms (changes requested):

openoms:
Thank you for catching this, merging with the correction.

openoms (approved):

PR #4241, opened 2023-11-09
Update blitz.web.api.sh

0xakihiko:
This script fails due to not having an upgraded Cython on the latest debians. This allows it to build and run.

The --use-pip517 was not sufficient and would still fail without it on debian 12 for me with cchardet. This fixes the install issues for cchardet.


0xakihiko:
Closed in favor of my other PR which includes another fix

PR #4242, opened 2023-11-09
Broken builds on amd64 for blitz api and core lightning (debian 12)

0xakihiko:
There was a missing dependency for psutil for core lightning.
The --use-pip didn't work correctly on debian 12, and the fix was to upgrade Cython.

### DIFF
--- a/home.admin/config.scripts/blitz.web.api.sh
+++ b/home.admin/config.scripts/blitz.web.api.sh
@@ -17,7 +17,7 @@ if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ] || [ "$1" = "-help" ];
   echo "blitz.web.api.sh on DEFAULT"
   echo "blitz.web.api.sh update-config"
   echo "blitz.web.api.sh update-code [?BRANCH]"
-  echo "blitz.web.api.sh off"
+  echo "blitz.web.api.sh off"f
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


home.admin/config.scripts/blitz.web.api.sh
```diff
@@ -17,7 +17,7 @@ if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ] || [ "$1" = "-help" ];
   echo "blitz.web.api.sh on DEFAULT"
   echo "blitz.web.api.sh update-config"
   echo "blitz.web.api.sh update-code [?BRANCH]"
-  echo "blitz.web.api.sh off"
+  echo "blitz.web.api.sh off"f
```
openoms: Fix a typo
```suggestion
  echo "blitz.web.api.sh off"
```

home.admin/config.scripts/cl.install.sh
```diff
@@ -42,7 +42,7 @@ function installDependencies() {
     libsqlite3-dev python3 net-tools zlib1g-dev libsodium-dev \
     gettext
   # additional requirements
-  sudo apt-get install -y postgresql libpq-dev
+  sudo apt-get install -y postgresql libpq-dev psutil
```
openoms: There is no Debian package called `psutil`. This is a pip package which is installed as a dependency of other packages.
```suggestion
  sudo apt-get install -y postgresql libpq-dev
```

openoms:
Thank you for this, will merge and test as part of #4242

PR #4248, opened 2023-11-14
(web.ui.sh) use npm instead of yarn for installation

cstenglein:
In https://github.com/raspiblitz/raspiblitz-web/pull/661 I'm migrating from yarn to npm so no extra install is necessary.

This adapts the script to use `npm` instead of `yarn`


openoms (approved):
looks straightforward

PR #4256, opened 2023-11-17
Rename Electrs INDEX option to REINDEX

Radiokot:

I suggest renaming the **INDEX** option in the Electrs configuration menu to **REINDEX** to better reflects its meaning. 

Although the option description says _"Delete&Rebuild Index"_, the name "INDEX" suggests that you get into some index settings where you can delete and rebuild the index or do something else. Instead, hitting this option, which is located among the harmless ones, just instantly deletes the index that took 2 days to build, without any confirmation.


openoms (approved):
Makes sense, thank you!

PR #4260, opened 2023-11-18
local build fixes

openoms:


openoms:
only changing the build process here so merging

PR #4265, opened 2023-11-22
Activate i2p more reliably

openoms:
Found on a recover that i2p is not working for bitcoind, aiming to make the activation more reliable here.


rootzoll (approved):
LGTM

Issue #4146, opened 2023-08-31
Remove from SSH menus BitcoinMinds/HomerDashboard


geco91:
As shown in the survey those two projects did not gained much traction in the raspiblitz user base and are not direct Bitcoin/Lightning tools - so to reduce complexity it recommended to at least remove them from the SSH menus and just keep them as "deprecated" ssh config scripts.

ChuckNorrison:
I would recommend extension based behaviour of the blitz menu. The menu could only show entries if the underlying script exists. So some could add a script as extension and than show the menu entry to install/configure.

Than raspiblitz could be delivered in a minimal state and get extended with more script files from a extensions menu.

Otherwise its hard to decide which feature should be dumped afterwards. I think no feature should be dumped after it was added to release version.

rootzoll:
An extensions menu sounds quite complicated - would need some kind of central extension repository server, etc .. I would recommed to keep this simple for now: 

The scripts itself can stay in the config scripts folder - maybe marked as deprecated/not-supported in the header/info section. This way people that really want to use them can do so. 

But config-scripts that can be triggered from the SSH menus signal to new user that this script is an active/maintained part of raspiblitz. So it makes sense to also remove such scripts from there.

rootzoll:
Removing Bitcoinminds & Homer also from provisioning/recover installs. If someone still wants them auto-install then they need to get added to `/mnt/hdd/app-data/custom-installs.sh`manually.

BTW if devs want to keep those config-scripts for the future on raspiblitz they should change from editing the "raspiblitz.conf" to adding them on install to the `custom-installs.sh`.

rootzoll:
OK PR waiting to be merged.

PR #4269, opened 2023-11-22
Deprecate & Remove Apps

rootzoll:
As discussed in #4146 "Homer Dashboard" & "Bitcoinminds.org" ger removed von the menus and internals. And "Itchysats" gets completly removed because its an unmainted and archived project.


openoms (approved):
Agreed as discussed.

PR #4270, opened 2023-11-22
fix gnome login screen: create admin as a non system user

openoms:
Aiming to solve the problem with the recent amd64 builds where: 
* `admin` does not appear on the gnome login screen (created as a system user)
* `pi` does appear as the only user to log in with (created as UID 1000)

So creating admin as a non-system user and relaxing the default permissions for the home dir.

`pi` can be a system user as not used for login.


rootzoll (approved):
LGTM

rootzoll (approved):

openoms:
want to keep the amd64 builds passing with this one, so will only make a minimal change to create the admin user as a non-system user

PR #4273, opened 2023-11-24
Update Fulcrum to 1.9.7

steepdawn974:
This PR updates the Fulcrum bonus script to the latest version: 1.9.7. 

Changes
- set fulcrumVersion="1.9.7"
- Fix broken links to signature files. They have changed upstream
- Switch x86 $build to `x86_64-linux` . The `ub16` build no more has a signature file upstream. 

Tested on:  Debian GNU/Linux 11 (bullseye)

### DIFF
--- a/home.admin/config.scripts/bonus.fulcrum.sh
+++ b/home.admin/config.scripts/bonus.fulcrum.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # https://github.com/cculianu/Fulcrum/releases
-fulcrumVersion="1.9.6"
+fulcrumVersion="1.9.7"
 
 # command info
 if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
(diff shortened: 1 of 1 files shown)


openoms:
Thanks! Tested ok on amd64 Debian12 as well.

openoms (approved):

steepdawn974:
@openoms I noticed that the `blitz.shutdown` REBOOT / OFF options are not aware of fulcrum (or any other custom service for that matter) and do not gracefully shut it down. 

I am not super familiar codebase, but **is there some sort of config file/redis entry that keep track of the installed custom services**? Then I would add some logic to the shudown scripts to also gracefully shutdown these additional sevices.

openoms:
> I am not super familiar codebase, but **is there some sort of config file/redis entry that keep track of the installed custom services**? Then I would add some logic to the shudown scripts to also gracefully shutdown these additional sevices. 

yes, all the installed services are recorded in `/mnt/hdd/raspiblitz.conf` for example:
```
fulcrum=on
```

steepdawn974:
Thanks! 
I just saw that I can also check `sudo ls /etc/systemd/system/` .. 

Will put a few more enhancements together and raise another PR then .

Issue #3114, opened 2022-05-12
shutdown / reboot leads to "unexpected power loss" on SSD


PatrickScheich:
After i had lost a raspiblitz instance to a failing HDD and my last SSD had shown some unrecoverable read errors.

I installed the smartmon-tools on my raspiblitz and tried to figure out if there are any metrics, obtainable via `smartctl -a /dev/sda`, which would help to determine the health status of my SSD.

```
root~# smartctl -a /dev/sda

smartctl 7.2 2020-12-30 r5155 [aarch64-linux-5.10.92-v8+] (local build)
Copyright (C) 2002-20, Bruce Allen, Christian Franke, www.smartmontools.org

=== START OF INFORMATION SECTION ===
Model Family:     Marvell based SanDisk SSDs
Device Model:     SanDisk SSD PLUS 1000GB
Serial Number:    211702A00D3C
LU WWN Device Id: 5 001b44 8babd6476
Firmware Version: UH4504RL
User Capacity:    1,000,204,886,016 bytes [1.00 TB]
Sector Size:      512 bytes logical/physical
Rotation Rate:    Solid State Device
Form Factor:      2.5 inches
TRIM Command:     Available, deterministic
Device is:        In smartctl database [for details use: -P show]
ATA Version is:   ACS-3, ACS-2 T13/2015-D revision 3
SATA Version is:  SATA 3.2, 6.0 Gb/s (current: 6.0 Gb/s)
Local Time is:    Thu May 12 02:07:43 2022 CEST
SMART support is: Available - device has SMART capability.
SMART support is: Enabled

=== START OF READ SMART DATA SECTION ===
SMART overall-health self-assessment test result: PASSED

General SMART Values:
Offline data collection status:  (0x00) Offline data collection activity
                                        was never started.
                                        Auto Offline Data Collection: Disabled.
Self-test execution status:      (   0) The previous self-test routine completed
                                        without error or no self-test has ever
                                        been run.
Total time to complete Offline
data collection:                (  120) seconds.
Offline data collection
capabilities:                    (0x11) SMART execute Offline immediate.
                                        No Auto Offline data collection support.
                                        Suspend Offline collection upon new
                                        command.
                                        No Offline surface scan supported.
                                        Self-test supported.
                                        No Conveyance Self-test supported.
                                        No Selective Self-test supported.
SMART capabilities:            (0x0003) Saves SMART data before entering
                                        power-saving mode.
                                        Supports SMART auto save timer.
[...]

PR #4287, opened 2023-12-01
3114 poweroff

rootzoll:
not solving #3114 but trying to at least unmount HDD before shutting down/restart.

### DIFF
--- a/home.admin/config.scripts/blitz.shutdown.sh
+++ b/home.admin/config.scripts/blitz.shutdown.sh
@@ -16,7 +16,7 @@ if [ "$1" = "reboot" ]; then
   /home/admin/_cache.sh set state "reboot"
   /home/admin/_cache.sh set message "$2"
 else
-  shutdownParams="-h now"
+  shutdownParams="-P -h now"
   echo "Then wait 5 seconds and disconnect power."
   /home/admin/_cache.sh set state "shutdown"
   /home/admin/_cache.sh set message ""
@@ -74,6 +74,15 @@ if [ "${isBTRFS}" == "1" ] && [ "${isMounted}" == "1" ]; then
   sudo btrfs scrub start /mnt/hdd/
 fi
 sync
+
+# unmount HDD - try to kill all processes first #3114
+for pid in $(sudo lsof -t "/mnt/hdd"); do
+    echo "# kill -SIGTERM $pid"
+    sudo kill -SIGTERM $pid    # Send SIGTERM signal
+    sleep 5                    # Wait for the process to terminate
+done
+echo "# unmount /mnt/hdd"
+sudo umount -l "/mnt/hdd"
 
 echo "starting shutdown ..."
 sudo shutdown ${shutdownParams}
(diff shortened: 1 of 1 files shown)


home.admin/config.scripts/blitz.shutdown.sh
```diff
@@ -75,10 +75,20 @@ if [ "${isBTRFS}" == "1" ] && [ "${isMounted}" == "1" ]; then
...
 
+# unmount HDD - try to kill all processes first #3114
+for pid in $(sudo lsof -t "/mnt/hdd"); do
+  process_name=$(ps -p $pid -o comm=)
+  echo "# kill -SIGTERM $pid ($process_name)"
+  sudo kill -SIGTERM $pid # Send SIGTERM signal
+  sleep 5                 # Wait for the process to terminate
+done
+echo "# unmount /mnt/hdd"
+sudo umount "/mnt/hdd"
```
openoms, line 90: Seems that `umount` needs to be forced if fails. I am getting:
```
Green activity light stays dark and LCD turns white when shutdown complete.
Then wait 5 seconds and disconnect power.
-----------------------------------------------
stop electrs - please wait ..
stop lightning - please wait ..
stop bitcoind (1) - please wait ..
Bitcoin Core stopping
stop bitcoind (2) - please wait ..
# kill -SIGTERM 1161 (tor)
# unmount /mnt/hdd
umount: /mnt/hdd: target is busy.
starting shutdown ...

Broadcast message from root@raspberrypi on pts/2 (Mon 2023-12-04 10:30:08 GMT):

The system will power off now!


Broadcast message from root@raspberrypi on pts/2 (Mon 2023-12-04 10:30:08 GMT):

The system will power off now!

WARN: Missing $DBUS_SESSION_BUS_ADDRESS .. 
RUNNING FALLBACK SHUTDOWN .. 
```

openoms:
```suggestion
if ! sudo umount /mnt/hdd; then
  echo "# use 'umount -l /mnt/hdd'"
  sudo umount -l /mnt/hdd
fi
```

rootzoll: I wanted not to force unmount because this could also cause data corruption .. basically the shutdown should take care of forcing an unmount in the end when all other systemd processes already stopped.

So i like your idea much better then my solution here much better then mine in this PR:
https://github.com/raspiblitz/raspiblitz/issues/3114#issuecomment-1837191811

I am open to close this PR in favor of it.

openoms: using `umount -l` did not make any difference in my tests (see https://github.com/raspiblitz/raspiblitz/issues/3114#issuecomment-1838422095), so agree that this is not needed

openoms (approved):
It does make sense to kill the processes using the disk even that doesn't make it unmountable so this could be merged as is.
Then in a follow up PR could turn the the `blitz.shutdown.sh` to the  `ExecStart=` of the clean-shutdown systemd service which then will be running very time when shutdown or reboot is called anywhere.

Issue #3872, opened 2023-05-24
Bitcoind setting tweaks


openoms:
From the logs:
```
Bitcoin Core version v24.0.1 (release build)
Warning: net name 'tor' is deprecated and will be removed in the future. You should use 'onion' instead.
```

`shrinkdebugfile=1`
reduces the bitcoin debug.log size to 10 MB on restart. It could be used alongside logrotate.

PR #4297, opened 2023-12-05
#3872 adding "shrinkdebugfile=1" to bitcoin.conf

rootzoll:
Adresses: #3872 If there is no `shrinkdebugfile=` entry it will now add it to bitcoin.conf when Bitcoin is turned on (eg on recover).

### DIFF
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -1,0 +1,20 @@
+name: Spelling
+
+on:
+  workflow_dispatch:
+  push:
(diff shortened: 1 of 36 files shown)


rootzoll:
OK worked on my recover test. @openoms please review

openoms:
Should we add this to the default bitcoin.conf settings? 

Currently it is only added when the testnets are installed, but also has effect on the mainnet logs

rootzoll:
OK tested with running RaspiBlitzes - merging.

Issue #4304, opened 2023-12-07
[Research] Can RPi5 Power-Button be used to trigger shutdown?


rootzoll:
See if the PRi5 power button works like the optional GPIO pins buttons - see: https://chat.openai.com/share/60f9ce66-67e7-444b-8d8c-7fad893e05cc or if there is some other way to utiluze it.

openoms:
Yes, the power button works on my RPi5, it shuts down seemingly gracefully after like 5-10 seconds when pressed.

With a systemd service running:
```
Before=halt.target shutdown.target reboot.target
```
the safe shutdown script should be run also when the poweroff is triggered via the button.

https://github.com/raspiblitz/raspiblitz/issues/3114#issuecomment-1837191811

Edit:
 couldn't yet make `shutdown -r` or `systemctl reboot` / `shutdown` trigger the `blitz.shutdown.sh` script, but will continue to iterate.

openoms:
From:
https://forums.raspberrypi.com/viewtopic.php?t=359649
from the power button and even from the touchscreen the events can be picked up with:
```
sudo su -
# install evtest
apt install -y evtest

# watch the power button
evtest --grab /dev/input/event0

# touchscreen
evtest --grab /dev/input/event1
```


This works:


```
# place in /etc/systemd/logind.conf
HandlePowerKey=ignore

# restart logind
sudo systemctl restart systemd-logind

# in the sudoers file (sudo visudo)
# add permission for nobody to use sudo with the shutdown script
nobody ALL=(ALL) NOPASSWD: sudo /home/admin/config.scripts/blitz.shutdown.sh

# create /etc/triggerhappy/triggers.d/powerbutton.conf
echo "KEY_POWER    1    "sudo /home/admin/config.scripts/blitz.shutdown.sh" | 
  sudo tee -a sudo nano /etc/triggerhappy/triggers.d/powerbutton.conf

# restart triggerhappy
sudo systemctl  restart triggerhappy

# watch triggerhappy
sudo journalctl -fu triggerhappy

# press do power button to initiate the safe shutdown
```

openoms:
A quick script to test, aimed to be part of the `build_sdcard.sh` later.
Works on the RPi5 for me. Would like to test on amd64 as well (and maybe someone has an Rpi4 with a power button set):

```
# *** SAFE SHUTDOWN ***
# Modify /etc/systemd/logind.conf
if ! systemd-analyze cat-config systemd/logind.conf | 
[...]

PR #4312, opened 2023-12-10
trigger safe shutdown with rpi5 power button

openoms:
implements #4304


home.admin/config.scripts/blitz.shutdown.sh
```diff
@@ -26,82 +41,88 @@ fi
...
+  btrfs scrub start /mnt/hdd/
 fi
 sync
 
 # unmount HDD - try to kill all processes first #3114
 echo "# Killing the processes using /mnt/hdd"
-processesUsingDisk=$(sudo lsof -t "/mnt/hdd")
+processesUsingDisk=$(lsof -t "/mnt/hdd")
 if [ -n "$processesUsingDisk" ]; then
   while read -r pid; do
```
openoms, line 103: @rootzoll please check this while loop doesn't run forever

openoms: it looks ok: https://chat.openai.com/share/3e713490-b15c-4f9e-81e5-afbf31a2cabf

the line which generates the list of processes 
```
processesUsingDisk=$(lsof -t "/mnt/hdd"
```
 is now outside of the loop so every process receives the shutdown signal only once.

rootzoll:
The Loop looks fine. Will make one last test before merge.

rootzoll (approved):

Issue #4318, opened 2023-12-13
raspios_arm64-2023-12-06


rootzoll:
There is a new RaspiOS base image
https://downloads.raspberrypi.org/raspios_arm64/images/raspios_arm64-2023-12-06/

rootzoll:
So far sio good. Closing for ReleaseCandidate.

PR #4319, opened 2023-12-13
raspios_arm64-2023-12-06

rootzoll:
Updates Documentation to new RaspiOS base image #4318


rootzoll:
@openoms how do we change the ACTIONS to use the new image?
https://downloads.raspberrypi.org/raspios_arm64/images/raspios_arm64-2023-12-06/

openoms:
best to change the default link and image hash here in the Packer config file: 
https://github.com/raspiblitz/raspiblitz/blob/dev/ci/arm64-rpi/build.arm64-rpi.pkr.hcl#L4-L5

PR #4320, opened 2023-12-14
Add lndk as a service option

orbitalturtle:
This PR adds LNDK, which runs on top of LND, as a service option: https://github.com/lndk-org/lndk

I borrowed a lot from some of the other bonus scripts, such as LIT and the TEOS plugin. Credit where it's due :).

This is my first PR to Raspiblitz, sorry if I missed anything!


rootzoll:
Install works and service seems to run stable ... nice. But how can i check that its also functioning? Normally an installed service adds itself to the MAINMENU and shows at least some info for the user to help on its further user journey.

@orbitalturtle I can offer two ways to integrate:

A) Add the `menu` function with some more user info and have it part of the MAINMENU.
B) Remove it from the install menu for now and just ship it as an additional config script that informed users can call thru the terminal.

orbitalturtle:
@rootzoll Awesome, thanks for taking a look.

I like the idea of option 1 if possible... For now, LNDK just auto-forwards onion messages, increasing the anonymity set of sending onion messages on the network, improving privacy for those that use them. So if possible, I think it'd be great to have it in the UI make it as easy as possible to flip on & increase that anonymity set.

But... for now this happens automatically in the background without user interaction and I can't think of much else to add to the main menu 😅  Unless we just add a note like `Your node is now forwarding onion messages! Check sudo systemctl status lndk to see if it's running properly.` or something like that. But if that's not really enough, that's understandable.

In the next release, users will be able to pay BOLT 12 offers with the lndk-cli. And, we'll write logs to a file, making it easier to actually check thestatus of LNDK. But for now there isn't much instruction needed, LNDK just kind of runs in the background.

rootzoll:
@orbitalturtle "Your node is now forwarding onion messages! Check sudo systemctl status lndk to see if it's running properly." would be enough for a first menu item .. add a URL to it where people can get more background info and your good.

orbitalturtle:
@rootzoll Sounds good! Just added that menu option and description.

Issue #4259, opened 2023-11-17
[RP5] NVMe Board Support


rootzoll:
https://www.youtube.com/watch?v=EXWu4SUsaY8

kravens:
I've ordered the bottom (long) NVMe hat, so I'll be able to test with this thing soon 😄

openoms:
Handy WIP list of disks tested: https://pipci.jeffgeerling.com//hats/pineberry-pi-hatdrive-top.html

rootzoll:
This tutorial is also very helpfull in understanding how to be able to boot from NVMe:
https://www.jeffgeerling.com/blog/2023/nvme-ssd-boot-raspberry-pi-5

So for the future a FRESH SETUP could work like this:
- user boot up from sd card with fresh raspiblitz image
- new NVMe gets partion like: 65gb ext4 system, rest as data drive
- then sd card get cloned to nvme system partion (rpi-clone)
- sd card stays before NVMe in boot order
- sd card renders itself as un-bootable (see notes below)
- system reboots and runs/setups now from NVMe

And an UPDATE/RECOVER could then work like this:
- write a fresh image to the sd card and boot up
- because sd card is now bootable again it will start the new image
- sd card system will clone itself to NVMe system partion (rpi-clone)
- sd card renders itself as un-bootable (see notes below)
- system reboots and runs/recovers now from NVMe

--- SD CARD RENDERS ITSELF UNBOOTABLE --

It is possible to render a Raspberry Pi's SD card unbootable through a write action, which would cause the device to skip it in the boot order on the next restart. Here's how it can happen:

1. **Corrupting Boot Files**: If critical boot files on the SD card are overwritten or corrupted, the Raspberry Pi will not be able to boot from it. These files include the bootloader, kernel, or other essential files in the boot partition.

2. **Damaging the File System**: Writing actions that damage the file system structure (like the partition table or file allocation table) can make the SD card unbootable. This could happen due to improper formatting or partitioning actions.

3. **Overwriting Critical Configurations**: Altering certain configuration files (like `cmdline.txt` or `config.txt` in the boot partition) with incorrect settings can prevent the Raspberry Pi from booting.

4. **Hardware-Level Damage**: Although less common, a write action that causes excessive wear on the SD card could lead to hardware-level damage, making it unreadable.

5. **Writing Incorrect Boot Code**: Intentionally or accidentally writing incorrect or incompatible boot code to the boot sector of the SD card would result in a failure to boot.

[...]

PR #4322, opened 2023-12-15
#4259 activating nvme hat on sd card build

rootzoll:
adressing #4259


rootzoll:
tested ok on my test build - merged into dev for rc1

PR #4335, opened 2023-12-21
Local release image build with Packer 

rootzoll:
OK I streamlied the packer image build on my decidated Laptop running a Debian LIVE image.

Most things work, but some details to discuss with @openoms 

A) arm64-fatpack does not run thru 

Not with the script I was building - but just by calling the make command the build of the arm64-fatpack fails with:

```
==> arm.raspiblitz-arm64-rpi: https://downloads.raspberrypi.org/raspios_arm64/images/raspios_arm64-2023-12-06/2023-12-05-raspios-bookworm-arm64.img.xz?archive=false&checksum=sha256%3A5c54f0572d61e443a32dfa80aa8d918049814bfc70ab977f2d545eef45f1658e => /home/user/.cache/packer/6754bd243accdb4ae9941ce7d93be7d8d7963068.xz
    arm.raspiblitz-arm64-rpi: unpacking /home/user/.cache/packer/6754bd243accdb4ae9941ce7d93be7d8d7963068.xz to raspiblitz-arm64-rpi-lean.img
    arm.raspiblitz-arm64-rpi: unpacking with custom command: [xz --decompress /tmp/image971241824/6754bd243accdb4ae9941ce7d93be7d8d7963068.xz]
    arm.raspiblitz-arm64-rpi: resizing the image file raspiblitz-arm64-rpi-lean.img to 28G
    arm.raspiblitz-arm64-rpi: expanding partition no. 2 on the image raspiblitz-arm64-rpi-lean.img
==> arm.raspiblitz-arm64-rpi: error while expanding partition exec: "parted": executable file not found in $PATH:
Build 'arm.raspiblitz-arm64-rpi' errored after 6 minutes 27 seconds: build was halted 
```

B) sudo debian user warings

When running the the build for arm64-lean I get a lot of those warinigs .. is that sane? Can it be prevented?

`sudo: unable to resolve host debian: Name or service not known`

C) amd64-lean: qcow2 zipping?

When build an image for x86 .. why not as part of the packer process convert the qcow2 to img? I build a part in my packer.sh script where I take the qcow2.gz - unzip it, convert it to img and zip it again. It works, but feels that this would make more sense of having it as part of the packer make process - or is there a speiacl reason to keep it with qcow2?

### DIFF
--- a/FAQ.dev.md
+++ b/FAQ.dev.md
@@ -9,8 +9,12 @@
 * Make sure the packer script is in root of the 128GB-USBDrive
   * If it is not there download: `curl -O -L https://raw.githubusercontent.com/raspiblitz/raspiblitz/dev/ci/packer.sh`
 * Security read/check script and then start build with (replace parameters):
-* `sudo bash ./packer.sh [BRANCH] [arm|x86] [min|fat]`
-
+* `sudo bash ./packer.sh [BRANCH] [arm|x86] [min|fat] [?LASTCOMMITHASH]`
+  * BRANCH = the branch of which the image should be build
+  * `[arm|x86]` = The architecture the image is targeting
+  * `[min|fat]` = lean or fatpack
+  * LASTCOMMITHASH (optional) = security check & copy the latest commit hash of the branch you want to build
+ 
 ### What is the process of creating a new RaspberryPi SD card image release manually?
 
 Checklist before making a SD card image release:
--- a/FAQ.dev.md
+++ b/FAQ.dev.md
@@ -7,7 +7,8 @@
   * Connect a additional 128GB USB with NFTS formatted 
 * Using Filemanager open the 128GB-USBDrive and right-click "Open in Terminal"
 * Make sure the packer script is in root of the 128GB-USBDrive
-  * If it is not there download: `curl -O -L https://raw.githubusercontent.com/raspiblitz/raspiblitz/dev/ci/packer.sh`
+  * If it is not there download: 
+  * `curl -O -L https://raw.githubusercontent.com/raspiblitz/raspiblitz/dev/ci/packer.sh`
 * Security read/check script and then start build with (replace parameters):
 * `sudo bash ./packer.sh [BRANCH] [arm|x86] [min|fat] [?LASTCOMMITHASH]`
   * `BRANCH` = the branch name on this repo of which the image should be build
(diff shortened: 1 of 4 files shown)


openoms:
> A) arm64-fatpack does not run thru


tested to run the
```
# make arm64-rpi-fatpack-image  
```
from root (after `sudo su`).
The build goes through ok, but there is a problem with the file naming I need to fix:
```
...
Build 'arm.raspiblitz-arm64-rpi' finished after 34 minutes 26 seconds.

==> Wait completed after 34 minutes 26 seconds

==> Builds finished. The artifacts of successful builds are:
--> arm.raspiblitz-arm64-rpi: raspiblitz-arm64-rpi-lean.img
# Compute the checksum of the raw image
cd ci/arm64-rpi/packer-builder-arm  && \
sha256sum raspiblitz-arm64-rpi-fatpack.img > raspiblitz-arm64-rpi-fatpack.img.sha256
sha256sum: raspiblitz-arm64-rpi-fatpack.img: No such file or directory
make: *** [Makefile:110: arm64-rpi-fatpack-image] Error 1


ls -la ci/arm64-rpi/packer-builder-arm/raspiblitz*
-rw-r--r-- 1 root root           0 Dec 22 17:28 ci/arm64-rpi/packer-builder-arm/raspiblitz-arm64-rpi-fatpack.img.sha256
-rw-r--r-- 1 root root 30064771072 Dec 22 17:28 ci/arm64-rpi/packer-builder-arm/raspiblitz-arm64-rpi-lean.img
```
Aiming to open a PR with fix to this branch of yours -> #4337
This adds back the sudo to the make commands as well. 

> B) sudo debian user warings
> 
> When running the the build for arm64-lean I get a lot of those warnings .. is that sane? Can it be prevented?
> 
> `sudo: unable to resolve host debian: Name or service not known`

Was aware of these as they present in the github actions logs as well, but couldn't find the stop all of them apart from stop using sudo.
The errors don't seem to cause any issues so ignored so far.

Might be related to the hostname not being set properly, possibly in the `/etc/hosts` file. This would need to be set inside the docker container so likely would need a change in  https://github.com/mkaczanowski/packer-builder-arm.

> C) amd64-lean: qcow2 zipping?

I think the qcow2 was the smallest size so kept that way as it is very important to be able to fit into the github artifact limits. For the local build there is no such limit so would be good to have an .img outcome. Some who use the image in a VM might find .qcow2 useful so this can be up for discussion.

openoms:
Rebasing on the latest state of `dev`

openoms:
Will run a test build with the new script

PR #4337, opened 2023-12-23
fix set_variables.sh path for arm64-rpi.local

openoms:


openoms:
The fatpack build has finished ok with these settings (was running for 5+ hours)
image

PR #4347, opened 2024-01-02
build(arm64-rpi): make /dev/shm world writable for the build

openoms:
Fixing the display driver install errors in:
https://github.com/raspiblitz/raspiblitz/actions/runs/7371236475/job/20058409304


openoms:
the fix worked so merging

Issue #4247, opened 2023-11-14
No Apps show up after Fatpack installation


rootzoll:
### Discussed in https://github.com/raspiblitz/raspiblitz/discussions/4244

I installed the latest Fatpack RaspiBlitz on my Raspi 4, the Bitcoin node is 100% syncrhonized for some days now.
Unfortunately, no Apps show up in the Web-UI. There is only this yellow rotating half-circle when I click on the Apps button.
Does anyone know what the problem is?</div>

rootzoll:
@fusion44 and @cstenglein .. what can be the issue here and how can we improve App Listing?

cstenglein:
The blitz API should send a `installed_app_status` SSE message on startup.

Until the message arrives, there is the loading screen.

Maybe it's in that script: https://github.com/fusion44/blitz_api/blob/dev/app/apps/impl/raspiblitz.py

rootzoll:
TODO:
@rootzoll take a look at API how it can be made more stable
@cstenglein maybe do a fallback with an API request

fusion44:
The API actually tries to get the app status when sending the warmup data to new SSE clients:
https://github.com/fusion44/blitz_api/blob/9adcc866fb30502253070211604781fcca865c18/app/main.py#L316C13-L316C57

If it works on some machines and on some it doesn't, then it probably is a problem getting the status itself:
https://github.com/fusion44/blitz_api/blob/9adcc866fb30502253070211604781fcca865c18/app/apps/impl/raspiblitz.py#L43C8-L43C8

I think it'd be helpful to call the script status manually on failing machines:
`sh bonus.electrs.sh status`

daviogg:
Sorry, is there a workaround to solve this problem? I'm stuck too, I can't load the apps tab
 raspiblitz v 1.10 and with webUI version 1.3.0-dev

daviogg:
I did not initially install the lighting wallet components. After doing so, the Apps tab is loaded correctly

rootzoll:
TODO: Test v1.10.0 fatpack install with just Bitcoin

helishmopter:
Apologies for double posting this here again (just posted under Apps tab won't load on WebUI (#3608) #4012), but I'm really stuck with this issue..

[...]

PR #4378, opened 2024-01-19
#4247 Fix Fatpack / git-verify with expired keys

rootzoll:


openoms:
Should enforce the output to be English

openoms (approved):
approved with changes

PR #4382, opened 2024-01-24
SFTP StaticChannelBackup

bountiful388:
**Describe the bug**
I believe the sftp upload commands for SCB backups are incorrect

`/home.admin/_background.sh` Lines 528, 529, 634, 635

**Original Code** 
```sh
sftp ${sftpBackupOptions} ${localBackupPath} ${sftpBackupTarget}/
sftp ${sftpBackupOptions} ${localTimestampedPath} ${sftpBackupTarget}/
```

**Possible Fix:**
```sh
sftp ${sftpBackupOptions} ${sftpBackupTarget}/ <<< "put ${localBackupPath}"
sftp ${sftpBackupOptions} ${sftpBackupTarget}/ <<< "put ${localTimestampedPath}"
```

Also: Documentation for "SCP Backup Target" may need to be updated to reflect sftp
https://github.com/raspiblitz/raspiblitz/blob/v1.10/README.md#b-scp-backup-target


rootzoll:
Yes you found an inconsistany thx - but I would recommend to solve it the other way round. Instead of correcting the SFTP command I would change it back to SCP ... that would no need in documentation and would keep old configs valid.

rootzoll (approved):

PR #4389, opened 2024-01-29
fix rtl repo, typos, set off if install fails

openoms:
fixes: https://github.com/raspiblitz/raspiblitz/issues/4388


openoms:
Tested on RPi5 with CLN, RTL install works well now.

rootzoll (approved):
tACK

PR #4397, opened 2024-02-01
(ci): update to github actions v4

cstenglein:

See https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/

Should enable faster artifact upload and fixes deprecation notice.


openoms (approved):
thanks, it doesn't look like it breaks anything

PR #4400, opened 2024-02-06
build: remove externally-managed flag for python

### LINKED ISSUE #4170: Python 3.11 error on Debian 12: This environment is externally managed
Mentioned in:
https://github.com/raspiblitz/raspiblitz/issues/4154#issuecomment-1716008904
and
https://github.com/raspiblitz/raspiblitz/pull/4153

```

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.11/README.venv for more information.```
```

Possible workarounds are discussed extensively in:
https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3
and in: https://stackoverflow.com/questions/75602063/pip-install-r-requirements-txt-is-failing-this-environment-is-externally-mana

Either use in the pip config:
```
[global]
break-system-packages = true
```
or
```
sudo rm /usr/lib/python3.*/EXTERNALLY-MANAGED
```
to the same effect.

openoms:
Test if the fix #4170 is still needed.

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -391,14 +391,6 @@ else
   exit 1
 fi
 
-# don't protect system packages from pip install
-# tracking issue: https://github.com/raspiblitz/raspiblitz/issues/4170
-for PYTHONDIR in /usr/lib/python3.*; do
-  if [ -f "$PYTHONDIR/EXTERNALLY-MANAGED" ]; then
-    rm "$PYTHONDIR/EXTERNALLY-MANAGED"
-  fi
-done
-
 # make sure /usr/bin/pip exists (and calls pip3 in Debian Buster)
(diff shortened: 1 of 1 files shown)


rootzoll (approved):
tACK (tested ci build with recover & python backend works)

PR #4409, opened 2024-02-11
btcpay: add database backup and restore option

openoms:
added the pg backup and restore option similar to LNbits

- [x] test with a migrating BTCPayServer from arm64-RPi to amd64

image

image

image


rootzoll (approved):
untACK .. mergingt to be tested during RC pahse

PR #4410, opened 2024-02-11
revert: build: remove externally-managed flag for python

openoms:
Reverts raspiblitz/raspiblitz#4400
Discussed in #4170


openoms:
Debian v12.5 is out https://www.debian.org/News/2024/20240210 , updating will fix the amd64 builds, so the 404 error is not a blocker here

rootzoll (approved):

PR #4427, opened 2024-02-18
Fix Pyblock 2.2.3

rootzoll:
Latest pyblock 2.4 needs python 3.12.0 while RaspiBlitz has now 3.12.0 ... so we fall back onto pyblock 2.2.3

But no installing the old 2.2.3 thru pypi package it was not working because jq==1.2.2 was no longer working with our updated python environment. So I created a workaround that installes 2.2.3 from github but chnages jq manually to 1.2.3 before running install.


rootzoll:
#4427 merged

PR #4430, opened 2024-02-20
Patch 1

TrezorHannes:
- added log-extension by increasing the log size and number of logfiles.  This is essential, since `loglevel=debug` averages at about 40-60minutes until rotating and discarding after 3 logfiles per default
- added sync-freelist=false option for users who grow beyond > 30 channels. Context [detail can be found here](https://github.com/lightningnetwork/lnd/blob/a3bf2c70e5aabf01c5db536f4cb90d2080c65b05/sample-lnd.conf#L112), but from many many runners I know, going beyond 30 channels with freelist basically degrades your node performance, until at >50 it becomes almost non responsive.


rootzoll:
Good input to better manange the log files. But the template file is no longer the place to set the initial values - most is happening here https://github.com/raspiblitz/raspiblitz/blob/b3a14cfc356fdf753bbf496b9465cb88190f0621/home.admin/config.scripts/lnd.install.sh#L305

And for v1.11.0 we already changed to `debuglevel=info` to reduce logs and tested it so far. So I pushed this to a later release to be refactored and then have a long testing period.

Also marking some TODOS:
- remove old templates, create all from script (see above)
- test new log settings

TrezorHannes:
> Good input to better manange the log files. But the template file is no longer the place to set the initial values - most is happening here
> 
> https://github.com/raspiblitz/raspiblitz/blob/b3a14cfc356fdf753bbf496b9465cb88190f0621/home.admin/config.scripts/lnd.install.sh#L305

Alright thank you, that makes sense. I'll close my PR and move over to the install script instead, reviewing whether we have tuning elements missing for the  v1.11.0 release instead.

Issue #4436, opened 2024-02-24
[API/WebUI] Electrs will not be shown as installed 


rootzoll:
With v1.11.0rc3 (with fixed API) Electrs is installed (thru ssh menues previously) but does but still shows in WebUI as uninstalled.

This is the SSH menus:
![Bildschirmfoto 2024-02-24 um 20 58 06](https://github.com/raspiblitz/raspiblitz/assets/2472225/00e1ca0d-47be-4ede-a4a4-f3b5d3ae7f40)

This is the WebUI:
![Bildschirmfoto 2024-02-24 um 20 58 00](https://github.com/raspiblitz/raspiblitz/assets/2472225/3d883d4a-a043-427b-a6ef-b78d7af1f1b2)

@fusion could this still be a side effect from the API?

rootzoll:
Sidenote: Electrs is still building the index ... maybe thats why its not shown as installed?

fusion44:
This is because every other script returns the installed status as `installed=1`, but electrs returns `serviceInstalled=1`. I'll add a special path for this. We should unify this for the next major release. Not sure why I didn't catch this during development. Doesn't seem to have been changed recently.

Electrs: https://github.com/raspiblitz/raspiblitz/blob/ec67c362f755ad968001d10d7815406cd8309a1e/home.admin/config.scripts/bonus.electrs.sh#L39
RTL: https://github.com/raspiblitz/raspiblitz/blob/ec67c362f755ad968001d10d7815406cd8309a1e/home.admin/config.scripts/bonus.rtl.sh#L71

rootzoll:
@fusion44 I will change script to "installed"

rootzoll:
BTW it was not unified because we didnt prepare the electrs script for fatpack/webui yet. I will go thru.

rootzoll:
@fusion44 normally "installed" for a fatpack/webui app just means the user/codebase is preinstalled but not "on". Thats why there is the value `configured` that indicates if the App is "on" (1) or "off" (0). Thats how it should be when unified ... but it seems we dont have that 100% the same with all WebUI apps yet.

```
configured=1
installed=1
```

But you can also look into the `raspiblitz.conf` with the value `ElectRS` to check if its installed - that is much quicker if you dont need any other values.

rootzoll:
Prepared a PR #4438 to fix the install script - testing at the moment.

fusion44:
Ok perfect. I'll adapt once the pr is merged.


[...]

PR #4438, opened 2024-02-25
Fix Electrs Fatpack/WebUI

rootzoll:
This should address #4436 from the install script side and mke it ready for fatpack/WebUi.

### DIFF
--- a/home.admin/config.scripts/blitz.debug.sh
+++ b/home.admin/config.scripts/blitz.debug.sh
@@ -308,6 +308,8 @@ if [ "${ElectRS}" == "on" ]; then
   echo
   echo "*** ElectRS Status ***"
   sudo /home/admin/config.scripts/bonus.electrs.sh status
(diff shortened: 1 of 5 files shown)


openoms (approved):
utACK

rootzoll:
OK seems to work .. was not able to finish latest test because install failed - but that seems to be related to network issues. The rest looked good. Merging.

info: installing component 'rustfmt'
    Updating crates.io index
error: Unable to update registry `crates-io`

```
Caused by:
  failed to fetch `https://github.com/rust-lang/crates.io-index`

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  SSL error: received early EOF; class=Ssl (16); code=Eof (-20)
```

Issue #4441, opened 2024-02-25
CI: let fatpack image depend on lean image


cstenglein:
If possible, this should drastically reduce build time for the fatpack and with the addition of #4440, it should be possible to build the fatpack in the GitHub ci

PR #4454, opened 2024-03-07
(ci): build fatpack based on lean image

cstenglein:
This is a quick draft I came up with, don't think it will work yet, but trying to get out the idea.

This is the idea described in #4441 

The only difference between the lean and the fat image is the `blitz.fatpack.sh` script. So we could build a lean image and then a fatpack on top of it. 

This should shorten build times for the fatpack and make it (hopefully) possible to build the fatpack in github actions.

### DIFF
--- a/.github/workflows/arm64-rpi-lean-image.yml
+++ b/.github/workflows/arm64-rpi-lean-image.yml
@@ -75,8 +75,8 @@ jobs:
       - name: Maximize build space
         uses: easimon/maximize-build-space@master
         with:
-          root-reserve-mb: 40000
-          temp-reserve-mb: 40000
+          root-reserve-mb: 31000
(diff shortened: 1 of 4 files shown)


rootzoll:
Thats a good idea - hope it works :)

openoms:
If unpacking the image in the action (on the github runner) gets into space issues can use this hack which should be able to accommodate even the 30GB+ unpecked image: https://github.com/easimon/maximize-build-space

Already used for the amd64 image with desktop:
https://github.com/raspiblitz/raspiblitz/blob/b02db0b46c6466fc2d7ac278fdf5518c383b478f/.github/workflows/amd64-lean-image.yml#L34-L38

cstenglein:
Will be continued in #4465

PR #4465, opened 2024-03-10
ci: add base image build

### LINKED ISSUE #4441: CI: let fatpack image depend on lean image

openoms:
add stop file and --image_size 20G

the artifact to be used in: https://github.com/raspiblitz/raspiblitz/pull/4454


.github/workflows/arm64-rpi-base-image.yml
```diff
@@ -0,0 +1,83 @@
+name: arm64-rpi-base-image-build
```
cstenglein, line 1: Looking good, not sure if we should build the lean / fatpack here as well with the `needs` syntax.

cstenglein (approved):

.github/workflows/arm64-rpi-base-image.yml
openoms: I think the next step could be to use the base-image artifact built in this PR in #4454. Would only merge if the added action is useful.

openoms: ok, pulled the changes from #4454 and unifying in the `arm64-rpi-fatpack-image` action

.github/workflows/arm64-rpi-fatpack-image.yml
```diff
@@ -72,13 +50,89 @@ jobs:
...
+            echo "GITHUB_USER=$(echo ${{github.repository}} | cut -d'/' -f1)" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Download base image
+        uses: actions/download-artifact@v4
+        with:
+          name: raspiblitz-arm64-rpi-base
```
cstenglein: The name here must be the same as the name you uploaded it with.
So `raspiblitz-arm64-rpi-image-${{ env.BUILD_DATE }}-${{ env.BUILD_VERSION }}`
But be careful that the `BUILD_DATE` & `BUILD_VERSION` are really the same

rootzoll (approved):
untACK

openoms:
so far the LND, CLN (long complie), web UI and API installations ran through ok.

Joinmarket and BTCPay had problems, will try to test them separately.

The JM install issue is reproduced in so can work on it there: https://github.com/openoms/raspiblitz/pull/82

openoms:
Reverted the expirementation with the fatpack in Actions.

Only generating the base image here as the original scope of the PR which works now.

Will continue to try to the fatpack build work in a separate PR building on the base image.

openoms:
merging as only affecting the github workflows and ci directory

PR #4479, opened 2024-03-16
base image update: raspios_arm64-2024-03-15

openoms:


openoms:
Recovered successfully without any new issues on RPi5 using the lean image from here, merging to be part of the next rc.

PR #4573, opened 2024-05-16
add elements install script for Liquid support

openoms:
adding elements node, service and config similar to bitcoin core.


openoms:
Takes a long time to sync on Tor + I2P only so best to enable clearnet for IBD by adding temporarily to `elementsconf`:
```
onlynet=ipv4
onlynet=ipv6
```
Once synced it had no problem to keep up on Tor and I2P only.

Could also add to Specter Desktop (installed with #3709 ) simply following https://specter-desktop-docs.netlify.app/elements/#liquid-configuration-in-specter:
image

The CLI only install script could be added for wider experimentation.

Note regarding using Specter: likely not functional with Liquid currently: https://github.com/cryptoadvance/specter-desktop/issues/2429

rootzoll (commented):
tACK - tested install and systemd service is running OK after install

rootzoll (approved):

rootzoll:
@openoms might add a note to CHANGES

PR #4579, opened 2024-05-20
Improve VM install & add Bitcoin Testnet Option

rootzoll:
This contains samll edit needed to get it running as a VM iwith UTM/qemu on MacOS. 

I also added a `bitcoin.testnet.sh` script that switches bitcoind internally to testnet while to the other apps on RaspiBlitz it acts like mainnet. So this is not the run testnet in parralel - this is really just for developers to have a small footprint VM to test additional apps with bitcoin. Just Bitcoin ... not lightning for now.


cwiggs:
I tried to use `bitcoin.testnet.sh` today but it seems like it expects the drive and mainnet config files to already exist.  Should I setup the server with mainnet and then run `bitcoin.testnet.sh` after the bitcoin service is running?

I'd also love to add a guide on how to do this to the docs if possible.

Thanks.

rootzoll:
Yes it needs a default setup (drive+mainnet) and can the should the n change it to testnet.

Issue #4551, opened 2024-04-23
Update LND for v1.11.1


rootzoll:
especially with lnd 0.17.5-beta having a fix to work with bitcoind 27

rootzoll:
The idea is to update RaspiBlitz v1.11.1 lnd to 0.17.5 which is just a higher patch lever and allows to set bitcoind recommended update to 27.0 while leaving the base version to bitcoin 26.0.0.

Only breaking changes of bitcoin 27.0 would be ... 

_New mempool.dat Format: The format for the mempool.dat file has been changed to include XOR’ing of transaction contents. This new format is incompatible with previous versions, but a temporary setting -persistmempoolv1 allows users to fall back to the legacy format if needed._

.. so for RaspiBlitz v1.12.0 we should then go with boitcoin 27.0 by default.

rootzoll:
PR #4601 open and tested OK

PR #4601, opened 2024-05-29
Update lnd 0.17.5 & recommended bitcoin update 27.0

rootzoll:
adresses: #4551

### DIFF
--- a/home.admin/00raspiblitz.sh
+++ b/home.admin/00raspiblitz.sh
@@ -185,8 +185,14 @@ do
         sleep 3
         continue
       fi
-    else
+    elif [ "${lightning}" = "lnd" ]; then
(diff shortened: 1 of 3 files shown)


rootzoll:
OK I tested:
- Do a "recommended Bitcoin update to 27.0" - worked OK
- Did a minimal image sd card recovery provisioned with these scripts and it recovered OK to Bitcoin 27.0.
- Then installed lnd and it installed lnd 0.17.5 OK
... recommend merging.

openoms:
was using LND 0.17.5 all right, but now LND v0.18.0 is out. Updated ok.

rootzoll:
I would not add lnd 0.18.0 as default yet (never go with a .0 ;) .. lnd 018.x is a good candidate for RaspiBlitz v1.12.0

openoms:
Agreed.

PR #4606, opened 2024-05-31
ask before reboot in UPDATE menu

openoms:
I prefer not to reboot after updating core components
So far was using the cli to update, but introducing a question before rebooting the system would make it easier to use.


geco91 (approved):
LGTM

rootzoll (approved):
LGTM

PR #4609, opened 2024-06-02
#4554 Deactivate Touchscreen Option in SSH Menu

### LINKED ISSUE #4554: Deactivate Touchscreen and remove from the ssh menu until fixed
The Rapsiblitz LCD stays black after final reboot. With ssh it's possible to check out the normal view and everything else. 
I tried to update it again but the same Result came out. 

It worked well with the old version. 

Thanks!

rootzoll:
fixes: #4554


openoms:
Should we also deactivate the option if found to be on in the raspiblitz.conf? 
People updating from old versions could get stuck with the balck screen.

Issue #4615, opened 2024-06-06
LIT not opening in browser


thecharacter:
I installed LIT, then went to open my browser, but got an error message: Safari cannot connect to the server.

**I tried to do it manually in the command line and got the error:**
lit@raspberrypi:~ $ litcli sessions add --label="default" --type=admin
[litcli] rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp [::1]:8443: connect: connection refused"

When the Lightning Terminal did come up, it rejected my password, but I know the password is correct because it works with Thunderhub. When I restart the service from the command line, I get the error below.

**Here's what I get from the command prompt:**
litd.service - litd Service
     Loaded: loaded (/etc/systemd/system/litd.service; enabled; preset: enabled)
     Active: activating (auto-restart) (Result: exit-code) since Thu 2024-06-06 04:01:10 BST; 34s ago
    Process: 827729 ExecStart=/usr/local/bin/litd (code=exited, status=1/FAILURE)
   Main PID: 827729 (code=exited, status=1/FAILURE)
        CPU: 824ms

**Here's my debug file. It appears something is going on with the TLS handshake:**

***************************************************************
* RASPIBLITZ DEBUG LOGS 
***************************************************************
blitzversion: 1.11.0
chainnetwork: bitcoin / main
 03:47:46 up  2:36,  3 users,  load average: 1.96, 1.89, 1.99

*** FAILED SERVICES ***
list any servcies with problems: sudo systemctl list-units --failed
  UNIT LOAD ACTIVE SUB DESCRIPTION
0 loaded units listed.

*** SETUPPHASE / BOOTSTRAP ***
see logs: cat /home/admin/raspiblitz.log
setupPhase--> done
state--> ready

*** BACKGROUNDSERVICE ***
to monitor Background service call: sudo journalctl -f -u background

*** BLOCKCHAIN (MAINNET) SYSTEMD STATUS ***
● bitcoind.service - Bitcoin daemon on mainnet
     Loaded: loaded (/etc/systemd/system/bitcoind.service; enabled; preset: enabled)
     Active: active (running) since Thu 2024-06-06 01:13:42 BST; 2h 34min ago
   Main PID: 6819 (bitcoind)
      Tasks: 19 (limit: 4454)
        CPU: 1h 17min 35.845s
     CGroup: /system.slice/bitcoind.service
[...]

PR #4617, opened 2024-06-10
Update LIT v0.12.5

rootzoll:
May address: #4615


rootzoll:
Tested OK on my machine.

PR #4641, opened 2024-07-05
i2p: limit addseednodes to 21 new peers, format script

openoms:


rootzoll:
Script ran thru adding i2p nodes - but got stuck for minutes on the log message:
`# Display sudo tail -n 100  | grep i2p`

Those info is basically nice to have but not critial - will remove the lines.

rootzoll:
now it runs thru and works fine - merging for v1.11.1rc1 release

rootzoll (approved):
tACK

PR #4647, opened 2024-07-13
Public-Pool implementation Issue#4581

WantClue:
This PR contributes to the Raspiblitz implementation in order to make the Public-Pool available on the Raspiblitz.
Therefore, this PR should solve Issue#4581

Additional Information:
- Due to low CPU power, it's recommended to only use this bonus.script on Pi5 with NVMe or an x86 System (it will fail and run into issues on any lower Pi variant)
- Installing this on the /home dir will work for a while but the SQL Database for the Pool will increase to about 2-3GB therefore it will fail the process if the /home dir gets to 100% use.

The PR does implicate all necessary actions in the script needed for the Raspiblitz to work.

Probably additions:

I will focus on adding a functionality that this bonus.script can only be installed on a x86 or Pi5 Variant.

Current Issue / Help wanted:

Currently this PR focus on installing Public Pool on the /mnt/hdd due to issues with user rights that I coudn't identify YET it will fail installing. If I find a solution to this (ATM) behavior, I will add this to the PR.

### DIFF
--- a/home.admin/config.scripts/bonus.public-pool.sh
+++ b/home.admin/config.scripts/bonus.public-pool.sh
@@ -1,0 +1,219 @@
+#!/bin/bash
+
+APPID="publicpool"
+VERSION="0.1"
+GITHUB_REPO="https://github.com/benjamin-wilson/public-pool.git"
+GITHUB_REPO_UI="https://github.com/benjamin-wilson/public-pool-ui.git"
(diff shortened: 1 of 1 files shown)


rootzoll:
@WantClue thanks for the PR. I am merging this into `4581-publicpool` branch on the raspiblitz repo for testing and final work.

PR #4665, opened 2024-07-28
(00raspiblitz.sh): fix typos; use -z for empty check; condensed some code

cstenglein:
tested by navigating the menu a bit

So first of all I fixed some typos, there won't be much to discuss there. :)

- for the equality checks I used `-z` instead, which should now also consider if the variable is unset
- used `elif` instead of many nested `if` statements and condensed some if checks, which condensed the `MAKE SURE BLOCKCHAIN/LN IS SYNC` check quite a bit
- used loops for multiple identical assignments, could be a style question there
- also used `&&` for single assignments, which is also a style question and thus should be discussed if wanted

### DIFF
--- a/home.admin/00raspiblitz.sh
+++ b/home.admin/00raspiblitz.sh
@@ -15,7 +15,7 @@ infoFile="/home/admin/raspiblitz.info"
 source ${infoFile}
 
 # check that basic system phase/state information is available
-if [ "${setupPhase}" == "" ] || [ "${state}" == "" ]; then
+if [ -z "${setupPhase}" ] || [ -z "${state}" ]; then
   echo "setupPhase(${setupPhase}) state(${state})"
   echo "FAIL: ${infoFile} does not exist or missing state."
   echo "Check logs & bootstrap.service for errors and report to devs."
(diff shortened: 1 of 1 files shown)


rootzoll:
Lookingt good, but lots of little changes .. would move it to v1.11.2 and then merge early for RCs so it gets some testing time.

cstenglein:
If you Like I can also Split the Changes in either:

- different commits
- different PRs

:)

rootzoll:
No need to split it .. would just to test it a bit longer before having it part of a final release.

cstenglein:
Sure, wasn't meant for 1.11.1 ;)

home.admin/00raspiblitz.sh
```diff
@@ -171,24 +152,10 @@ do
...
-        sleep 3
-        continue
-      fi  
-    else
-      if [ "${btc_default_synced}" != "1" ]; then
-        /home/admin/setup.scripts/eventBlockchainSync.sh ssh
-        sleep 3
-        continue
-      fi
+    if [ "${btc_default_synced}" != "1" ] || [ "${ln_default_ready}" == "0" ] || [ "${ln_default_sync_chain}" == "0" ] || [ "${ln_default_sync_initial_done}" == "0" ]; then
```
cstenglein, line 155: Please check if that is correct since you checked for `lightning`

rootzoll: This should would because on a system where lightning is not installed `ln_default_ready` and the other status values are just empty and would not trigger the OR clauses.

openoms (approved):
LGTM

rootzoll (approved):
tACK (basic tested on a bitcoin only & lightning lnd node)

Issue #4690, opened 2024-08-10
FRESHSETUP error using CLN v0.11.1


openoms:
Due to: https://github.com/raspiblitz/raspiblitz/issues/4170#issuecomment-2280612931

the mnemonic dependency is missing so new seedwords can't be generated.

**To Reproduce**
FRESHSETUP - CLN 

**Debug file**
```
##################
ERROR in _provision.setup.sh --> cl-wallet-new-noseeds on Sat 10 Aug 10:18:08 BST 2024\nDETAIL --> /home/admin/config.scripts/cl.hsmtool.sh new-force mainnet --> seedwords=''\nDEBUG --> /home/admin/raspiblitz.provision-setup.log
***********************************************************
ERROR - please report to development team
***********************************************************
state(error) message(cl-wallet-new-noseeds)
```

## Solution:

Exit to the terminal and delete the file:
```
sudo rm /usr/lib/python3.11/EXTERNALLY-MANAGED
```
before running the setup. 

or run with the specific user:
```
pip config set global.break-system-packages true
```

If ran into the error already also delete:
```
sudo rm /home/admin/provision.flag
rm  /home/admin/raspiblitz.info
```
and reboot to start fresh (mind that the PASSWORD A is already applied).

rootzoll:
OK merged the PR and tested. Looking good now - closing issue.

PR #4692, opened 2024-08-13
fix: pip config set global.break-system-packages true for non venv installs

openoms:
Setting 
```
pip config set global.break-system-packages true
```
for the respective users where a venv is not used.

The main problem solved with this is that cannot get the CLN plugins to respect the venvs even if they are used.

fixing: #4690
Related: #4170


rootzoll (approved):
LGTM

rootzoll:
also cherrypicked over to v1.11 branch

PR #4696, opened 2024-08-14
ci: pin packer version and update plugin for the local run

openoms:
This fixes the local packer image build process.


rootzoll (approved):
LGTM

PR #4700, opened 2024-08-15
add pip3 config set global.break-system-packages true for non-venv installs

openoms:


rootzoll:
also cehrry-picked into v1.11 branch

PR #4703, opened 2024-08-17
add shellcheck to scan bash scripts for errors

### LINKED ISSUE #2702: CONTRIBUTING tips
This is to add later on to CONTRIBUTING.md

- Use [shellcheck](https://github.com/koalaman/shellcheck) -  a static analysis tool for shell scripts -  it has a debiand packages with the same name and also can be added as a plugin to Vim, VSCode(ium) etc
- Lean from the [pure-bash-bible](https://github.com/dylanaraps/pure-bash-bible) - A collection of pure bash alternatives to external processes - avoid using heavy commands when there is already support using shell builtins
- Bash is not needed most of the cases, prefer [pure-sh-bible](https://github.com/dylanaraps/pure-sh-bible) for Posix compliancy
- Basic git commands to checkout upstream branch
- Make a script model, variables at the beggining, using case statement instead of if-then-else

cstenglein:
adds shellcheck from https://github.com/marketplace/actions/shellcheck to check the bash scripts.

Only fails on errors.

fixes #2702

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,21 @@
+name: ShellCheck
+
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
+    paths:
+      - '**/*.sh'
+
+jobs:
+  shellcheck:
+    name: Shellcheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


openoms:
this is likely a duplicate of https://github.com/raspiblitz/raspiblitz/pull/4533

cstenglein:
Argh, true, thank you!

PR #4730, opened 2024-09-06
Bonus: several `elementsd` enhancements (Liquid)

steepdawn974:
This PR does:

- Add elementsd to `blitz.shutdown.sh`  to ensure graceful shutdown and wallet release
- Sets the walletdir param in the bootstrapped `elements.conf` to the respective app-data dir. Otherwise it would default to `~/.elements/wallets` (which is likely wrong, for consistency reasons)


openoms (approved):
untested, but straightforward
utACK

rootzoll (approved):
utACK too - to be tested in v1.11.3RC1 - merging

Issue #4073, opened 2023-08-09
 Migration to VM: lcd vs. headless


Firewalker99:
Hi, 
if you want to migrate from a Raspberry Pi to VM (proxmox) and raspiblitz.conf the hdd is "lcd" the system reboots continiously till you change the raspiblitz.conf to "headlesss"
Maybe a info dialog helps to prevent this error in the future.

LOG:==> /home/admin/raspiblitz.systeminit.log <==

check that display class in raspiblitz.conf from HDD is different from as it is now in raspiblitz.info ...
infoFileDisplayClass(headless)
confFileDisplayClass(lcd)
DISPLAY RESTORE - need to update displayClass from (headless) to (lcd)'
# blitz.display.sh set-display lcd
baseimage(debian)
# old(headless)
# new(lcd)                    
err='baseimage not supported'            
No UASP FIX needed

rootzoll:
Thanks for the report .. we will optimize migration to VM in a later release - any earlier community patches welcome.

PR #4734, opened 2024-09-12
#4073 On migration from RaspberryPi to VM set display to headless

rootzoll:
Always when migration import is happening on a VM set display to "headless" - adresses #4073


rootzoll:
needs one migration testrun before merge ... from RaspberryPi with LCD to fresh VM in Proxmox

PR #4751, opened 2024-09-25
fix openchannels check or cln in menu 

openoms:


surfac3:
Tested OK

Can use the SENT in CLN menu item.

No longer getting 
image

rootzoll (approved):
utACK

Issue #4687, opened 2024-08-08
[Firewall] check open port 8332


rootzoll:
As reported on twitter:
https://x.com/BLupos/status/1821323398158672084

rootzoll:
PR #4757 ready to merge

rootzoll:
PR merged

PR #4757, opened 2024-09-26
#4687 open Bitcoin RPC 8332 port to local network by default

rootzoll:
adresses #4687


home.admin/_provision_.sh
```diff
@@ -147,6 +147,8 @@ ufw allow 18333 comment 'bitcoin testnet'
 echo "allow: bitcoin mainnet"
 ufw allow 8333 comment 'bitcoin mainnet'
 echo 'allow: lightning testnet'
+ufw allow 8332 comment 'bitcoin mainnet RPC'
+echo 'allow: lightning testnet'
```
cstenglein, line 151: Looks like a copy paste error since it now appears 2 times

rootzoll: korrekt - passe ich an

PR #4764, opened 2024-09-29
#2624 Update: Telegraf Metrics for InfluxDB/Grafana

rootzoll:
adresses #2624


rootzoll:
Testing:
- [x] See if Telegraf stays stable after update/recover

PR #4767, opened 2024-09-30
fix (bonus.nodejs.sh): fix install to get current version of nodejs

### LINKED ISSUE #4587: Update Node.js
```
node -v
v20.5.1
```
Latest LTS is:
v20.13.1

All Web Apps need to be tested.

cstenglein:
Tested on my node.

- Changed the inconsistent spelling to `Node.js`.
- Uses the `nodistro` distro and removed the `deb-src` since it is not available anymore.

For reference see the install script at https://deb.nodesource.com/setup_20.x.

This then installs the newest 20.x version of nodejs (current 20.17.0)

closes #4587


rootzoll (approved):
ACK - I tested install and got v20.17.0 instead of v20.5.1 which we shipped v1.11.2 with. Based on my research this contains no kreaking changes.

PR #4770, opened 2024-10-01
#4566 Fixing Locale

### LINKED ISSUE #4566: Make sure VMs run on lang=en_US.UTF-8
I used a german debian12 to run on my mac UTM/qemu and ran the install script on. Some scripts have problems with the german lang - it should make sure (tested or set) that it is english during install script.

`cat /etc/default/locale 2>/dev/null | grep -c "LANG=en_US.UTF-8"`

rootzoll:
fixes #4566


rootzoll:
tested sd card build

Issue #4607, opened 2024-06-01
[feature-request] Add Tailscale to config.scripts


NilByte:
Tailscale seems quite easy to install following this guide:

https://tailscale.com/download/linux/rpi-bullseye

To get scanning of QR code for Bitbanana connect to work, I had to add line in lnd.conf:

`tlsextradomain=<my tailscale domain>`

Latest Bitbanana app 0.8.1 has added Tailscale as VPN option and it works well. No slow response time in app as when it's connected over Tor.

rootzoll:
PR to add tailscale: #4712

PR #4787, opened 2024-10-05
#4607 Add Tailscale to SSH menus

rootzoll:
adresses #4607


rootzoll:
tested to be working also after a recover/update - merging

Issue #4821, opened 2024-10-27
[VM] Shared Folder Code Sync


geco91:
Make it possible when running RaspiBlitz in a VM (on local dev laptop) to sync code inside VM with shared folder - so there is no need to go thru Github.

geco91:
recommend moving to v1.12.x

Issue #4822, opened 2024-10-27
[VM] Install failed-> "fatal: fetch-pack:  invalid index-pack output"


mr-q17:
Install Raspiblitz via Proxmox / Debian 12.7.0 amd64

- Fresh Debian install, VM via Proxmox 8.2-2

- First install with:
`wget https://raw.githubusercontent.com/rootzoll/raspiblitz/dev/build_sdcard.sh`

- 2nd try:
`wget https://raw.githubusercontent.com/rootzoll/raspiblitz/v1.11/build_sdcard.sh`
 
- Both ends with:
`fatal: fetch-pack: invalid index-pack output`

<img width="1049" alt="Bildschirmfoto 2024-10-27 um 23 09 48" src="https://github.com/user-attachments/assets/dfd8625e-b7b2-416e-8a5d-119e68e5bbc4">

adevoss:
Just a guess: Firewall / VPN issue?
I had issues before with VPN and github, not only with raspiblitz. Try installing Raspiblitz without VPN.

geco91:
Its an issue that happens sometimes and sometimes not - its related to github server response (and internet stability).

geco91:
#4823 will not solve but try to reduce this happening

PR #4823, opened 2024-10-27
#4821 Sync a Sharedfolder for local VM development

geco91:
addresses #4821


geco91:
also addresses #4822 by reducing the depth of the github repo to be cloned on build time and so minizimizing the risk of the cloning failing.

PR #4844, opened 2024-11-13
specter update to 2.0.5

openoms:


rootzoll (changes requested):
Install script runs thru.

@openoms please add a CHANGES entry and then this PR is good to go.

PR #4853, opened 2024-11-29
cln: set systemd service to Restart=always

openoms:
prevents cln to remain off after dependency failure (when bitcoind is syncing)


rootzoll (approved):
utACK

PR #4854, opened 2024-11-30
arm64 base image update to 2024-11-19-raspios-bookworm-arm64

openoms:


rootzoll:
I think its not related to the new base image, with this update we should also take care of the following error/warnings during build on RaspberryPi:

```
# Adding debian to /etc/hosts
Failed to restart networking.service: Unit networking.service not found.
```

But other than this .. sd card build and fatpack ran thru without erros on the the new base OS image.

rootzoll (approved):
tACK

Issue #4903, opened 2024-12-23
Public Pool. Error getmininginfo 503 - "Work queue depth exceeded".


NilByte:
From debug log:

```
*** LAST 20 PUBLIPOOL LOGS ***
sudo journalctl -u publicpool -b --no-pager -n20
Dec 22 18:34:44 raspberrypi npm[379721]: Error getmininginfo 503 - "Work queue depth exceeded"
Dec 22 18:34:45 raspberrypi npm[379721]: Error getblocktemplate: Error: ESOCKETTIMEDOUT
Dec 22 18:34:45 raspberrypi npm[379721]: /home/publicpool/publicpool/src/services/bitcoin-rpc.service.ts:131
Dec 22 18:34:45 raspberrypi npm[379721]:             throw new Error('Error getblocktemplate');
Dec 22 18:34:45 raspberrypi npm[379721]:                   ^
Dec 22 18:34:45 raspberrypi npm[379721]: Error: Error getblocktemplate
Dec 22 18:34:45 raspberrypi npm[379721]:     at BitcoinRpcService.getBlockTemplate (/home/publicpool/publicpool/src/services/bitcoin-rpc.service.ts:131:19)
Dec 22 18:34:45 raspberrypi npm[379721]:     at processTicksAndRejections (node:internal/process/task_queues:95:5)
Dec 22 18:34:45 raspberrypi npm[377558]: npm notice
Dec 22 18:34:45 raspberrypi npm[377558]: npm notice New major version of npm available! 10.8.2 -> 11.0.0
Dec 22 18:34:45 raspberrypi npm[377558]: npm notice Changelog: https://github.com/npm/cli/releases/tag/v11.0.0
Dec 22 18:34:45 raspberrypi npm[377558]: npm notice To update run: npm install -g npm@11.0.0
Dec 22 18:34:45 raspberrypi npm[377558]: npm notice
Dec 22 18:34:45 raspberrypi systemd[1]: publicpool.service: Main process exited, code=exited, status=1/FAILURE
Dec 22 18:34:45 raspberrypi systemd[1]: publicpool.service: Failed with result 'exit-code'.
Dec 22 18:34:45 raspberrypi systemd[1]: publicpool.service: Consumed 18min 7.285s CPU time.
Dec 22 18:35:15 raspberrypi systemd[1]: publicpool.service: Scheduled restart job, restart counter is at 30.
[...]

PR #4915, opened 2025-01-04
#4903 increase rpcworkqueue

rootzoll:
addresses #4903


home.admin/config.scripts/blitz.i2pd.sh
openoms, line 1: are the changes in this file needed?

PR #4928, opened 2025-01-18
amd64 base image update to debian-12.9.0-amd64-netinst.iso

openoms:


openoms:
merging after the successful builds to unblock the amd64 image builds

PR #4929, opened 2025-01-18
check for physical rpi in bootstrap, only run uasp-fix on rpi

openoms:
using a similar pattern to checking for a VM

attempts to fix the underlying problem in: https://github.com/raspiblitz/raspiblitz/issues/4927


rootzoll:
In the rewrite for v1.12.0 storage concept this is already fixed - see branch `4324-bootdrive`:

https://github.com/raspiblitz/raspiblitz/blob/3e342f0801884e310f866c5edbf265f72ee4163d/home.admin/_bootstrap.sh#L421

PR #4942, opened 2025-01-25
Switch to bitcoin user for version verification

frennkie:
We normally run bitcoind as user bitcoin.
Running it here as **root** could potentially at some point cause issues (e.g. if files are created and the ownership is then off).

What do you think @openoms ?

- [X] That its based against the `dev` branch - not the default release branch.


openoms (approved):

rootzoll (approved):

rootzoll:
tACK - @frennkie thanks will merge

PR #4962, opened 2025-02-23
nodejs update to 22

openoms:
required by some newer apps I experiment with


cstenglein:
WebUI already works with both 20.x and 22., so no issue there.

rootzoll (approved):
lest test during release candidate

PR #4968, opened 2025-03-10
add knots option

Retropex:
- [x] That its based against the `dev` branch - not the default release branch.

This is not ready to merge, its to start receiving advice on what should I add/remove.

Close #4678

PR #4969, opened 2025-03-13
add datum

Retropex:

Add bonus app datum

### DIFF
--- a/home.admin/config.scripts/bonus.datum.sh
+++ b/home.admin/config.scripts/bonus.datum.sh
@@ -1,0 +1,411 @@
+#!/bin/bash
+
+APPID="datum"
+
+# clean human readable version - will be displayed in UI
+# just numbers only separated by dots (2 or 0.1 or 1.3.4 or 3.4.5.2)
+VERSION="0.4.1"
+
+# the git repo to get the source code from for install
+GITHUB_REPO="https://github.com/OCEAN-xyz/datum_gateway.git"
+
+# the github tag of the version of the source code to install
+# can also be a commit hash
+# if empty it will use the latest source version
+GITHUB_TAG="v0.4.1beta"
+
+# the github signature to verify the author
+# leave GITHUB_SIGN_AUTHOR empty to skip verifying
+GITHUB_SIGN_AUTHOR="luke-jr"
+GITHUB_SIGN_PUBKEYLINK="https://keyserver.ubuntu.com/pks/lookup?op=get&search=0x1a3e761f19d2cc7785c5502ea291a2c45d0c504a"
+GITHUB_SIGN_FINGERPRINT="A291A2C45D0C504A"
+
+# port numbers the app should run on
+# delete if not an web app
+PORT_CLEAR="21000"
+
+# BASIC COMMANDLINE OPTIONS
+# you can add more actions or parameters if needed - for example see the bonus.rtl.sh
+# to see how you can deal with an app that installs multiple instances depending on
+# lightning implementation or testnets - but this should be OK for a start:
+if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
+  echo "# bonus.${APPID}.sh status    -> status information (key=value)"
+  echo "# bonus.${APPID}.sh on        -> install the app"
+  echo "# bonus.${APPID}.sh off       -> uninstall the app"
+  echo "# bonus.${APPID}.sh menu      -> SSH menu dialog"
+  exit 1
+fi
+
+# echoing comments is useful for logs - but start output with # when not a key=value
+echo "# Running: 'bonus.${APPID}.sh $*'"
+
+# check & load raspiblitz config
+source /mnt/hdd/raspiblitz.conf
+
+#########################
+# INFO
+#########################
+
+# this section is always executed to gather status information that
+# all the following commands can use & execute on
+
+# check if app is already installed
+isInstalled=$(sudo ls /etc/systemd/system/${APPID}.service 2>/dev/null | grep -c "${APPID}.service")
+
+# check if service is running
+isRunning=$(systemctl status ${APPID} 2>/dev/null | grep -c 'active (running)')
+
+if [ "${isInstalled}" == "1" ]; then
+
+  localIP=$(hostname -I | awk '{print $1}')
+
+fi
+
+# if the action parameter `status` was called - just stop here and output all
+# status information as a key=value list
+if [ "$1" = "status" ]; then
+  echo "appID='${APPID}'"
+  echo "version='${VERSION}'"
+  echo "githubRepo='${GITHUB_REPO}'"
+  echo "githubVersion='${GITHUB_TAG}'"
+  echo "githubSignature='${GITHUB_SIGNATURE}'"
+  echo "isInstalled=${isInstalled}"
+  echo "isRunning=${isRunning}"
+  if [ "${isInstalled}" == "1" ]; then
+    echo "portCLEAR=${PORT_CLEAR}"
+    echo "localIP='${localIP}'"
+  fi
+  exit
+fi
+
+##########################
+# MENU
+#########################
+
+# The `menu` action should give at least a SSH info dialog - when an webapp show
+# URL to call (http & https+fingerprint) otherwise some instruction how to start it.
+
+# This SSH dialog will be later called by the MAIN MENU to be available to the user
+# when app is installed.
+
+# This menu can also have some more complex structure if you want to make it easy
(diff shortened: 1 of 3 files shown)


Retropex:
updated to datum `v0.3.1beta`

Retropex:
Need #4968 first to be merged.

home.admin/config.scripts/bonus.datum.sh
```diff
@@ -0,0 +1,418 @@
...
+  # and copy the skeleton files
+  sudo adduser --system --group --shell /bin/bash --home /home/${APPID} ${APPID} || exit 1
+  # copy the skeleton files for login
+  sudo -u ${APPID} cp -r /etc/skel/. /home/${APPID}/
+
+  # add user to special groups with special access rights
+  # BACKGROUND there are some unix groups available that will give the access to
+  # like for example to the lnd admin macaroons - to check all groups available use:
+  # `cut -d: -f1 /etc/group | sort` command on raspiblitz commandline
+  echo "# add use to special groups"
```
steepdawn974: It is recommended to run `datum` under the same user as bitcoind, i.e. `bitcoin`
https://github.com/OCEAN-xyz/datum_gateway

>  It is suggested you run the DATUM Gateway as the same user as your full node and utilize the following configuration line in your bitcoin.conf

I don't see a good reason for adding the user to the `lndadmin` group...

home.admin/config.scripts/bonus.datum.sh
```diff
@@ -0,0 +1,418 @@
...
+  echo "# mark app as uninstalled in raspiblitz config"
+  /home/admin/config.scripts/blitz.conf.sh set ${APPID} "off"
+
+  echo "# deleting useless bitcoin.conf entry"
+  sudo -u bitcoin sed -i "/^blockmaxsize=/d" "/mnt/hdd/bitcoin/bitcoin.conf"
+  sudo -u bitcoin sed -i "/^blockmaxweight=/d" "/mnt/hdd/bitcoin/bitcoin.conf"
+  sudo -u bitcoin sed -i "/^maxmempool=/d" "/mnt/hdd/bitcoin/bitcoin.conf"
+  sudo -u bitcoin sed -i "/^blocknotify=/d" "/mnt/hdd/bitcoin/bitcoin.conf"
+  sudo -u bitcoin sed -i "s/^dnsseed=.*/dnsseed=0/" "/mnt/hdd/bitcoin/bitcoin.conf"
+  sudo -u bitcoin bash -c 'echo "onlynet=onion" >> /mnt/hdd/bitcoin/bitcoin.conf'
```
steepdawn974: I would add another comment line here, since the following commands are not about "deleting useless bitcoin.conf" entries

home.admin/config.scripts/bonus.datum.sh
```diff
@@ -0,0 +1,418 @@
...
+  # dont run install if already installed
+  if [ ${isInstalled} -eq 1 ]; then
+    echo "# ${APPID}.service is already installed."
+    exit 1
+  fi
+
+  echo "# Installing ${APPID} ..."
+
+  # check and install NodeJS - if already installed it will skip
+  /home/admin/config.scripts/bonus.nodejs.sh on
```
steepdawn974: Why do we need NodeJS installed for `datum`?

If this is for the Raspi webui, then I reckon you first should check if this is a headless node, and only if not --> install Node.

home.admin/config.scripts/bonus.datum.sh
```diff
@@ -0,0 +1,418 @@
...
+  }
+}
+" > /mnt/hdd/app-data/datum/datum_config.json' _ "$PASS"
+
+  # Configure bitcoind
+
+  sudo -u bitcoin bash -c 'echo "blockmaxsize=3985000" >> /mnt/hdd/bitcoin/bitcoin.conf'
+  sudo -u bitcoin bash -c 'echo "blockmaxweight=3985000" >> /mnt/hdd/bitcoin/bitcoin.conf'
+  sudo -u bitcoin bash -c 'echo "maxmempool=1000" >> /mnt/hdd/bitcoin/bitcoin.conf'
+  sudo -u bitcoin bash -c 'echo "blocknotify=curl -s -m 5 127.0.0.1:21000/NOTIFY" >> /mnt/hdd/bitcoin/bitcoin.conf'
```
steepdawn974: If `bitcoind` and `datum` are on the same machine (as we have on the raspiblitz) then the docs suggest to use:
`blocknotify=killall -USR1 datum_gateway`

https://github.com/OCEAN-xyz/datum_gateway

home.admin/config.scripts/bonus.datum.sh
```diff
@@ -0,0 +1,418 @@
...
+    \"pooled_mining_only\": true
+  }
+}
+" > /mnt/hdd/app-data/datum/datum_config.json' _ "$PASS"
+
+  # Configure bitcoind
+
+  sudo -u bitcoin bash -c 'echo "blockmaxsize=3985000" >> /mnt/hdd/bitcoin/bitcoin.conf'
+  sudo -u bitcoin bash -c 'echo "blockmaxweight=3985000" >> /mnt/hdd/bitcoin/bitcoin.conf'
+  sudo -u bitcoin bash -c 'echo "maxmempool=1000" >> /mnt/hdd/bitcoin/bitcoin.conf'
```
steepdawn974: There is another recommended option: `blockreconstructionextratxn=1000000` . Any good reason to not include it here?
https://github.com/OCEAN-xyz/datum_gateway#node-configuration

steepdawn974 (commented):
Thank you for this, @Retropex ! Saves me a lot of work, as I was just about to implement the same thing ;) 

I added a few inline comments

steepdawn974:
> Need #4968 first to be merged.

Knots is not a strict requirement for DATUM; it can be run against the default bitcoind as well.
This PR (once approved) can be merged without #4968 imho.

home.admin/config.scripts/bonus.datum.sh
Retropex: This is a leftover from template script.

home.admin/config.scripts/bonus.datum.sh
Retropex: For better security I prefer keeping users separated, since datum is connected with RPC URL it's not a problem if they are not under the same user.

`lndadmin` was a leftover too.

home.admin/config.scripts/bonus.datum.sh
Retropex: Forgot that one, thanks.

home.admin/config.scripts/bonus.datum.sh
Retropex: This would require a permission elevation of `bitcoin`.

Retropex:
@steepdawn974 Thanks for the review!

While it's true that DATUM can work with Core it's pretty pointless since Core have almost no option to customize your block template.

steepdawn974:
> @steepdawn974 Thanks for the review!
> 
> While it's true that DATUM can work with Core it's pretty pointless since Core have almost no option to customize your block template.

@Retropex Yes, but my point was just that there is no *technical* blocker/dependency for this PR being merged. I was just afraid that we needlessly hold up this useful feature for another PR that might still take some time. But since it seems https://github.com/raspiblitz/raspiblitz/pull/4968 is well on its way to be merged, it's fine imho to release those two together

rootzoll:
Due to time contrains pushing to v1.12.2 - lets see first if all works out with the Knots integration in v1.12.0

steepdawn974:
Tested this on a RaspiBlitz and ran into a bug that breaks bitcoind on install.

The script writes to `/mnt/hdd/bitcoin/bitcoin.conf`, but on RaspiBlitz `/mnt/hdd/bitcoin` is a symlink to the **data directory** (`/mnt/hdd/app-storage/bitcoin`), not the config directory. The actual config lives at `/mnt/hdd/app-data/bitcoin/bitcoin.conf`.

This causes two problems:

1. **bitcoind won't start** — a stray `bitcoin.conf` gets created in the datadir, which conflicts with the `-conf=` argument. bitcoind refuses to run with: `Data directory contains a "bitcoin.conf" file which is ignored, because a different configuration file from command line argument -conf= is being used instead.`
2. **datum can't connect to bitcoind** 

The fix is to replace all `/mnt/hdd/bitcoin/bitcoin.conf` references with `/mnt/hdd/app-data/bitcoin/bitcoin.conf` throughout the script (install, prestart, and uninstall sections).

After patching the paths manually, both bitcoind and datum start up fine.

Retropex:
Sorry for the delay, I have now added your fix!

PR #5001, opened 2025-05-15
#4916 enforce ssh certs permissions

### LINKED ISSUE #4916: SSH connection refused
**Describe the bug**
A clear and concise description about the issue you encounter.

I can no longer ssh into raspiblitz on raspberry pi 5. I always get a connection refused message.
**To Reproduce**
Steps to reproduce the behavior.
IMG_4720

I was having issues with some of the applications not working and used ChatGPT to try and address them (checking channels using raspiblitz menu in command didn’t work, loop not working in RTL, among other things). Chatgpt somewhat buggered it up more so I decided I would start over. Backed up my LND and reflashed the SD card. After booting I was able to ssh in once, and then after that I get the connection refused message. 
-I have attempted adding an ssh file with no extension to the boot partition.
-I have turned off firewall and tried again with no luck.
-I have tried booting with an external monitor plugged in but I only get system status messages (pictures
 is attached) and the keyboard doesn’t do anything (using a Bluetooth usb for keyboard) so I can’t do anything with a command line.
-I was able to access the Web UI but now I’m stuck on waitsync with the note of provision done. I wasn’t able to backup my LND again so the backup is probably a day or so old.

I’m fine with starting over except that I have funds on the lightning network that I don’t want to lose. I went into the SSD and copied off the app-data folder and the lnd folder but I’m afraid to wipe the SSD and SD to start over in risk of losing the funds.

**Debug file**
If possible, please add a debug report. To do that, drag a text file with the debug report in here.

None.

rootzoll:
closes #4916

### DIFF
--- a/home.admin/config.scripts/blitz.ssh.sh
+++ b/home.admin/config.scripts/blitz.ssh.sh
@@ -134,6 +134,10 @@ if [ "$1" = "checkrepair" ]; then
     /home/admin/config.scripts/blitz.ssh.sh renew
   fi
 
+  # make sure permissions are correct
+  sudo chown root:root /etc/ssh/ssh_host_*
+  sudo chmod 600 /etc/ssh/ssh_host_*
+
   # check if SSHD service is NOT running & active
   sshdRunning=$(sudo systemctl status sshd | grep -c "active (running)")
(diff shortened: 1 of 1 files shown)


rootzoll:
OK test looks good. Merging.

PR #5016, opened 2025-05-18
btc-rpc-explorer: get working with Node22

rootzoll:


rootzoll:
OK on the latest comits in the GitHub Repo it seems to work .. so we switching from version (latest is over 2 years old) to github commit hashes here for the install script.

Issue #4648, opened 2024-07-17
[sdcard] Check Tor keys & update


rootzoll:
Tor Key expired, need to add the new one
They also have an APT repo for the Keys, so we dont need to add them manually
And Tor 0.4.8 has the PoW protection, which we could enable

cstenglein:
More info:

When installing tor currently, the installation fails:

```
W: Failed to fetch tor+http://apow7mjfryruh65chtdydfmqfpj5btws7nbocgtaovhvezgccyjazpqd.onion/torproject.org/dists/bookworm/InRelease  The following signatures were invalid: EXPKEYSIG 74A941BA219EC810 deb.torproject.org archive signing key
```


We could use the package `apt install apt-transport-tor` for the keys:

https://support.torproject.org/apt/apt-over-tor/

Alternative:

```
wget -qO- https://deb.torproject.org/torproject.org/A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89.asc | gpg --dearmor | tee /usr/share/keyrings/deb.torproject.org-keyring.gpg >/dev/null
```

See: https://support.torproject.org/apt/tor-deb-repo/

rootzoll:
So it the moment we have tor installed on the base image and we should be good for v1.11.1 & 1.11.2:

```
admin@192.168.178.60:~ ₿ tor --version
Tor version 0.4.7.16.
Tor is running on Linux with Libevent 2.1.12-stable, OpenSSL 3.0.13, Zlib 1.2.13, Liblzma 5.4.1, Libzstd 1.5.4 and Glibc 2.36 as libc.
Tor compiled with GCC version 12.2.0
```

But yes for the next build of the base image with v1.12.0 we should adapt the install. Moving to v1.12.0 milestone

openoms:
In the meantime:

to refresh the Tor key use:

menu - UPDATE - TOR

or run:
```
wget -qO- https://deb.torproject.org/torproject.org/A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89.asc | gpg --dearmor | sudo tee /etc/apt/trusted.gpg.d/torproject.gpg >/dev/null
```
https://t.me/RaspiBlitzFAQ/800

rootzoll:
For v1.11.2 & v1.11.3 this will be updates with PR #4697 adding the new key by release script.

Moving issue to v1.12.0 with TODOS:
- [ ] Check if new base image is still oudated signing key
[...]

PR #5018, opened 2025-05-18
4648 improve check tor valid signature

rootzoll:
before key was downloaded with just 32 bytes and was invalid ... now adding deeper validation


rootzoll:
#4648 improved validation merged

PR #5024, opened 2025-05-19
5015 Repair Delete all Data But Blockchain

### LINKED ISSUE #5015: Repair Delete everything but Blockchain
It looks like after the new drive layout the repair option to delete all but keep blockchain needs update.

rootzoll:
closes #5015


rootzoll:
test again with a node with at leat 1GB blockchain and tehn check if setup-fresh & keep-blockchain works .. if not then deac again for new drive layout.

PR #5031, opened 2025-05-25
amd64 base image update to debian 12.11.0

openoms:
Debian 12.11.0 is out:
https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/


rootzoll (approved):
utACK

PR #5062, opened 2025-06-15
cl.check.sh: fix substitutions

openoms:
tested with mainnet and signet nodes

(the current on dev is adding duplicate port numbers and host entry)


openoms:
fixes https://github.com/raspiblitz/raspiblitz/issues/5061

rootzoll (approved):
utACK

Issue #4967, opened 2025-03-07
Update to CLN v25.02 - Wallet won't unlock.


tobomobo:
**Describe the bug**
(Recklessly) updated to newest CLN Version [v25.02](https://github.com/ElementsProject/lightning/releases/tag/v25.02), but now the wallet won’t unlock anymore. System reboots didn't help. 

I was running [v24.11.1](https://github.com/ElementsProject/lightning/releases/tag/v24.11.1) without issues before the upgrade.

**Debug file**

***************************************************************
* RASPIBLITZ DEBUG LOGS 
***************************************************************
blitzversion: 1.11.4
commit-release: dev
commit-active: 53f6603e
chainnetwork: bitcoin / main
 09:58:50 up 5 min,  1 user,  load average: 0.72, 0.86, 0.45

*** FAILED SERVICES ***
list any servcies with problems: sudo systemctl list-units --failed
  UNIT LOAD ACTIVE SUB DESCRIPTION
0 loaded units listed.

*** SETUPPHASE / BOOTSTRAP ***
see logs: cat /home/admin/raspiblitz.log
setupPhase--> done
state--> ready

*** BACKGROUNDSERVICE ***
to monitor Background service call: sudo journalctl -f -u background

*** BLOCKCHAIN (MAINNET) SYSTEMD STATUS ***
● bitcoind.service - Bitcoin daemon on mainnet
     Loaded: loaded (/etc/systemd/system/bitcoind.service; enabled; preset: enabled)
     Active: active (running) since Fri 2025-03-07 09:53:54 CET; 4min 55s ago
    Process: 8937 ExecStartPre=/home/admin/config.scripts/bitcoin.check.sh prestart mainnet (code=exited, status=0/SUCCESS)
    Process: 8951 ExecStart=/usr/local/bin/bitcoind -daemonwait -conf=/mnt/hdd/bitcoin/bitcoin.conf -datadir=/mnt/hdd/bitcoin (code=exited, status=0/SUCCESS)
   Main PID: 8956 (bitcoind)
      Tasks: 147 (limit: 9473)
     Memory: 3.3G
        CPU: 56.263s
     CGroup: /system.slice/bitcoind.service
             └─8956 /usr/local/bin/bitcoind -daemonwait -conf=/mnt/hdd/bitcoin/bitcoin.conf -datadir=/mnt/hdd/bitcoin

[...]

PR #5068, opened 2025-07-05
#4967 Update Core Lightning v25.5

rootzoll:
@openoms it installs ok any objections to add thsi for v1.12?


openoms:
Tested update on amd64 and rpi, does work without changes in the install script

PR #5149, opened 2025-10-14
update knots to v29.2

Retropex:

Replace #5115


rootzoll (approved):
utACK

PR #5151, opened 2025-10-15
Add bitcoin-to-knots.sh script for switch Bitcoin Core to Bitcoin KNOTS.

SatoshiNakamotoBitcoin:
This script updates Bitcoin KNOTS from Bitcoin Core, providing options for info, tested, reckless, or custom updates. It includes version checks, downloads, and installations based on user input.

Based on the gist by @BlueeeMoon https://x.com/moon33_blue/status/1963060753088315454

Launch via:
ssh raspiblitz
cd config.scripts
./bitcoin-to-knots.sh custom 29.1.knots20250903 skipverify
 
If you have a old version of Knots, you can run the same script without this part of the script to skip the switching proccess:

`
isBitcoinCoreInstalled=$(sudo -u bitcoin bitcoind --version | head -n 1 | grep -c "Bitcoin Core")
if [ ${isBitcoinCoreInstalled} -eq 0 ]; then
  echo "# if you have Bitcoin KNOTS installed - you first need to switch back to Bitcoin Core"
  echo "# error='Bitcoin Core is not installed'"
  sleep 3
  exit 1
fi
`

@rootzoll

### DIFF
--- a/home.admin/config.scripts/bitcoin-to-knots.sh
+++ b/home.admin/config.scripts/bitcoin-to-knots.sh
@@ -1,0 +1,307 @@
+#!/bin/bash
+
+# command info
+if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
+  echo "Interim optional Bitcoin KNOTS updates between RaspiBlitz releases."
+  echo "bitcoin.update.sh [info|tested|reckless|custom]"
+  echo "info -> get actual state and possible actions"
+  echo "tested -> only do a tested update by the RaspiBlitz team"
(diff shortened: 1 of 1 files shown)


rootzoll:
This is already a script to switch to knots (included since last release v1.12.0) ... why is a second one needed?
https://github.com/raspiblitz/raspiblitz/blob/dev/home.admin/config.scripts/bonus.knots.sh

SatoshiNakamotoBitcoin:
If there's already an option, then it's not necessary.

Reading about users creating their own scripts, I understood that option didn't exist.

I'm glad to know they can switch easily to Knots.

PR #5195, opened 2025-11-29
specter update to 2.1.1 and config fixes

openoms:
- fixes some json parsing problems

- the passwordaspin function is broken, because the directory where Specter looks fro the conf is changed:
https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/config.py#L90-L93
so setting the hash of the password_B

- [x] need to deal with the case when PASSSWORD_B is changed


rootzoll (approved):
ACK

PR #5235, opened 2026-02-06
Fix LightningTipBot source

ChuckNorrison:
The git url is not available anymore. I made a fork of the working repository when this setup was written.


rootzoll (approved):
Good to see this kept alive

PR #5240, opened 2026-02-26
Add JoinMarket-NG integration

m0wer:
Adds Raspiblitz integration for JoinMarket-NG.

### Summary

- add install, uninstall, update, and menu integration for JoinMarket-NG
- run JoinMarket-NG under a dedicated `joinmarketng` user with data stored under `/mnt/hdd/app-data/joinmarket-ng`
- add a maker systemd service, password handling, and narrow sudoers rules for the menu-driven flows
- use the bundled upstream `jm-ng` TUI and fix issues found during testing around formatting, password prompts, update flow, and status reporting
- add bats coverage for the bonus script and command wiring
- fix related hardcoded `bitcoind.service` references so signet/testnet setups behave correctly

### Testing

- `bats test/bonus.joinmarket-ng.bats`
- `bats test/commands.joinmarket-ng.bats`
- manual testing on signet for wallet, taker, maker, bonds, logs, and update flows

Reference for signet setup: https://primal.net/a/naddr1qvzqqqr4gupzqaze6ve67esxdurxe7rhjmnfpke6jml52d8eah6w4d6d7teqw2ymqqn8yctnwp5kymrfw3az6er9wejkcmmsd4jkuapdwekj6um9w36hqttnd9nkuet5ukt7zt

### Notes

- data is preserved across reinstall and uninstall
- maker does not auto-start; wallet setup is still explicit and user-driven


home.admin/config.scripts/blitz.passwords.sh
```diff
@@ -376,6 +376,14 @@ elif [ "${abcd}" = "b" ]; then
     sudo systemctl restart joinmarket-api.service
   fi
 
+  # JoinMarket-NG
+  if [ "${joinmarketNG}" == "on" ]; then
+    echo "# changing the RPC password for JOINMARKET-NG"
+    sudo sed -i "s/^rpc_password = .*/rpc_password = \"${newPassword}\"/" /home/joinmarketng/.joinmarket-ng/config.toml
+    echo "# changing the password for the 'joinmarketng' user"
+    echo "joinmarketng:${newPassword}" | sudo chpasswd
```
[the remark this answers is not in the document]
m0wer: This is how it's done for `bonus.joinmarket.sh`. Open to suggestions if there's a better way.

home.admin/config.scripts/menu.joinmarket-ng.sh
```diff
@@ -0,0 +1,457 @@
...
+              clear
+              if [ -z "$CURRENT_WALLET" ]; then
+                  echo "ERROR: No wallet configured. Set up a wallet first (W -> SEL or NEW)."
+              else
+                  echo "Starting JoinMarket-NG maker..."
+                  echo ""
+                  read -s -p "Wallet encryption password (Enter to skip if unencrypted): " WALLET_PWD
+                  echo ""
+                  # Write password to env file (mode 600, owned by joinmarketng)
+                  echo "MNEMONIC_PASSWORD=${WALLET_PWD}" > "$MAKER_ENV"
```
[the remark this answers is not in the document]
m0wer: Already fixed.

openoms:
Thanks for the PR, will use it as a base for testing.

m0wer:
I have been testing on signet and all looks good: wallet (balance, send, history, ...), taker, maker (FBs, start, stop, logs, ...).

I've made some improvements and fixes. Also added some tests based on the ones you had for PostgreSQL in a separate commit. They should run fine in the GitHub workflow, but can also be tested locally:

```shell
$ docker run --rm -it \
  -v "$(pwd):/repo:ro" \
  -w /repo/test \
  ubuntu:24.04 \
  bash -c '
    export DEBIAN_FRONTEND=noninteractive
    apt-get update -qq && apt-get install -y -qq bats > /dev/null 2>&1
    bats --formatter pretty ./bonus.joinmarket-ng.bats
  '
bonus.joinmarket-ng.bats
 ✓ status before install shows isInstalled=0
 ✓ install joinmarket-ng (on)
 ✓ status after install shows isInstalled=1 and correct version
 ✓ jm-wallet --help works
 ✓ jm-wallet subcommands respond to --help
 ✓ jm-maker --help works
 ✓ jm-maker subcommands respond to --help
 ✓ jm-taker --help works
 ✓ jm-taker subcommands respond to --help
 ✓ store-password sets mnemonic_password in config.toml
 ✓ wipe-password removes mnemonic_password from config.toml
 ✓ prestart injects RPC credentials into config.toml
 ✓ uninstall joinmarket-ng (off)
 ✓ status after uninstall shows isInstalled=0
 ✓ reinstall preserves existing config.toml
 ✓ raspiblitz.conf has joinmarketNG=on after install
 ✓ unknown parameter exits with failure
 ✓ final cleanup uninstall

18 tests, 0 failures
```

m0wer:
Hi! Improved the menu for prompting parameters and a few other things.

Anything I can do to help you review this?

openoms:
Installs all right, there are some formatting issues in the menu, should be easy to fix:
image

image


Saving the mnemonic password did not work 
 - there should not be a sudo password needed here
 - the password was not saved in the config despite the output
image

m0wer:
Fixed the mentioned issues and did some refactoring. Moved out the TUI logic to jm-ng repo since it's also useful for non raspiblitz users.

Ready for testing!

m0wer:
@openoms hare are some questions about your expectations/preferences:

1. Is the expected pattern on Raspiblitz that wallet apps run under a dedicated Linux user, and the admin user accesses them via `sudo -u <appuser> ...`?
2. Is `0700` on the wallet directory, owned by the app user, the preferred default for apps that manage wallet material?
3. For menu actions triggered from the app user, is adding narrow `NOPASSWD` sudoers entries the expected approach when root-owned helpers need to run?

openoms:
>     1. Is the expected pattern on Raspiblitz that wallet apps run under a dedicated Linux user, and the admin user accesses them via `sudo -u <appuser> ...`?
> 

Lightning and bitcoin-cli applications work this way so it is acceptable.
Joininbox does switch to it's own user, but I see no strong advantage of working that way.

>     2. Is `0700` on the wallet directory, owned by the app user, the preferred default for apps that manage wallet material?

Yes, this should provide that minimal user separation, so at last non sudo capable users can't access the wallet files.

>     3. For menu actions triggered from the app user, is adding narrow `NOPASSWD` sudoers entries the expected approach when root-owned helpers need to run?

This is how joininbox works too, would be nice to avoid, but ok if necessary.

openoms:
Really happy to see the bats test added, but it also needs to pass to be able to merge @m0wer

I do continue with some manual test on signet.

m0wer:
bats fixed!

openoms:
Install from this PR succeeded and update to jm-ng 0.37.1 without problems.

home.admin/config.scripts/bonus.joinmarket-ng.sh
```diff
@@ -0,0 +1,977 @@
...
+  elif [ -f "${ENV_FILE}" ] && grep -q '^MNEMONIC_PASSWORD=' "${ENV_FILE}"; then
+    echo "# Wallet password already staged in .maker.env, no prompt needed."
+  else
+    read -r -s -p "Wallet encryption password (Enter to skip if unencrypted): " WALLET_PWD
+    echo ""
+    # Write the password for the systemd EnvironmentFile using the
+    # double-quoted form so special characters survive. systemd interprets
+    # C-style escapes inside double quotes, so escape backslash and quote.
+    ESCAPED_PWD=$(printf '%s' "${WALLET_PWD}" | sed -e 's/\\/\\\\/g' -e 's/"/\\"/g')
+    printf 'MNEMONIC_PASSWORD="%s"\n' "${ESCAPED_PWD}" > "${ENV_FILE}"
```
openoms: **[P1] Keep this secret write out of the passwordless root helper.** `ENV_FILE` is below a path controlled by `joinmarketng`, while this branch runs as root through the sudoers entry. That violates the trusted-path boundary and permits a root-file overwrite if the app-owned target is replaced. The same boundary problem exists in `store-password`. Please perform config/env writes as `joinmarketng`, keep only fixed service actions privileged, and place any genuinely root-owned transient state under a root-owned directory that the app account cannot relink. Per `SECURITY.md`, this describes impact and remediation without publishing a reproduction.

home.admin/config.scripts/bonus.joinmarket-ng.sh
```diff
@@ -0,0 +1,977 @@
...
+  echo "# bonus.${APPID}.sh on        -> install the app"
+  echo "# bonus.${APPID}.sh off       -> uninstall the app"
+  echo "# bonus.${APPID}.sh menu      -> SSH menu dialog"
+  echo "# bonus.${APPID}.sh verify-release [TAG] -> verify release signatures only"
+  echo "# bonus.${APPID}.sh prestart  -> will be called by systemd before start"
+  echo "# bonus.${APPID}.sh update [TAG] -> update to latest or specific version"
+  exit 1
+fi
+
+echo "# Running: 'bonus.${APPID}.sh $*'"
```
openoms: **[P1] Do not log secret-bearing arguments.** `store-password` receives the wallet password as `$2`, so this `$*` output prints the encryption secret; the argument is also visible to process and audit tooling. Change the interface to read the secret from private stdin or a dedicated file descriptor, and log only the command name with sensitive arguments redacted.

home.admin/config.scripts/bonus.joinmarket-ng.sh
```diff
@@ -0,0 +1,977 @@
...
+  if [ "${UPDATE_TAG}" != "main" ] && ! installed_release_matches "${VERIFIED_COMMIT}"; then
+    echo "# FAIL - installed packages do not match verified release commit"
+    exit 1
+  fi
+
+  # Menu script is bundled in the jmcore pip package — updated automatically
+  # by the pip upgrade above. No separate download needed.
+
+  # Only restart maker if it was running before the update (and password file still exists)
+  if [ "${MAKER_WAS_RUNNING}" = "1" ]; then
```
openoms, line 1271: **[P1] Restore the maker that was running before the update.** Stopping the unit runs `ExecStopPost`, which removes `.maker.env`, so this later existence check always fails for the temporary-password path; the permanently stored-password path never needs that file and also fails the check. The update therefore reports success but leaves every previously running maker stopped. Track the password mode before stopping, preserve temporary state safely when appropriate, and restart directly when `toml_has_wallet_password` succeeds.

home.admin/config.scripts/blitz.passwords.sh
```diff
@@ -376,6 +376,16 @@ elif [ "${abcd}" = "b" ]; then
     sudo systemctl restart joinmarket-api.service
   fi
 
+  # JoinMarket-NG
+  if [ "${joinmarketNG}" == "on" ]; then
+    echo "# changing the RPC password for JOINMARKET-NG"
+    # Escape sed metacharacters in the password for safe replacement
+    escapedNewPassword=$(printf '%s' "${newPassword}" | sed -e 's/[&\\/]/\\&/g')
+    sudo sed -i "s/^rpc_password = .*/rpc_password = \"${escapedNewPassword}\"/" /home/joinmarketng/.joinmarket-ng/config.toml
```
openoms: **[P1] Restart the maker after rotating Bitcoin RPC credentials.** Bitcoin Core is restarted earlier in this flow, while JoinMarket-NG loads `rpc_password` at process startup. Updating the file alone leaves a running maker using the old credential, and `Wants=/After=` does not propagate a dependency restart. After the TOML-safe update, restart `joinmarket-ng-maker` so `prestart` and the maker load the new value.

home.admin/config.scripts/bonus.joinmarket-ng.sh
```diff
@@ -0,0 +1,977 @@
...
+  if ! curl -sfL "${GITHUB_RELEASES}/${TAG}/release-manifest-${TAG}.txt" -o "${MANIFEST}"; then
+    echo "# FAIL: Could not download release manifest for ${TAG}"
+    rm -rf "${TMPDIR}"
+    [ -n "${GNUPGHOME_ORIG}" ] && export GNUPGHOME="${GNUPGHOME_ORIG}" || unset GNUPGHOME
+    return 1
+  fi
+
+  # Download trusted-keys.txt
+  local TRUSTED_KEYS="${TMPDIR}/trusted-keys.txt"
+  if ! curl -sfL "${GITHUB_RAW}/signatures/trusted-keys.txt" -o "${TRUSTED_KEYS}"; then
```
openoms: **[P1] Pin the signer trust root in RaspiBlitz.** The trusted fingerprint list, public keys, and signatures are all fetched from the same mutable upstream `main` branch that this code is attempting to authenticate, and the imported key fingerprint is not checked against a RaspiBlitz-owned allowlist. A repository compromise can therefore replace the trust material along with release metadata. Vendor trusted fingerprints or a keyring here, inspect `gpg --status-fd` for the expected `VALIDSIG`, and download the config template from the resulting `VERIFIED_COMMIT` rather than mutable `${GITHUB_TAG}`.

home.admin/config.scripts/bonus.joinmarket-ng.sh
```diff
@@ -0,0 +1,977 @@
...
+  sudo -u ${USER_JM} mkdir -p /mnt/hdd/app-data/${APPID}/wallets
+  sudo chmod 700 /mnt/hdd/app-data/${APPID}/wallets
+
+  # Create logs subdirectory (readable by the user for menu access)
+  sudo -u ${USER_JM} mkdir -p /mnt/hdd/app-data/${APPID}/logs
+
+  # Symlink to home for standard JM-NG path (~/.joinmarket-ng)
+  # JM-NG expects data in ~/.joinmarket-ng by default
+  # We link the folder itself
+  if [ -d "/home/${USER_JM}/.joinmarket-ng" ] && [ ! -L "/home/${USER_JM}/.joinmarket-ng" ]; then
```
openoms: **[P2] Fail closed when this path is a real directory.** The service, password helper, and menu later assume `~/.joinmarket-ng` is the HDD-data symlink, but this branch only warns and continues. That creates split configuration/log/password paths, and `userdel -rf` on uninstall can remove data from the unexpected real home directory. Abort with a remediation message, or migrate the directory explicitly before creating and verifying the symlink.

home.admin/config.scripts/bonus.joinmarket-ng.sh
```diff
@@ -0,0 +1,977 @@
...
+  exit $?
+fi
+
+##########################
+# MAKER-STOP
+##########################
+
+if [ "$1" = "maker-stop" ]; then
+  echo "Stopping ${APPID}-maker..."
+  systemctl stop ${APPID}-maker
```
openoms: **[P2] Preserve the `systemctl stop` result.** The following `rm` and `echo` overwrite `$?`, so the branch returns success even when the maker could not be stopped. Capture `stop_rc=$?` immediately, perform cleanup, and `exit "$stop_rc"` so callers can detect the failure.

openoms (changes requested):
Requesting changes on `c65967fd2fe83a511fa32dbe2400c3f450c5424e`.

The integration and test coverage are substantial, and the exact-head Ubuntu suite passes, but the current implementation is not safe to merge yet.

Required changes:

1. **Separate unprivileged secret writes from privileged service control.** The passwordless root helper writes through paths controlled by `joinmarketng`, which violates the trusted-path boundary and can turn a compromised app account into a root-file overwrite. Remove root from `store-password` and `.maker.env` writes; retain only fixed, narrowly scoped service operations in a root-owned helper. Treat this under `SECURITY.md`; the inline comment intentionally omits reproduction details.
2. **Remove the wallet password from command arguments and logs.** `store-password "$password"` exposes the wallet-encryption secret through process/audit data, while the generic `$*` logging prints it. Read it through a private stdin/file-descriptor interface and redact secret-bearing commands.
3. **Restore maker state after updates.** `systemctl stop` runs `ExecStopPost`, which deletes `.maker.env`; the subsequent file check therefore prevents every previously running maker from restarting. Handle stored-password and temporary-password modes explicitly.
4. **Restart JoinMarket-NG after Password-B rotation.** Bitcoin Core restarts with new RPC credentials, but the maker keeps the credentials loaded at process startup. Restart `joinmarket-ng-maker` after safely updating its config.
5. **Establish a RaspiBlitz-owned trust root.** Do not fetch `trusted-keys.txt`, public keys, and signatures from the same mutable upstream branch being authenticated. Pin trusted signer fingerprints or a keyring locally, verify the actual `VALIDSIG` fingerprint, and fetch the config template from `VERIFIED_COMMIT` rather than `${GITHUB_TAG}`.
6. **Fail closed on a real `~/.joinmarket-ng` directory.** Continuing without the expected symlink produces split data paths and makes uninstall capable of deleting the unexpected home-directory data.
7. **Propagate maker-stop failures.** Capture the `systemctl stop` result before cleanup and return it instead of the final `echo` status.

Validation performed on this head:

- Ubuntu 24.04 Bats: 45/45 passed
- Bash syntax and command-wiring tests passed
- ShellCheck warning confirms the masked `maker-stop` status
- Current GitHub checks for spelling, ShellCheck, PostgreSQL Bats, and JoinMarket-NG Bats pass; the ARM64 image build was still running when this review was submitted
- Current `dev` is ahead of the recorded PR base; merge-tree analysis found no conflict

SOLID review: the first finding is also a concrete single-responsibility/dependency-boundary problem—the same privileged script combines trust verification, user-controlled secret storage, package installation/update, and service control. I found no additional Liskov, open/closed, or interface-segregation issue that warrants a separate change request.

openoms:
Reviewed the latest head, `d58af7238d978a7c981e670da1cd9d45f82491cf`, with the remaining fixes to be pushed as a follow-up to this review.

The recent changes address the wallet-password argv/log exposure, pin the signer trust root and template source, reject an unexpected real home-data directory, and restore maker state after successful updates. The local Ubuntu 24.04 Bats run passed **64/64 tests**.

The following items remain for the follow-up push:

1. **Close the remaining privileged log-path boundary** (`bonus.joinmarket-ng.sh:420-421`). The system unit opens stdout/stderr destinations under app-controlled paths with elevated privileges. This leaves a root-file-write risk despite the narrower service helper. Use journald, or a root-owned log path that the app user cannot redirect. Keeping reproduction details out of this public comment.
2. **Preserve temporary wallet credentials across Password B rotation** (`blitz.passwords.sh:379-386`, `bonus.joinmarket-ng.sh:408-413`). `try-restart` runs `ExecStopPost`, deleting `.maker.env` before the new maker starts. Apply the same secure credential preservation used by the update flow, and exercise temporary-password rotation in a test.
3. **Restart the correct Bitcoin daemon during Password B rotation** (`blitz.passwords.sh:335-347`). The existing `${network}d` restart does not select `tbitcoind` or `sbitcoind`. The new maker restart can then load the new credential while the active chain daemon still uses the old one. Use the chain-aware mapping already added elsewhere in this PR.
4. **Surface maker reload failures during rotation** (`blitz.passwords.sh:386,449`). Check the restart result instead of reaching the unconditional success message when JoinMarket-NG failed to reload. Include a recovery instruction because the Bitcoin credential may already have changed.
5. **Propagate Bitcoin wallet-enablement failures** (`bonus.joinmarket-ng.sh:862`, `network.wallet.sh:47-62`). Check the installer call, and make the helper propagate configuration-write and daemon-restart failures rather than always returning success.
6. **Move `StartLimitIntervalSec=0` into `[Unit]`** (`bonus.joinmarket-ng.sh:419`). It is currently emitted under `[Service]`, where systemd ignores it. Add `systemd-analyze verify` coverage for the generated unit.
7. **Migrate misplaced legacy wallet passwords into `[wallet]`** (`bonus.joinmarket-ng.sh:737-750`). The current replacement leaves a top-level `mnemonic_password` at top level, where upstream ignores it. Make the update section-aware and add the legacy-config migration test.
8. **Complete the Bats workflow path filters** (`.github/workflows/test-bats.yml:10-30`). Include `_commands.sh`, `network.wallet.sh`, and the workflow itself so changes to those files cannot skip the relevant tests.

Also exclude the armored trusted-key file from spelling checks rather than changing key material. The reviewed spelling failure consists of false positives inside that file.

The green Bats result does not validate real systemd lifecycle behavior: several relevant checks only search source text, and the container suite uses a systemctl stub. The follow-up should add executable failure-path and credential-lifecycle coverage, especially for items 1-5. No additional SOLID-specific finding.

This records the remaining work for the follow-up fixes; it is not an approval of the current head. The privilege-boundary issue should be closed before merge.

openoms:
Follow-up fixes are pushed in `75d5ec5d17ce67d58a62c7623eafd10c73290d8d`, addressing each item from the review above:

1. **Privileged log-path boundary closed** - the maker unit no longer uses `StandardOutput/StandardError=append:` into the app-owned data directory. Output goes to the journal (default), viewable via `journalctl -u joinmarket-ng-maker`, which also makes the `blitz.debug.sh` journalctl section functional.
2. **Temporary wallet credentials preserved across Password B rotation** - `.maker.env` is backed up before stop and restored between stop and start, since `ExecStopPost` deletes it. Backup is root-only (mktemp 600) and removed afterwards.
3. **Chain-aware daemon restart** - Password B rotation now restarts `tbitcoind`/`sbitcoind`/`bitcoind` based on `chain` instead of the hardcoded `${network}d`.
4. **Reload failures surfaced** - a failed maker reload prints a FAIL with a recovery command and makes the overall Password B run exit 1 instead of the unconditional OK.
5. **Wallet-enablement failures propagate** - `network.wallet.sh` now returns nonzero when the config write or daemon restart fails (both `on` and `off`), and the JoinMarket-NG installer aborts in that case.
6. **`StartLimitIntervalSec=0` moved to `[Unit]`** - systemd honors it there; under `[Service]` it was silently ignored.
7. **Legacy top-level `mnemonic_password` migration** - `store-password` now rewrites section-aware: stale/misplaced entries (including legacy top-level keys) are removed, the fresh value lands under `[wallet]`, and the result is TOML round-trip validated before writing.
8. **Bats workflow path filters completed** - `_commands.sh`, `network.wallet.sh`, and the workflow file itself now trigger the suite.

Additionally, `typos.toml` now excludes `*.asc`, which fixes the spelling check failure on the pinned release keyring (false positives inside armored key material).

Validation on this head:

- Ubuntu 24.04 Bats: **68/68 passed** (4 new tests: legacy password migration, credential-preserving rotation wiring, visible reload failure, chain-specific restart, wallet-enablement failure propagation)
- ShellCheck (severity error) clean on all touched scripts
- Bash syntax checks pass
- The bats environment still stubs systemd; real-unit behavior (journal logging, restart policy) should be confirmed by the ARM64/image CI and a manual signet run.

Not approval of earlier heads; this closes the follow-up list as far as static and container testing can show. Manual verification on a signet node is still recommended before merge.

Issue #5234, opened 2026-02-02
Specter can't talk to Bitcoin Core.


surfac3:
**Describe the bug**
Specter doesnt connect to Core. After login user is kicked that there is no RPC connection

**To Reproduce**
Update blitz from previous version to 1.12 with functioning specter with data

**Debug file**

Feb 02 06:05:04 raspberrypi python3[795322]: [2026-02-02 06:05:04,693] INFO in wallet_manager: Skipping further update because self.working_folder is None (and data_folder = /home/specter/.specter/>
Feb 02 06:05:04 raspberrypi python3[795322]: [2026-02-02 06:05:04,695] WARNING in wallet_manager: Specter seems to be disconnected from Bitcoin Core. Skipping wallets update.
Feb 02 06:05:24 raspberrypi python3[795322]: [2026-02-02 06:05:24,712] INFO in config_manager: Config loaded from file /home/specter/.specter/config.json
Feb 02 06:05:24 raspberrypi python3[795322]: [2026-02-02 06:05:24,722] INFO in wallet_manager: starting update of wallet_manager (threading: True , comment: via user)
Feb 02 06:05:24 raspberrypi python3[795322]: [2026-02-02 06:05:24,722] INFO in wallet_manager: Skipping further update because self.working_folder is None (and data_folder = /home/specter/.specter/>
Feb 02 06:05:24 raspberrypi python3[795322]: [2026-02-02 06:05:24,723] WARNING in wallet_manager: Specter seems to be disconnected from Bitcoin Core. Skipping wallets update.
Feb 02 06:05:44 raspberrypi python3[795322]: [2026-02-02 06:05:44,732] INFO in config_manager: Config loaded from file /home/specter/.specter/config.json
Feb 02 06:05:44 raspberrypi python3[795322]: [2026-02-02 06:05:44,736] INFO in wallet_manager: starting update of wallet_manager (threading: True , comment: via user)
Feb 02 06:05:44 raspberrypi python3[795322]: [2026-02-02 06:05:44,737] INFO in wallet_manager: Skipping further update because self.working_folder is None (and data_folder = /home/specter/.specter/>
Feb 02 06:05:44 raspberrypi python3[795322]: [2026-02-02 06:05:44,737] WARNING in wallet_manager: Specter seems to be disconnected from Bitcoin Core. Skipping wallets update.

**This issue is not present where there is a fresh install of Specter.**

al-munazzim:
[...]

PR #5244, opened 2026-03-01
#5234 add Enviroment value (#5236)

rootzoll:
Fixing LNbits group permissions to use LND admin macro.


rootzoll:
wrong base - closing

PR #5254, opened 2026-03-13
SD card image filename incorrect in build_sdcard.sh

### LINKED ISSUE #5253: SD card image filename incorrect in build_sdcard.sh
**Describe the bug**
SD card image filename is not the right one in build_sdcard.sh in latest v1.12 release

**Improvement required**
It is not a good idea to separate the filename and the base URL to download the file, I propose du merge the 2 lines with one URL

rofra:
fixes: #5253

### DIFF
--- a/build_sdcard.sh
+++ b/build_sdcard.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 #########################################################################
-# Build your SD card image based on: 2024-03-15-raspios-bookworm-arm64.img.xz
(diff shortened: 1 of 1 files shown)


rofra:
For the automatic build error, you could use the following URL in archives because standard cdimage/releases/current only keeps the current (latest) version.

Here is a good one: https://cdimage.debian.org/cdimage/archive/13.2.0/amd64/iso-cd/debian-13.2.0-amd64-netinst.iso

PR #5264, opened 2026-03-29
ci(amd64): improve Debian ISO resolution and add packer syntax check

openoms:
Port Debian amd64 build reliability improvements for rolling point releases and checksum signature verification, plus CI syntax validation for Packer templates.

Reference: https://github.com/openoms/joininbox/pull/181

### DIFF
--- a/ci/amd64/debian/build.amd64-debian.pkr.hcl
+++ b/ci/amd64/debian/build.amd64-debian.pkr.hcl
@@ -1,8 +1,10 @@
 # images, checksums and signatures are at:
 # https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/
-# Defaults below are safe fallbacks.
-# ci/amd64/packer.build.amd64-debian.sh resolves and injects the latest point-release
-# ISO name and matching checksum at runtime to avoid point-release breakage.
+# NOTE: This template is intended to be invoked via the wrapper script:
+#   ci/amd64/packer.build.amd64-debian.sh
+# The wrapper resolves and injects the latest point-release ISO name and matching checksum
+# at runtime. Defaults below are placeholders and are not guaranteed to work if you run
+# `packer build` directly.
 variable "iso_name" { default = "debian-13-amd64-netinst.iso" }
 variable "iso_checksum" { default = "file:https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/SHA256SUMS" }
 
--- a/ci/amd64/debian/build.amd64-debian.pkr.hcl
+++ b/ci/amd64/debian/build.amd64-debian.pkr.hcl
@@ -1,7 +1,10 @@
 # images, checksums and signatures are at:
 # https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/
-variable "iso_name" { default = "debian-13.2.0-amd64-netinst.iso" }
-variable "iso_checksum" { default = "677c4d57aa034dc192b5191870141057574c1b05df2b9569c0ee08aa4e32125d" }
+# Defaults below are safe fallbacks.
+# ci/amd64/packer.build.amd64-debian.sh resolves and injects the latest point-release
+# ISO name and matching checksum at runtime to avoid point-release breakage.
+variable "iso_name" { default = "debian-13-amd64-netinst.iso" }
+variable "iso_checksum" { default = "file:https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/SHA256SUMS" }
 
 variable "pack" { default = "lean" }
 variable "github_user" { default = "raspiblitz" }
@@ -59,7 +62,7 @@ source "qemu" "debian" {
   disk_size        = var.image_size
   http_directory   = "./http"
   iso_checksum     = var.iso_checksum
-  iso_url          = "https://cdimage.debian.org/cdimage/release/current/amd64/iso-cd/${var.iso_name}"
+  iso_url          = "https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/${var.iso_name}"
   memory           = var.memory
   output_directory = "../builds/${local.name_template}-qemu"
   shutdown_command = "echo 'raspiblitz' | sudo /sbin/shutdown -hP now"
--- a/.github/workflows/amd64-lean-image.yml
+++ b/.github/workflows/amd64-lean-image.yml
@@ -34,8 +34,9 @@ jobs:
       - name: Maximize build space
         uses: easimon/maximize-build-space@master
         with:
-          root-reserve-mb: 12288
-          temp-reserve-mb: 12288
+          # Keep conservative reserves to avoid negative allocation sizes on runner images.
+          root-reserve-mb: 4096
+          temp-reserve-mb: 4096
 
       - uses: actions/checkout@v4
 
--- a/.github/workflows/amd64-lean-image.yml
+++ b/.github/workflows/amd64-lean-image.yml
@@ -34,9 +34,10 @@ jobs:
       - name: Maximize build space
         uses: easimon/maximize-build-space@master
         with:
-          # Keep conservative reserves to avoid negative allocation sizes on runner images.
+          # Keep conservative reserves and a smaller swap size so temp PV sizing stays positive.
           root-reserve-mb: 4096
-          temp-reserve-mb: 4096
+          temp-reserve-mb: 2048
+          swap-size-mb: 1024
 
       - uses: actions/checkout@v4
 
--- a/.github/workflows/packer-syntax-check.yml
+++ b/.github/workflows/packer-syntax-check.yml
@@ -1,0 +1,60 @@
+name: Packer Syntax Check
+
+concurrency:
+  group: packer-syntax-check-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+on:
+  workflow_dispatch:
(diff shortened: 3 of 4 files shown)


ci/amd64/debian/build.amd64-debian.pkr.hcl
```diff
@@ -1,7 +1,10 @@
 # images, checksums and signatures are at:
 # https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/
-variable "iso_name" { default = "debian-13.2.0-amd64-netinst.iso" }
-variable "iso_checksum" { default = "677c4d57aa034dc192b5191870141057574c1b05df2b9569c0ee08aa4e32125d" }
+# Defaults below are safe fallbacks.
+# ci/amd64/packer.build.amd64-debian.sh resolves and injects the latest point-release
+# ISO name and matching checksum at runtime to avoid point-release breakage.
```
openoms:
```suggestion
# NOTE: This template is intended to be invoked via the wrapper script:
#   ci/amd64/packer.build.amd64-debian.sh
# The wrapper resolves and injects the latest point-release ISO name and matching checksum
# at runtime. Defaults below are placeholders and are not guaranteed to work if you run
# `packer build` directly.
```

openoms:
@autoblitzbot review

autoblitzbot (commented):
Quick review: CI is green and the Debian ISO resolution/signature verification flow looks solid overall. Two follow-up nits I'd still consider before merge:

- `ci/amd64/packer.build.amd64-debian.sh` uses `gpg` (`--show-keys` / `--import`) but only ensures `gpgv` is installed. On minimal environments that can fail, so I'd install/check `gnupg` alongside `gpgv`.
- `.github/workflows/packer-syntax-check.yml` currently includes `*.pkrvars.hcl` in the scan and validates each match as a standalone file. That's fine today because there are no var files there, but it will break once one is added. I'd either restrict syntax validation to `*.pkr.hcl` or pass var files only via `-var-file`.

autoblitzbot (commented):
Quick review: CI is green, but I still see three robustness issues worth fixing before merge:

- `ci/amd64/packer.build.amd64-debian.sh` uses `gpg` for fingerprint inspection/import, but the script only checks or installs `gpgv`. On a minimal host without `gnupg`, that path will fail.
- `DEBIAN_MAJOR` is only used for filename filtering, while downloads always come from `debian-cd/current/...`. If the major is overridden, or once stable moves past 13, resolution will break unless the URL also tracks the selected major (or the variable is removed).
- `.github/workflows/packer-syntax-check.yml` validates `*.pkrvars.hcl` as standalone templates. That will start failing as soon as var files appear. Safer to validate only `*.pkr.hcl`, or pass var files via `-var-file` instead of validating them directly.

The overall direction looks good otherwise.

autoblitzbot (commented):
Good direction overall, and the CI is green now.

Two things still look worth tightening before merge:

1. `packer-syntax-check.yml` currently finds both `*.pkr.hcl` and `*.pkrvars.hcl`, but then runs `packer validate -syntax-only` on each file as if it were a template. That will break as soon as a real var file is added. I would either restrict the `find` to `*.pkr.hcl` or validate templates while passing var files separately.

2. `DEBIAN_MAJOR` is configurable, but the download path is still hard-wired to `debian-cd/current/...`, so overriding the major will not really work once stable moves on. I would either remove `DEBIAN_MAJOR` or make the URL selection match the requested major.

Everything else looks sensible to me.

autoblitzbot (changes requested):
Nice direction overall, and the CI is green, but I see two blockers before I’d approve this:

1. `.github/workflows/packer-syntax-check.yml` currently finds both `*.pkr.hcl` and `*.pkrvars.hcl`, then runs `packer validate -syntax-only` on each path as if it were a standalone template. That will break as soon as a real `.pkrvars.hcl` file lands, because var files are not valid templates by themselves. I’d either limit the loop to `*.pkr.hcl`, or pair var files with the relevant template instead of validating them directly.

2. `ci/amd64/packer.build.amd64-debian.sh` uses `gpg` for fingerprint inspection/import, but only checks and installs `gpgv`. On a minimal environment without `gpg`/`gnupg`, this fails before signature verification. Please check/install `gpg` together with `gpgv`.

Once those are addressed, this looks good to me.

autoblitzbot (commented):
Quick review after the ping from @openoms:

Overall this looks solid and the CI is green. Two follow-ups stood out to me:

1. `.github/workflows/packer-syntax-check.yml` currently discovers `*.pkrvars.hcl` files and runs `packer validate -syntax-only` on them directly. That is fine today because there are no matching var files in-tree, but if one gets added later the workflow will likely start failing because var files are not standalone templates. I would either restrict the loop to `*.pkr.hcl` or pass var files to template validation instead of validating them directly.

2. `ci/amd64/packer.build.amd64-debian.sh` uses both `gpg` and `gpgv`, but only checks/installs `gpgv`. GitHub runners already have `gpg`, so this does not break CI, but an explicit `gpg`/`gnupg` check would make the wrapper more portable outside Actions.

So: good direction, but I would treat those as follow-up fixes before calling it fully done.

autoblitzbot (commented):
Gave this a pass. The Debian ISO resolution + signature verification flow looks solid, and CI is green. One thing I'd still tweak before merge: the new syntax workflow currently validates  files as if they were standalone templates. Those var files are not valid templates on their own, so this will start failing once any  files get added. I'd either limit discovery to  or pass var files via  instead of validating them directly.

autoblitzbot (commented):
Gave this a pass. The Debian ISO resolution + signature verification flow looks solid, and CI is green.

One thing I'd still tweak before merge: the new syntax workflow currently validates `*.pkrvars.hcl` files as if they were standalone templates. Those var files are not valid templates on their own, so this will start failing once any `.pkrvars.hcl` files get added. I'd either limit discovery to `*.pkr.hcl` or pass var files via `-var-file` instead of validating them directly.

autoblitzbot (commented):
Quick review: this is heading in the right direction and CI is green, but I still see two correctness issues before I would approve:

- `.github/workflows/packer-syntax-check.yml` validates `*.pkrvars.hcl` as if they were standalone templates. Those files are var files, not templates, so this workflow will start failing as soon as one is added. I would either limit the loop to `*.pkr.hcl` or validate templates while passing matching var files via `-var-file`.
- `ci/amd64/packer.build.amd64-debian.sh` introduces `DEBIAN_MAJOR`, but `debian_iso_dir` is still hard-coded to `debian-cd/current/...`. If `DEBIAN_MAJOR` is overridden, or once stable moves past 13, the requested major and the resolved ISO source can diverge. I would either remove `DEBIAN_MAJOR` or switch to a major-specific URL that matches the selected release.

Optional hardening: the script calls `gpg` for fingerprint inspection/import, but only checks or installs `gpgv`. Adding an explicit `gpg`/`gnupg` check would make the script more portable on minimal environments.

autoblitzbot (approved):
Reviewed. CI is green, and resolving the current Debian netinst ISO from signed SHA256SUMS is a solid improvement over pinning a point release. Non-blocking follow-up: the new syntax-check workflow currently includes *.pkrvars.hcl; if var files are added later, validating them directly may fail, so it may be worth narrowing that to *.pkr.hcl or passing var files via -var-file.

autoblitzbot (commented):
Quick review from Autoblitz:

Overall this looks solid and the CI results are green. Two follow-ups stood out to me:

1. `ci/amd64/packer.build.amd64-debian.sh` uses `gpg --show-keys` and `gpg --import`, but the script only guarantees `gpgv`. On lean environments where `gnupg` is missing, the fingerprint/import step can fail before signature verification.
2. `.github/workflows/packer-syntax-check.yml` includes `*.pkrvars.hcl` in the `find` list and then runs `packer validate -syntax-only` on each file as if it were a template. That will likely break once a standalone var file is added.

So I would treat it as basically good, but with those two portability / future-proofing nits worth tightening up.

autoblitzbot (commented):
Solid change overall. The Debian ISO resolution + signed SHA256SUMS flow makes the build much less brittle, and CI is green.\n\nOne follow-up nit I noticed: the new syntax-check workflow currently finds both  and , then runs  on each path as if it were a template. That works today because there do not seem to be any  files in-tree, but it could become a future footgun once one gets added.\n\nNot blocking this PR, just worth tightening later so var files are not validated as standalone templates.

autoblitzbot (approved):
Canonical summary is here: https://github.com/raspiblitz/raspiblitz/pull/5264#issuecomment-4359693671 . This approval supersedes my earlier duplicate reviews caused by bot re-processing.

autoblitzbot (commented):
Quick pass: overall this looks solid and CI is green. Two follow-up nits worth considering:

- `packer-syntax-check.yml` currently finds `*.pkrvars.hcl` and validates them as if they were templates. That works today because there do not seem to be any var files in scope, but it will start failing once one is added. Safer to validate only `*.pkr.hcl`, or pass var files via `-var-file` instead of validating them directly.
- `DEBIAN_MAJOR` is configurable, but the download path is still hard-wired to `debian-cd/current/...` (current stable). That is fine for now, but it means the major override and future stable transitions can drift apart. If you want to keep `DEBIAN_MAJOR`, it may be worth pointing it at a matching major-specific location or dropping the knob.

rootzoll (approved):

PR #5268, opened 2026-04-11
Fix: Randomize UUIDs on systemcopy

autoblitzbot:
When copying the system via `blitz.data.sh copy-system`, the destination NVMe receives a cloned partition setup. This PR explicitly randomizes the filesytem UUID (using `tune2fs -U random`) and the MBR/GPT Disk Identifier (using `fdisk` / `sgdisk`) before writing the new UUIDs to `fstab` and `cmdline.txt`.

This ensures the cloned SSD/NVMe drive has a unique PARTUUID from the installer media, preventing split-brain boot loops.

**Test Report (PiKVM environment):**
I tested this exact logic on a physical Raspberry Pi 5 equipped with a PiKVM.
While this fix completely resolves the identical PARTUUID issue (Option A), I discovered during testing that if the USB Virtual Media is still plugged in, the EEPROM `BOOT_ORDER=0xf461` prioritize USB (4) over NVMe (6). 

Since `kill-boot` fails to delete the partition on read-only/virtual media, the Pi falls back to booting the USB installer regardless of the new unique UUIDs on the NVMe. Therefore, while this PR is a valid preventative measure, an interactive prompt (Option B: 'Please disconnect USB drive now') will still be required for fully headless immutable setups like PiKVM to prevent the loop.

### DIFF
--- a/home.admin/_bootstrap.sh
+++ b/home.admin/_bootstrap.sh
@@ -1003,14 +1003,28 @@ if [ "${scenario}" != "ready" ] ; then
       # put flag file into old system 
       touch /home/admin/systemcopy.flag
 
-      # disable old system boot
-      echo "# disable old system boot" >> ${logFile}
-      /home/admin/config.scripts/blitz.data.sh kill-boot ${installDevice} >> ${logFile}
-      if [ $? -eq 1 ]; then
-        echo "FAIL: blitz.data.sh kill-boot \"${installDevice}\" failed" >> ${logFile}
-        /home/admin/_cache.sh set state "error"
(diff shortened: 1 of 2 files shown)


rootzoll:
@autoblitzbot Can you detect if it's a booting situation from PiKVM or a booting situation from SD card done by the user? If yes then use automatic kill-boot when sd card and for PiKVM setup ask the system to dettach the simulated USB flash drive and reboot (or trigger reboot and dettach flash drive .. what ever works cleaner). Only add solution to PR if possible and works tested.

rootzoll (approved):
LGTM and important for future automated testing thru PiKVM

PR #5270, opened 2026-04-20
update mempool app to v3.3.0

mononaut:
This update brings the mempool app to version 3.3.0.

Highlights from version 3.3.0:
- Taproot script tree visualization
- Sighash icons & highlighting
- Stale block comparisons & visualizations
- Annex goggles support & labeling
- Sub-1-Sat/vB support
- Ephemeral dust support
- Display signatures in PSBT previews
- /api/tx/:txId/merkle-proof API endpoint for non-esplora backends
- Angular 16 -> 20 upgrade
- Decimal fee recommendations
- Taproot witness annotations
- Miscellaneous bugfixes and improvements
- Regtest support

Full release notes are found at https://github.com/mempool/mempool/releases

### DIFF
--- a/home.admin/config.scripts/bonus.mempool.sh
+++ b/home.admin/config.scripts/bonus.mempool.sh
@@ -2,7 +2,7 @@
 
 # https://github.com/mempool/mempool
 
-pinnedVersion="v3.2.1"
+pinnedVersion="v3.3.0"
 
 # command info
 if [ $# -eq 0 ] || [ "$1" = "-h" ] || [ "$1" = "-help" ]; then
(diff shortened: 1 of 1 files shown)


home.admin/config.scripts/bonus.mempool.sh
xphade, line 154: Fixed upstream since 3.0.0 AFAICT, so this can be removed.

xphade (commented):
Just some info from me updating to `v3.3.1`, so one point release newer than this PR:

 * PGP verification works
 * Workaround for mempool/mempool#5122 can be removed (see my comment)
 * I had issues with the `npm ci` calls
   * Got the following error message:
     > `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing. 
   * After changing the calls to `npm install` everything worked perfectly (and I can see that the `frontend/package-lock.json` file actually changed
   * Could be an upstream issue (outdated lock file) I guess, but I'm not too familiar with node

After that, everything worked perfectly for me. Tested it with a clean install (`off -> uninstall -> install -> on`).

PR #5276, opened 2026-04-30
cl-plugin.backup.sh: add existence check before removal

steepdawn974:
Wrap the backup file move operation in a conditional check to prevent errors when the backup file doesn't exist during plugin removal.

### DIFF
--- a/home.admin/config.scripts/cl-plugin.backup.sh
+++ b/home.admin/config.scripts/cl-plugin.backup.sh
@@ -118,10 +118,12 @@ elif
 then
   echo "# Removing the backup plugin"
   sudo rm -f /home/bitcoin/${netprefix}cl-plugins-enabled/backup.py
-  echo "# Backup the existing old backup on the SDcard"
-  now=$(date +"%Y_%m_%d_%H%M%S")
-  sudo mv /home/bitcoin/${netprefix}lightningd.sqlite3.backup \
-    /home/bitcoin/${netprefix}lightningd.sqlite3.backup.${now}
+  if [ -f /home/bitcoin/${netprefix}lightningd.sqlite3.backup ]; then
+    echo "# Backup the existing old backup on the SDcard"
+    now=$(date +"%Y_%m_%d_%H%M%S")
+    sudo mv /home/bitcoin/${netprefix}lightningd.sqlite3.backup \
+      /home/bitcoin/${netprefix}lightningd.sqlite3.backup.${now}
+  fi
   echo "# Removing the backup.lock file"
   sudo rm -f /home/bitcoin/.lightning/${CLNETWORK}/backup.lock
 


home.admin/config.scripts/cl-plugin.backup.sh
```diff
@@ -118,10 +118,12 @@ elif
 then
   echo "# Removing the backup plugin"
   sudo rm -f /home/bitcoin/${netprefix}cl-plugins-enabled/backup.py
-  echo "# Backup the existing old backup on the SDcard"
-  now=$(date +"%Y_%m_%d_%H%M%S")
-  sudo mv /home/bitcoin/${netprefix}lightningd.sqlite3.backup \
-    /home/bitcoin/${netprefix}lightningd.sqlite3.backup.${now}
+  if [ -f /home/bitcoin/${netprefix}lightningd.sqlite3.backup ]; then
```
openoms, line 121: Will need to check if the backup file is readable by the admin user.
If not might need to use
```suggestion
  if sudo ls /home/bitcoin/${netprefix}lightningd.sqlite3.backup; then
```

autoblitzbot (approved):
Reviewed the change. The added existence guard prevents the remove path from failing when the backup file is already absent, and the script still parses cleanly with bash -n.

home.admin/config.scripts/cl-plugin.backup.sh
steepdawn974: @openoms  Good point, but this does apply to all the other places this is used `if [ -f .. ]` as well

https://github.com/raspiblitz/raspiblitz/pull/5276/changes#diff-2d16d6fae7c6f8f031fbba663f1c6cb75fff106d5b8f5b88e622fef950215b1dL32
https://github.com/raspiblitz/raspiblitz/pull/5276/changes#diff-2d16d6fae7c6f8f031fbba663f1c6cb75fff106d5b8f5b88e622fef950215b1dL55
https://github.com/raspiblitz/raspiblitz/pull/5276/changes#diff-2d16d6fae7c6f8f031fbba663f1c6cb75fff106d5b8f5b88e622fef950215b1dL77

etc.


I suggest also we use `test`  instead of `ls` 
e.g

`if sudo test -f "/home/bitcoin/${netprefix}lightningd.sqlite3.backup"; then..`

autoblitzbot: I agree `sudo test -f "/home/bitcoin/${netprefix}lightningd.sqlite3.backup"` is the cleaner check here. It keeps the existence test under the same permissions as the later `sudo mv`, avoids noisy `ls` output, and fits this remove-path guard without broadening the PR. The older `[ -f ... ]` checks can be cleaned up separately if we want to make the whole script consistent.

PR #5278, opened 2026-05-08
docs: clarify Proxmox data disk mount path

autoblitzbot:
## Summary
- document that RaspiBlitz services expect the data disk at `/mnt/hdd`
- add verification commands before starting services
- add bind-mount example for Proxmox/Debian installs where the disk is mounted elsewhere, e.g. `/mnt/disk_storage`
- warn about hidden rootfs data if services are started before `/mnt/hdd` is mounted

## Context
This came from a Proxmox support case where the large data disk was mounted at `/mnt/disk_storage` while RaspiBlitz services used `/mnt/hdd`. That caused nginx TLS files and Electrs DB paths to resolve to the small root filesystem until `/mnt/hdd` was fixed.

## Testing
Docs-only change; verified markdown diff locally.

### DIFF
--- a/alternative.platforms/Proxmox/README.md
+++ b/alternative.platforms/Proxmox/README.md
@@ -115,6 +115,49 @@ Regardless of whether variant 1 or 2 was executed, the VM should now have 2 hard
 
 ![](images/2022-09-21_21-04.png)
 
+### Verify the data disk mount inside the VM
+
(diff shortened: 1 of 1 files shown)


openoms (approved):
improve docs as a result of telegram convo, ready to merge @rootzoll

PR #5279, opened 2026-05-12
Update Bitcoin Knots to v29.3.knots20260508

Retropex:

### DIFF
--- a/home.admin/config.scripts/bonus.knots.sh
+++ b/home.admin/config.scripts/bonus.knots.sh
@@ -12,10 +12,10 @@ APPID="knots" # one-word lower-case no-specials
 
 # clean human readable version - will be displayed in UI
 # just numbers only separated by dots (2 or 0.1 or 1.3.4 or 3.4.5.2)
-VERSION="29.2.2"
+VERSION="29.3"
 
 FILEMASTER="29.x"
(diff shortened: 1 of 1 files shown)


lindelleric:
Don't we also need to add this to the bitcoin.conf? 
```
consensusrules=rdts
```

EDIT: Like this? I looked at how it's done in https://github.com/Retropex/raspiblitz/blob/main/home.admin/config.scripts/blitz.i2pd.sh
```
# Set the RDTS config
/home/admin/config.scripts/blitz.conf.sh set consensusrules rdts /mnt/hdd/app-data/bitcoin/bitcoin.conf noquotes
```
And clean it up in off:
```
/home/admin/config.scripts/blitz.conf.sh delete consensusrules /mnt/hdd/app-data/bitcoin/bitcoin.conf noquotes
```

Retropex:
We should not add it automatically, the goal is to let the user know that this version will change the consensus rules.
That's said the GUIX build are compiled with `-DRDTS_CONSENT=RUNTIME_WARN` so it won't prevent the node from starting.

lindelleric:
I'd vote on the setting being added when knots is installed, with a message that is has been added, and where to manually remove it if you don't want to signal for the bip.

Another option would be to prompt the user if they want it during install, kinda like the GUI does.

PR #5243, opened 2026-02-27
btcpay update to 2.3.5, skip nbxplorer on update due to .net version requirement mismatch

openoms:
fixes: https://github.com/raspiblitz/raspiblitz/issues/5242

PR #5280, opened 2026-05-19
Update BTCPayServer to v2.3.9 with .NET 10

autoblitzbot:
Follow-up for #5243 / requested in https://github.com/raspiblitz/raspiblitz/pull/5243#issuecomment-4487369559

This updates the PR branch from BTCPayServer v2.3.5 to the current v2.3.9 release and updates the dependent runtime pieces.

Changes:
- BTCPayServer v2.3.9
- NBXplorer v2.6.7
- .NET SDK 10.0.300 with linux arm64/x64 SHA512 hashes from the official .NET release manifest
- shared DotNetInstall path for install and update, so existing installs get .NET 10 before rebuilding
- update path restores NBXplorer updates and verifies both NBXplorer and BTCPayServer checked-out commits with Nicolas Dorier's key before build

Verified locally:
- bash -n home.admin/config.scripts/bonus.btcpayserver.sh
- git diff --check
- git verify-commit HEAD passes for btcpayserver/btcpayserver v2.3.9
- git verify-commit HEAD passes for btcpayserver/NBXplorer v2.6.7

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ## What's new in Version 1.12.2 of RaspiBlitz?
 
-- Update: BTCPayServer v2.3.5 [details](https://github.com/btcpayserver/btcpayserver/releases/tag/v2.3.5)
+- Update: BTCPayServer v2.3.9 [details](https://github.com/btcpayserver/btcpayserver/releases/tag/v2.3.9)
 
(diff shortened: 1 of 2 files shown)


openoms:
Tested UPDATE manually on amd64, merging.

PR #5283, opened 2026-05-31
Bound bitcoin monitor RPC probes

vortexopenclaw:
## Problem

RaspiBlitz uses `bitcoin.monitor.sh` from background/status paths to collect Bitcoin Core status, network, info, and mempool data. Those paths currently call `bitcoin-cli` directly.

If Bitcoin Core is running but RPC is overloaded or slow, for example during recovery, sync, high disk I/O, or wallet loading, a single slow RPC call can block the monitor path indefinitely. That can leave background/status scans stuck, accumulate stale `bitcoin-cli` processes, and make the dashboard or terminal menu show stale or missing Bitcoin status data.

This is most visible when heavier RPC calls like `getnetworkinfo` or `getpeerinfo` stall even though lighter status checks can still recover.

## Fix

- add an 8-second timeout wrapper around the `bitcoin.monitor.sh` RPC probes used by `status`, `network`, `info`, and `mempool`
- treat RPC timeout or empty status output as not ready instead of accidentally reporting ready
- allow the `network` action to still report a peer count from `getconnectioncount`, then established P2P sockets, when heavier peer/network RPC data is unavailable

## User-visible result

When Bitcoin Core RPC is slow or overloaded, RaspiBlitz status collection now fails fast and reports a temporary RPC timeout instead of wedging the background/status scan. The network view can still show a useful peer count when the heavier network RPC data is unavailable.

## Scope

This only changes `home.admin/config.scripts/bitcoin.monitor.sh`. It does not change Bitcoin Core startup, service configuration, or node policy.

## Verification

- `bash -n home.admin/config.scripts/bitcoin.monitor.sh home.admin/00infoBlitz.sh home.admin/_background.scan.sh home.admin/_background.sh`
- `git diff --check`
- shell simulation: original unbounded `bitcoin-cli` form required an outer timeout and exited `124` after about 2 seconds
- shell simulation: wrapped form exited `124` after its own 1-second timeout
- shell simulation: status timeout maps to `btc_ready=0` with RPC timeout message
- shell simulation: network peer fallback returned `btc_peers=3` from `getconnectioncount` after `getnetworkinfo` timeout

### DIFF
--- a/home.admin/config.scripts/bitcoin.monitor.sh
+++ b/home.admin/config.scripts/bitcoin.monitor.sh
@@ -22,16 +22,57 @@ fi
 if [ "$1" == "mainnet" ]; then
   bitcoincli_alias="/usr/local/bin/bitcoin-cli -datadir=/home/bitcoin/.bitcoin -rpcport=8332"
   service_alias="bitcoind"
+  bitcoin_p2p_port="8333"
 elif [ "$1" == "testnet" ]; then
   bitcoincli_alias="/usr/local/bin/bitcoin-cli -datadir=/home/bitcoin/.bitcoin -rpcport=18332"
   service_alias="tbitcoind"
+  bitcoin_p2p_port="18333"
 elif [ "$1" == "signet" ]; then
   bitcoincli_alias="/usr/local/bin/bitcoin-cli -datadir=/home/bitcoin/.bitcoin -rpcport=38332"
   service_alias="sbitcoind"
+  bitcoin_p2p_port="38333"
 else
   echo "error='not supported net'"
   exit 1
 fi
+
+bitcoin_monitor_rpc_timeout="${BITCOIN_MONITOR_RPC_TIMEOUT:-8}"
+
+bitcoin_cli() {
+  timeout "${bitcoin_monitor_rpc_timeout}" ${bitcoincli_alias} "$@"
+}
+
+get_rpc_error_short() {
+  if [ "$1" == "124" ]; then
+    echo "RPC timeout"
+  else
+    echo "RPC failed"
+  fi
+}
+
+get_rpc_error_full() {
+  if [ "$1" == "124" ]; then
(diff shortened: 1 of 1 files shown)


vortexopenclaw:
Additional behavior proof:

I reproduced the failure mode with a simulated hanging `bitcoin-cli` call. The original monitor path required an outer timeout and exited `124` after about 2 seconds, showing that the RPC probe could block the caller indefinitely.

With this patch, the monitor's own timeout wrapper returned `124` after its configured deadline, so the status path fails fast instead of wedging background/status scans.

I also verified the user-facing fallback behavior:
- status timeout returns `btc_ready=0` with `btc_error_short=RPC timeout`
- network fallback returns `btc_peers=3` from `getconnectioncount` when `getnetworkinfo` times out

Static checks also passed:
- `bash -n home.admin/config.scripts/bitcoin.monitor.sh home.admin/00infoBlitz.sh home.admin/_background.scan.sh home.admin/_background.sh`
- `git diff --check`

vortexopenclaw:
Still relevant from an operator perspective.

I have had to wrap local RaspiBlitz health/status checks downstream so slow Bitcoin Core RPC does not make monitoring or automation look like a hard tool failure. The underlying problem this PR targets is the same one: background/status paths should not be able to hang indefinitely on a single slow RPC probe while the node is recovering, syncing, under disk I/O pressure, or wallet/RPC load is high.

The useful part of this PR is that it is narrow:

- bounds the monitor-facing RPC calls instead of changing Bitcoin Core service behavior
- reports temporary RPC timeout/not-ready state instead of wedging status collection
- keeps the network view useful by falling back to cheaper peer-count signals when heavier network RPC data is unavailable

From a user-visible behavior standpoint, I think the acceptance bar should be:

- a slow `getnetworkinfo` or `getpeerinfo` cannot block the background/status scan indefinitely
- timeout is visible as a temporary status condition, not a stale ready state
- peer count still has a lightweight fallback when the heavy network RPC path is unhealthy
- no changes to node policy, bitcoin.conf, or service startup semantics

PR #5291, opened 2026-06-12
albyhub update to 1.23.0 and fix systemd service parameter

openoms:


openoms:
@autoblitzbot add the UPDATE option for Albyhub to fix https://github.com/raspiblitz/raspiblitz/issues/5217

PR #5293, opened 2026-06-14
Add AlbyHub update option

autoblitzbot:
Adds an AlbyHub UPDATE option and a `bonus.albyhub.sh update` action for raspiblitz/raspiblitz#5291.\n\nI could not push directly to the source branch, so this is stacked on top of that PR branch.


openoms:
@autoblitzbot move the whiptail menu contents to the menu file and add the option to the UPDATE menu

openoms (approved):

PR #5299, opened 2026-06-27
Add Wasabi Wallet: client daemon (Services menu) + WabiSabi coordinator (bonus)

kravens:
## Summary

Adds two RaspiBlitz integrations for [WalletWasabi](https://github.com/WalletWasabi/WalletWasabi):

1. **`bonus.wasabid.sh`** — the headless **Wasabi client wallet daemon** (`WalletWasabi.Daemon`) with a local JSON-RPC interface, exposed as a regular **SERVICES menu** entry for everyday users.
2. **`bonus.wasabi.sh`** — the **WabiSabi coinjoin coordinator** backend, as an advanced **bonus script only** (intentionally *not* in the menu — running a coordinator is for advanced operators).

Both pin the latest official release (`v2.8.0`) and build from source.

## What's included

- `home.admin/config.scripts/bonus.wasabid.sh` — install/uninstall/on/off/status/menu/examples/update
- `home.admin/config.scripts/bonus.wasabi.sh` — install/uninstall/on/off/status/menu/update
- `home.admin/00settingsMenuServices.sh` — Services menu entry for the daemon (`wd`)
- `home.admin/_provision_.sh` — re-activate `wasabid` + `wasabi` after updates

## Daemon (client) details

- Dedicated `wasabid` user; per-user .NET SDK in `~/.dotnet`.
- Local-only JSON-RPC on `127.0.0.1:37128` with generated basic-auth credentials (no port exposed to the network).
- Wired to the **local bitcoind** via RPC (`UseBitcoinRpc`, `BitcoinRpcCredentialString`, `BitcoinRpcUri`) — uses the user's own node.
- Installs the dev-maintained **`wcli`** as a system command (auto-loads endpoint + credentials from `Config.json`, so users never type ports/passwords).
- `bonus.wasabid.sh examples` prints a copy-paste cheat sheet (`createwallet`, `recoverwallet`, `getnewaddress`, `startcoinjoin`, `payincoinjoin`) leading with `wcli`, curl as fallback.
- Verified against the daemon source: starts wallet-less (`InitializeNoWalletAsync`), RPC server up immediately, no restart loop before a wallet exists.

## Coordinator details

- Dedicated `wasabi` user; per-user .NET SDK.
- Binds `127.0.0.1:37126` only and is reached over a **Tor onion service by default** — no clearnet bind, no firewall port, no domain or port-forwarding, and the operator's IP stays private. (Operators who want clearnet can add their own reverse proxy.)
- **Ensures required `bitcoin.conf` settings** on activation: `txindex=1` (via `network.txindex.sh`), `blockfilterindex=1`, `peerblockfilters=1`, `server=1`, with a single bitcoind restart.
- Generates `Config.json` on first run (fresh `CoordinatorExtPubKey`), then patches in the bitcoind RPC connection; `chmod 600`.
- **Safety guard:** refuses to overwrite a pre-existing manually-installed coordinator service (won't clobber a hand-built deployment).

## Design notes

- **Services run a published build** (`dotnet publish` → `ExecStart=dotnet …/X.dll`), not `dotnet run` — no restore/build or network at service start.
- **.NET SDK channel is read from the project's `global.json`** — v2.7.2 used `8.0`, v2.8.0 uses `10.0`, provisioned automatically (verified on linux-arm64).
- **Network-aware**: `${network}.conf` path, `${network}d` service, Wasabi `Network` (Main/TestNet/RegTest).
- No secrets committed; RPC passwords are generated locally and stored `0600`.
- The coordinator changes are upstreamed and released in v2.8.0, so both scripts pin the `v2.8.0` tag (no master-tracking).

## Testing

- `bash -n` clean on all changed files; embedded Python validated with `ast.parse`.
- `.NET` channel parser verified for `8.0` / `10.0` / missing `global.json`.
- RPC method names, port, `UseTor` enum, and `wcli` quoting verified against the official docs and source.
- Config keys verified against a running v2.7.2 deployment.

## Notes for reviewers

- Coordinator and daemon target different user tiers; happy to split into two PRs if preferred.
- Defaults to mainnet config wiring; testnet/regtest mapped but less exercised.

---

---

## Update — v2.8.0 (verified on a linux-arm64 node)

Bumped both scripts from v2.7.2 to v2.8.0 and fixed issues that surfaced on v2.8.0, all verified end-to-end on a Raspberry Pi (linux-arm64):

1. **Build failure (NU1903).** v2.7.2 failed `dotnet publish`: the NuGet audit advisory [GHSA-2m69-gcr7-jv3q](https://github.com/advisories/GHSA-2m69-gcr7-jv3q) (`SQLitePCLRaw.lib.e_sqlite3`) was promoted to an error under the project's `TreatWarningsAsErrors`. v2.8.0 bumps the dependency and adds a solution-wide `NuGetAuditSuppress`, so the build is clean with no script-side flag.
2. **.NET 10.** v2.8.0's `global.json` moves the SDK channel `8.0` -> `10.0`; `ensure_dotnet_sdk` provisions .NET 10 on demand, no script change.
3. **Bundled Tor not executable.** `dotnet publish` drops the execute bit on the bundled native binaries (Tor, hwi), so the daemon aborted at startup with "Permission denied" launching Tor and restart-looped. Restored with `chmod +x` after publish (install + update paths).
4. **Config.json no longer honored.** v2.8.0 runs a config migration on load that resets hand-written `Config.json` keys to defaults -- `JsonRpcServerEnabled` flipped to false (RPC server never bound; `wcli` couldn't connect) and the `UseBitcoinRpc` wiring was dropped (client synced over P2P, not the local node). Settings now go through `WASABI_*` environment variables (higher precedence, survive the rewrite) from a root-owned `0600` `EnvironmentFile`. Result: `getstatus` reports the local node (`peers: []`) and RPC binds `127.0.0.1:37128`. Local RPC is auth-less (loopback only, no firewall port) -- upstream's own localhost RPC model.

### Additional fixes found while testing both services end-to-end on linux-arm64

5. **bitcoind `rpcport` was read network-blind.** RaspiBlitz's `bitcoin.conf` is network-scoped (`main.rpcport=8332`, `test.rpcport=18332`, …); both scripts collapsed the prefixes and took the last `rpcport`, wiring a mainnet node to `18332`. Now the active network's scoped value is read (with the standard per-network default as fallback).
6. **Daemon RPC switch name.** The daemon's `BitcoinRpcUri` is populated by the switch named `BitcoinRpcEndPoint` (there is no `UseBitcoinRpc` switch), so the env file now sets `WASABI_BITCOINRPCENDPOINT` + `WASABI_BITCOINRPCCREDENTIALSTRING`. Verified: daemon now syncs from the local node (`Fetched fee rate from RPC node`), not P2P.
7. **Coordinator binds localhost + onion only.** It previously bound a second endpoint on `127.0.0.1:5000`, which collides with common services (LNBits) and crashed Kestrel. It now binds a single `127.0.0.1:37126` endpoint and publishes a Tor onion unconditionally (ensuring Tor is installed first). Verified: coordinator reaches the node (`Bitcoin Node is fully synchronized`) and creates WabiSabi rounds.

_Note: the daemon (`bonus.wasabid.sh`) is verified on linux-arm64. The coordinator (`bonus.wasabi.sh`) carries the same v2.8.0 + Tor fixes but its config wiring is still being validated separately._

### DIFF
--- a/home.admin/00settingsMenuServices.sh
+++ b/home.admin/00settingsMenuServices.sh
@@ -17,6 +17,7 @@ if [ ${#fulcrum} -eq 0 ]; then fulcrum="off"; fi
 if [ ${#lndmanage} -eq 0 ]; then lndmanage="off"; fi
 if [ ${#joinmarket} -eq 0 ]; then joinmarket="off"; fi
 if [ ${#jam} -eq 0 ]; then jam="off"; fi
+if [ ${#wasabid} -eq 0 ]; then wasabid="off"; fi
 if [ ${#LNBits} -eq 0 ]; then LNBits="off"; fi
 if [ ${#mempoolExplorer} -eq 0 ]; then mempoolExplorer="off"; fi
 if [ ${#bos} -eq 0 ]; then bos="off"; fi
@@ -51,6 +52,7 @@ if [ "${network}" == "bitcoin" ]; then
   OPTIONS+=(aa 'BTC Mempool Space' ${mempoolExplorer})
   OPTIONS+=(ja 'BTC JoinMarket+JoininBox menu' ${joinmarket})
   OPTIONS+=(za 'BTC Jam (JoinMarket WebUI)' ${jam})
+  OPTIONS+=(wd 'BTC Wasabi Wallet (daemon + JSON-RPC)' ${wasabid})
   OPTIONS+=(wa 'BTC Download Bitcoin Whitepaper' ${whitepaper})
   OPTIONS+=(ls 'BTC Labelbase' ${labelbase})
   OPTIONS+=(pp 'BTC Publicpool (Solo Mining)' ${publicpool})  
@@ -605,6 +607,25 @@ Then try activating Jam again in SERVICES.\n
   fi
 else
   echo "Jam not changed."
+fi
+
+# Wasabi daemon (client wallet) process choice
+choice="off"; check=$(echo "${CHOICES}" | grep -c "wd")
+if [ ${check} -eq 1 ]; then choice="on"; fi
+if [ "${wasabid}" != "${choice}" ]; then
+  echo "Wasabi daemon setting changed .."
+  anychange=1
+  sudo /home/admin/config.scripts/bonus.wasabid.sh ${choice}
+  errorOnInstall=$?
+  if [ "${choice}" = "on" ]; then
+    if [ ${errorOnInstall} -eq 0 ]; then
+      sudo /home/admin/config.scripts/bonus.wasabid.sh menu
+    else
+      whiptail --title 'FAIL' --msgbox "Wasabi daemon installation is cancelled\nTry again from the menu or install from the terminal with:\nsudo /home/admin/config.scripts/bonus.wasabid.sh on" 9 65
+    fi
+  fi
+else
+  echo "Wasabi daemon not changed."
 fi
 
 # Mempool process choice
--- a/home.admin/00settingsMenuServices.sh
+++ b/home.admin/00settingsMenuServices.sh
@@ -610,7 +610,7 @@ else
 fi
 
 # Wasabi daemon (client wallet) process choice
-choice="off"; check=$(echo "${CHOICES}" | grep -c "wd")
+choice="off"; check=$(echo "${CHOICES}" | grep -cw "wd")
 if [ ${check} -eq 1 ]; then choice="on"; fi
 if [ "${wasabid}" != "${choice}" ]; then
   echo "Wasabi daemon setting changed .."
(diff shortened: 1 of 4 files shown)


kravens:
Tested on a Raspberry Pi5 with RaspiBlitz v1.12.0:
image
Looks like it all works, let me know what I missed to get this merged 😄 I've asked Claude to complete my partial bonus.wasabi.sh script I wrote last year. Now it should be feature complete with menu integration, so it can't get easier to start a coinjoin coordinator or use the commandline wcli to manage a headless wasabiwallet.

PR #5304, opened 2026-08-09
Improve brittle Public Pool setup script

xphade:
The Public Pool script and config is a bit brittle and actually broken right now because of upstream changes which are incompatible with the `environment.ts` changes in here. To fix this, I pinned the repositories to know-working versions (around the time the script was last touched). We now have to manually update it, but I think that's better than the script just randomly breaking.

I also added support for remote access via Tailscale.  The UI now uses relative API requests through the local Angular proxy instead of hardcoding the local IP. Finally, I also added the Pi hostname and Tailscale MagicDNS name (if available) to make browser usage a bit nicer (previously you would get a "Invalid Host Header" error when trying to access e.g. `http://raspiblitz.local:3335`.

General comment: There are still broader improvements worth considering. For example, the script builds the frontend but appears to ignore the resulting build and serves it using `ng serve` in a development-style setup. I intentionally limited this change to the minimum fixes needed for reliable installation and Tailscale access, though.

Tested it locally on my raspiblitz by a fresh install of this app.

### DIFF
--- a/home.admin/config.scripts/bonus.publicpool.sh
+++ b/home.admin/config.scripts/bonus.publicpool.sh
@@ -4,7 +4,8 @@ APPID="publicpool"
 VERSION="0.1"
 GITHUB_REPO="https://github.com/benjamin-wilson/public-pool.git"
 GITHUB_REPO_UI="https://github.com/benjamin-wilson/public-pool-ui.git"
-GITHUB_TAG=""
+GITHUB_COMMIT="b971e9ce4ccd23ae98536d57dcf63657ade7919f"
+GITHUB_COMMIT_UI="1c0b2d93e3ce0a81d4faa7b1d444ace936e3f63d"
 
 PORT_API="3334"
 PORT_STRATUM="3333"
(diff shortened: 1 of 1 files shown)


home.admin/config.scripts/bonus.publicpool.sh
```diff
@@ -100,27 +102,64 @@ if [ "$1" = "1" ] || [ "$1" = "on" ]; then
...
+    }
+}
+EOL
+
+  # Allow the Pi hostname and -- if installed -- the Tailscale MagicDNS name in Angular's host check.
+  hostName=$(hostname)
+  tailscaleDNS=""
+  if command -v tailscale >/dev/null 2>&1; then
+    tailscaleDNS=$(sudo tailscale status --json 2>/dev/null | jq -r '.Self.DNSName // empty' 2>/dev/null | sed 's/\.$//')
+  fi
```
xphade, line 146: One thing to mention: This will obviously not work (i.e., the hostname will not be added) if tailscale is installed after public pool. But I think that's fine for the setup script, users can also just add that line manually.

PR #5314, opened 2026-08-29
cl.install.sh: add update-binary path for prebuilt CLN tarballs

autoblitzbot:
## Why

The v26.06.7 security release is under a 2-week embargo: the source zip (`clightning-v26.06.7.zip`) is not published, only prebuilt tarballs. The existing `cl.install.sh update <version>` path downloads the source zip and builds from source, so it 404s for embargoed releases. Upstream explicitly recommends upgrading promptly via the tarballs rather than waiting for the source to drop.

## What

Adds `cl.install.sh update-binary <version>`:

- maps `uname -m` to the amd64/arm64 release assets (arm64 uses the separate `SHA256SUMS-<version>-arm64` manifest)
- picks the oldest Ubuntu build compatible with the local glibc (22.04/2.35, 24.04/2.39, 26.04/2.41) — on a Pi 5 with Debian 12 this selects the Ubuntu-22.04-arm64 build
- verifies the GPG-signed checksum manifest (reuses the existing cln/ngoline primary+fallback PGP key logic) and the tarball SHA256
- extracts with `--strip-components=2` into `/usr/local`, checks the installed version, restarts enabled lightningd services
- generalizes `verifySHA256SUMSSignature` to accept a manifest name (existing source-zip path unchanged)

## Tested

- `bash -n` syntax check
- download + checksum verification of `clightning-v26.06.7-Ubuntu-22.04-arm64.tar.xz` against `SHA256SUMS-v26.06.7-arm64`
- GPG verification of `SHA256SUMS-v26.06.7-arm64.asc` — good signature with the ngoline fallback key already used by the script
- glibc detection on Debian 12 (2.36 → Ubuntu 22.04 build)

Not run end-to-end on a live blitz (no node to hand) — the interactive confirmation and service restart steps would be worth a quick check on a test node.

### DIFF
--- a/.github/workflows/test-bats.yml
+++ b/.github/workflows/test-bats.yml
@@ -35,7 +35,7 @@ jobs:
         # no sudo: these only exercise the version / asset existence checks
         run: |
           cd test
-          bats --verbose-run ./cl.install.bats
+          bats ./cl.install.bats
 
       - name: Run the bats tests with postgresql 15
         run: |
--- a/.github/workflows/test-bats.yml
+++ b/.github/workflows/test-bats.yml
@@ -10,10 +10,14 @@ on:
     branches: ["dev"]
     paths:
       - "home.admin/config.scripts/bonus.postgresql.sh"
+      - "home.admin/config.scripts/cl.install.sh"
+      - "test/**"
   pull_request:
     branches: ["*"]
     paths:
       - "home.admin/config.scripts/bonus.postgresql.sh"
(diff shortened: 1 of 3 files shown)


openoms:
Runnig to this error on update:

image

@autoblitzbot create a bats test to avoid errors like this

openoms:
Tested ok with
menu - UPDATE - PATCH - PR - 5314
CLN version v26.06.7 is now installed on an amd64 Raspiblitz.

PR #5318, opened 2026-09-05
fix(nginx): forward WebSocket upgrades to Blitz API

cstenglein:
Enable HTTP/1.1 and forward Upgrade and Connection headers in both public API proxy templates so /api/ws reaches the WebSocket endpoint.

Validated both templates with nginx -t and isolated HTTP/HTTPS proxies against Blitz API: health, login, authenticated warmup, and 4401 authentication rejection.


rootzoll (commented):
utACK

PR #5320, opened 2026-09-12
Add menu option to disable public peer connectivity for LND and CLN

autoblitzbot:
## What

In view of the current vulnerabilities affecting publicly reachable Lightning nodes, this adds a per-implementation toggle (LND menu / CLN menu) to make the node **outbound-only**: not reachable and not announced on the network, while channels keep working over the outbound Tor connections.

### raspiblitz.conf (written by the menu toggle)
```ini
lndPublicPeers=off   # or =on (default / unset = on)
clPublicPeers=off    # or =on (default / unset = on)
```

### LND — /mnt/hdd/app-data/lnd/lnd.conf (applied by lnd.check.sh prestart)

When `lndPublicPeers=off`:
```diff
 [Application Options]
+nolisten=1                  # no p2p listener at all (incl. clearnet)
 [tor]
 tor.active=true             # unchanged - outbound over Tor stays
-tor.v3=true
+tor.v3=false                # no inbound Tor hidden service is created
```
Note: lnd rejects `nolisten` + `tor.v3=true` (`ValidateConfig: listening must be enabled when enabling inbound connections over Tor`, config.go: `DisableListen && Tor.V3`), so `tor.v3=false` is required. The hidden service is only created when `tor.v3=true`.

When re-enabled (`lndPublicPeers=on` / unset):
```diff
 [Application Options]
-nolisten=1                  # line removed
 [tor]
-tor.v3=false
+tor.v3=true                 # inbound hidden service restored
```

### CLN — /home/bitcoin/.lightning/config (applied by cl.check.sh prestart, honored at install by cl.install.sh)

When `clPublicPeers=off`:
```diff
 proxy=127.0.0.1:9050        # unchanged - outbound over Tor stays
 bind-addr=127.0.0.1:9736    # unchanged - localhost only
-addr=statictor:127.0.0.1:9051/torport=9736   # removed - no announced Tor address
 always-use-proxy=true       # unchanged
-announce-addr=...           # removed if present
```
No announced address = outbound-only, node is not announced in gossip.

When re-enabled with `runBehindTor=on`:
```diff
+addr=statictor:127.0.0.1:9051/torport=9736   # restored
```

## Other changes

- **`99lndMenu.sh` / `99clMenu.sh`** — new `Public Peer Connectivity: ON/OFF` menu entry. Toggling writes `lndPublicPeers` / `clPublicPeers` to `raspiblitz.conf` and restarts the service so the change is applied.
- The check scripts now *respect* the disabled setting instead of resetting the conf back to public on every start.

## Behavior

- Default is unchanged: unset = public connectivity **ON**, existing nodes are unaffected.
- Disabling = outbound-only node: existing channels keep working over outbound connections, but no new inbound channels can be opened *to* the node and it is no longer announced in gossip.

## Testing

- `bash -n` syntax check on all modified scripts: OK
- Simulated the sed/grep enforcement logic on sample `lnd.conf` / CLN `config`: disable strips announced addresses / sets `nolisten=1` + `tor.v3=false`, re-enable restores defaults: OK
- Verified against lnd master source that the `nolisten` + `tor.v3=true` combination is rejected and that the inbound hidden service is only created when `tor.v3=true`.

### DIFF
--- a/home.admin/99clMenu.sh
+++ b/home.admin/99clMenu.sh
@@ -23,6 +23,11 @@ OPTIONS=()
   OPTIONS+=(RECEIVE "Create an invoice / payment request")
   OPTIONS+=(SUMMARY "Information about this node")
   OPTIONS+=(NAME "Change the name / alias of the node")
+if [ "${clPublicPeers}" == "off" ]; then
+  OPTIONS+=(PUBPEERS "Public Peer Connectivity: OFF")
+else
+  OPTIONS+=(PUBPEERS "Public Peer Connectivity: ON")
+fi
 ln_getInfo=$($lightningcli_alias getinfo 2>/dev/null)
 ln_channels_online="$(echo "${ln_getInfo}" | jq -r '.num_active_channels')" 2>/dev/null
 cl_num_inactive_channels="$(echo "${ln_getInfo}" | jq -r '.num_inactive_channels')" 2>/dev/null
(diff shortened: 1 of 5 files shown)


openoms:
Testing: 

image

image

image

openoms:
image

image

openoms (approved):
Tested on Rpi5 Raspiblitz v0.12.1 both LND and CLN successfully.
Ready to merge.

